PR #1, opened 2017-08-18
Download action

d4s:
Added new action 'download' with unpack and install abilities.

### DIFF
--- a/debos.go
+++ b/debos.go
@@ -98,6 +98,7 @@ type DebosContext struct {
 	imageKernelRoot string       // Kernel cmdline root= snippet for the / of the image
 	recipeDir       string
 	Architecture    string
+	origins         map[string]string
 }
 
 type Action interface {
@@ -269,6 +270,12 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
+
+	// Initialise origins map
+	context.origins = make(map[string]string)
+	context.origins["artifacts"] = context.artifactdir
+	context.origins["filesystem"] = context.rootdir
+	context.origins["recipe"] = context.recipeDir
 
 	r := Recipe{}
 
--- a/debos.go
+++ b/debos.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strings"
 	"text/template"
 
 	"github.com/jessevdk/go-flags"
@@ -86,6 +87,19 @@ func CopyTree(sourcetree, desttree string) error {
 	}
 
 	return filepath.Walk(sourcetree, walker)
+}
+
+func RestrictedPath(prefix, dest string) (string, error) {
+	var err error
+	destination := path.Join(prefix, dest)
+	destination, err = filepath.Abs(destination)
+	if err != nil {
+		return "", err
+	}
+	if !strings.HasPrefix(destination, prefix) {
+		return "", fmt.Errorf("The resulting path points outside of prefix '%s': '%s'\n", prefix, destination)
+	}
+	return destination, nil
 }
 
 type DebosContext struct {
--- a/debos.go
+++ b/debos.go
@@ -175,6 +175,8 @@ func (y *YamlAction) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		y.Action = newFilesystemDeployAction()
 	case "raw":
 		y.Action = &RawAction{}
+	case "download":
+		y.Action = &DownloadAction{}
 	default:
 		log.Fatalf("Unknown action: %v", aux.Action)
 	}
--- a/download_action.go
+++ b/download_action.go
@@ -52,17 +52,43 @@ func DownloadHttpUrl(url, filename string) error {
 	return nil
 }
 
+// validateUrl checks if supported URL is passed from recipe
+// Return:
+// - parsed URL
+// - nil in case of success
+func (d *DownloadAction) validateUrl() (*url.URL, error) {
+
+	url, err := url.Parse(d.Url)
+	if err != nil {
+		return url, err
+	}
+
+	switch url.Scheme {
+	case "http", "https":
+		// Supported scheme
+	default:
+		return url, fmt.Errorf("Unsupported URL is provided: '%s'", url.String())
+	}
+
+	return url, nil
+}
+
+func (d *DownloadAction) Verify(context *DebosContext) error {
+
+	_, err := d.validateUrl()
+	return err
+}
+
 func (d *DownloadAction) Run(context *DebosContext) error {
 	var filename string
 	d.LogStart()
 
-	url, err := url.Parse(d.Url)
+	url, err := d.validateUrl()
 	if err != nil {
 		return err
 	}
 
 	if len(d.Filename) == 0 {
-		log.Printf("No output filename is provided for '%s'", d.Url)
 		// Trying to guess the name from URL Path
 		filename = path.Base(url.Path)
 	} else {
--- a/download_action.go
+++ b/download_action.go
@@ -1,0 +1,87 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"net/url"
+	"os"
+	"path"
+)
+
+type DownloadAction struct {
+	BaseAction `yaml:",inline"`
+	Url        string // URL for downloading
+	Filename   string // File name, overrides the name from URL.
+}
+
+// Function for downloading single file object with http(s) protocol
+func DownloadHttpUrl(url, filename string) error {
+	log.Printf("Download started: '%s' -> '%s'\n", url, filename)
+
+	// TODO: Proxy support?
+
+	// Check if file object already exists.
+	fi, err := os.Stat(filename)
+	if !os.IsNotExist(err) && !fi.Mode().IsRegular() {
+		return fmt.Errorf("Failed to download '%s': '%s' exists and it is not a regular file\n", url, filename)
+	}
+
+	resp, err := http.Get(url)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("Url '%s' returned status code %d (%s)\n", url, resp.StatusCode, http.StatusText(resp.StatusCode))
+	}
+
+	// Output file
+	output, err := os.Create(filename)
+	if err != nil {
+		return err
+	}
+	defer output.Close()
+
+	if _, err := io.Copy(output, resp.Body); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (d *DownloadAction) Run(context *DebosContext) error {
+	var filename string
+	d.LogStart()
+
+	url, err := url.Parse(d.Url)
+	if err != nil {
+		return err
+	}
+
+	if len(d.Filename) == 0 {
+		log.Printf("No output filename is provided for '%s'", d.Url)
+		// Trying to guess the name from URL Path
+		filename = path.Base(url.Path)
+	} else {
+		filename = path.Base(d.Filename)
+	}
+	if len(filename) == 0 {
+		return fmt.Errorf("Incorrect filename is provided for '%s'", d.Url)
+	}
+	filename = path.Join(context.scratchdir, filename)
+
+	switch url.Scheme {
+	case "http", "https":
+		err := DownloadHttpUrl(url.String(), filename)
+		if err != nil {
+			return err
+		}
+	default:
+		return fmt.Errorf("Unsupported URL is provided: '%s'", url.String())
+	}
+
+	return nil
+}
--- a/download_action.go
+++ b/download_action.go
@@ -16,6 +16,7 @@ type DownloadAction struct {
 	Filename    string // File name, overrides the name from URL.
 	Unpack      bool   // Unpack downloaded file to directory dedicated for download
 	Compression string // compression type
+	Name        string // exporting path to file or directory(in case of unpack)
 }
 
 // Function for downloading single file object with http(s) protocol
@@ -77,6 +78,9 @@ func (d *DownloadAction) validateUrl() (*url.URL, error) {
 
 func (d *DownloadAction) Verify(context *DebosContext) error {
 
+	if len(d.Name) == 0 {
+		return fmt.Errorf("Property 'name' is mandatory for download action\n")
+	}
 	_, err := d.validateUrl()
 	return err
 }
@@ -100,6 +104,7 @@ func (d *DownloadAction) Run(context *DebosContext) error {
 		return fmt.Errorf("Incorrect filename is provided for '%s'", d.Url)
 	}
 	filename = path.Join(context.scratchdir, filename)
+	originPath := filename
 
 	switch url.Scheme {
 	case "http", "https":
@@ -113,8 +118,14 @@ func (d *DownloadAction) Run(context *DebosContext) error {
 
 	if d.Unpack == true {
 		targetdir := filename + ".d"
-		return UnpackTarArchive(filename, targetdir, d.Compression, "--no-same-owner", "--no-same-permissions")
+		err := UnpackTarArchive(filename, targetdir, d.Compression, "--no-same-owner", "--no-same-permissions")
+		if err != nil {
+			return err
+		}
+		originPath = targetdir
 	}
+
+	context.origins[d.Name] = originPath
 
 	return nil
 }
--- a/download_action.go
+++ b/download_action.go
@@ -11,9 +11,11 @@ import (
 )
 
 type DownloadAction struct {
-	BaseAction `yaml:",inline"`
-	Url        string // URL for downloading
-	Filename   string // File name, overrides the name from URL.
+	BaseAction  `yaml:",inline"`
+	Url         string // URL for downloading
+	Filename    string // File name, overrides the name from URL.
+	Unpack      bool   // Unpack downloaded file to directory dedicated for download
+	Compression string // compression type
 }
 
 // Function for downloading single file object with http(s) protocol
@@ -107,6 +109,11 @@ func (d *DownloadAction) Run(context *DebosContext) error {
 		}
 	default:
 		return fmt.Errorf("Unsupported URL is provided: '%s'", url.String())
+	}
+
+	if d.Unpack == true {
+		targetdir := filename + ".d"
+		return UnpackTarArchive(filename, targetdir, d.Compression, "--no-same-owner", "--no-same-permissions")
 	}
 
 	return nil
--- a/overlay_action.go
+++ b/overlay_action.go
@@ -1,16 +1,30 @@
 package main
 
 import (
+	"fmt"
 	"path"
 )
 
 type OverlayAction struct {
-	BaseAction `yaml:",inline"`
-	Source     string
+	BaseAction  `yaml:",inline"`
+	Origin      string // origin of overlay, here the export from other action may be used
+	Source      string // external path there overlay is
+	Destination string // path inside of rootfs
 }
 
 func (overlay *OverlayAction) Run(context *DebosContext) error {
 	overlay.LogStart()
-	sourcedir := path.Join(context.recipeDir, overlay.Source)
-	return CopyTree(sourcedir, context.rootdir)
+	origin := context.recipeDir
+
+	//Trying to get a filename from exports first
+	if len(overlay.Origin) > 0 {
+		var found bool
+		if origin, found = context.origins[overlay.Origin]; !found {
+			return fmt.Errorf("Origin not found '%s'", overlay.Origin)
+		}
+	}
+
+	sourcedir := path.Join(origin, overlay.Source)
+
+	return CopyTree(sourcedir, path.Join(context.rootdir, overlay.Destination))
 }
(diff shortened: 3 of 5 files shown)


download_action.go
```diff
@@ -0,0 +1,104 @@
...
+	"net/http"
+	"net/url"
+	"os"
+	"path"
+)
+
+type DownloadAction struct {
+	BaseAction `yaml:",inline"`
+	Url        string // URL for downloading
+	Filename   string // File name, overrides the name from URL. TODO: optional
```
sjoerdsimons: I'd leave this one out in the initial commits. Whether it's useful will probably depend on how we end up using things and it'll clash with unpacking?

download_action.go
```diff
@@ -100,5 +102,9 @@ func (d *DownloadAction) Run(context *DebosContext) error {
 		return fmt.Errorf("Unsupported URL is provided: '%s'", url.String())
 	}
 
+	if d.Unpack == true {
+		return UnpackTarArchive(filename, context.rootdir, d.Compression)
```
sjoerdsimons: Download actions shouldn't unpack to the root directory. Both the download and the unpack should go into a seperate area such that follow-up actions can pick an choice what to put where rather then mandating a tarball following your root layout.

So the general idea is: 
* Download action, give the result some tag name.
* Overlay, copy, raw actions to transfer the result into the rootfs/image/whatever (this can be done N times).

e.g. one example could be downloading a tarball with firmware and then copying that firmware to sepcific location on the rootfs.

Fwiw the source field in the raw action is already going towards this, by having filesystem as the currently supported source (e.g. dd from something inside the filesystem we build)

download_action.go
```diff
@@ -71,17 +71,43 @@ func DownloadHttpUrl(url, filename string, forcedRewrite bool) error {
...
 	d.LogStart()
 
-	url, err := url.Parse(d.Url)
+	url, err := d.validateUrl()
 	if err != nil {
 		return err
 	}
 
 	if len(d.Filename) == 0 {
-		log.Printf("No output filename is provided for '%s'", d.Url)
```
sjoerdsimons: spurious removal?

download_action.go
```diff
@@ -128,8 +174,19 @@ func (d *DownloadAction) Run(context *DebosContext) error {
 		return fmt.Errorf("Unsupported URL is provided: '%s'", url.String())
 	}
 
+	// Unpack and Install are mutually exclusive
 	if d.Unpack == true {
 		return UnpackTarArchive(filename, context.rootdir, d.Compression)
+	} else if d.Install == true {
```
sjoerdsimons: This should be a seperate action rather then installing directly from a download action, see also the preivous comment :). That way we can also nicely support bundling a .deb in a recipe and install that.

unpack_action.go
```diff
@@ -12,12 +12,23 @@ type UnpackAction struct {
...
+	unpackTarOpt := map[string]string{
+		"gz":    "z",
+		"bzip2": "j",
+		"xz":    "J",
+	} // Trying to guess all other supported formats
+
+	os.MkdirAll(destination, 0755)
+	log.Printf("Unpacking %s\n", infile)
+
+	return Command{}.Run("unpack", "tar", "-x"+unpackTarOpt[compression], "-f", infile, "-C", destination)
```
d4s: not sure about it.
Modern tar implementations can guess compression type if you do not provide any option.
btw, this is the reason why there is no 'default' for compression switch.

download_action.go
d4s: Nope. Forgot to remove "TODO: optional" above. It is not mandatory.

unpack_action.go
sjoerdsimons: I'm not too fond if having it optional though, as that relies on what the system tar does. But while that's reasonable it still means that if a compression is given ti should be verified if it's one on the known list

download_action.go
```diff
@@ -0,0 +1,95 @@
...
+
+// Function for downloading single file object with http(s) protocol
+func DownloadHttpUrl(url, filename string, forcedRewrite bool) error {
+	log.Printf("Download started: '%s' -> '%s'\n", url, filename)
+
+	// TODO: Proxy support?
+
+	// Check the file. Assume we have the file if file already exists.
+	// Ignore check if flag 'forcedRewrite' is set
+	fi, err := os.Stat(filename)
```
sjoerdsimons: Don't do this in this simple way until there is a better way to validate things then just the name (e.g. by also saving and verifying the eTag of a http download).  I'd prefer to see this as a later addition after landing the download action

download_action.go
```diff
@@ -0,0 +1,95 @@
...
+	"net/url"
+	"os"
+	"path"
+)
+
+type DownloadAction struct {
+	BaseAction `yaml:",inline"`
+	Url        string // URL for downloading
+	Filename   string // File name, overrides the name from URL.
+	Forced     bool   `yaml:"force"` // Forced re-download of URL
```
sjoerdsimons: Drop the forcing, unless there is a strong use-case for it. In a recipe that would essentially mean always-redownload, which when caching is added would be a cache workaround. For one-time forcing the user can always drop the cache

download_action.go
```diff
@@ -0,0 +1,95 @@
...
+	"net/http"
+	"net/url"
+	"os"
+	"path"
+)
+
+type DownloadAction struct {
+	BaseAction `yaml:",inline"`
+	Url        string // URL for downloading
+	Filename   string // File name, overrides the name from URL.
```
sjoerdsimons: What's the use-case for this one?

download_action.go
```diff
@@ -11,10 +11,13 @@ import (
...
-	Url        string // URL for downloading
-	Filename   string // File name, overrides the name from URL.
-	Forced     bool   `yaml:"force"` // Forced re-download of URL
+	BaseAction  `yaml:",inline"`
+	Url         string // URL for downloading
+	Filename    string // File name, overrides the name from URL.
+	Forced      bool   `yaml:"force"` // Forced re-download of URL
+	Unpack      bool   // Unpack downloaded file to directory dedicated for download
+	Compression string // compression type
+	Makedir     bool   // Create a dedicated directory for extracting contents
```
sjoerdsimons: As discussed on chat, we should always make dedicated directories for extracting so there isn't a chance for accidental overrides. If the user wants download B to override parts of A that should be ensure when deploying the contents onto the filesystem rather then at download time

download_action.go
```diff
@@ -117,5 +129,17 @@ func (d *DownloadAction) Run(context *DebosContext) error {
...
+	if d.Unpack == true {
+		targetdir := context.downloaddir
+		if d.Makedir == true {
+			targetdir = filename + ".d"
+			if err := os.MkdirAll(targetdir, 0755); err != nil {
+				return err
+			}
+		}
+
+		return UnpackTarArchive(filename, targetdir, d.Compression, "--no-same-owner", "--no-same-permissions")
```
sjoerdsimons: --no-same-owner and no-same-permissions are interesting. I guess it's the right thing  to do indeed :). Does make me wonder if the tar refactoring was right by exposing tar options directly. Otoh it's internal api so that's fine ;)

download_action.go
```diff
@@ -60,17 +60,43 @@ func DownloadHttpUrl(url, filename string, forcedRewrite bool) error {
 	return nil
 }
 
+// Validate if supported parameters passed in recipe
```
sjoerdsimons: nitpick the go convention is that the first line starts with the name of what it describes, so validateUrl in this case ;)

overlay_action.go
```diff
@@ -1,15 +1,26 @@
...
+		sourcedir, err = context.searchpath.FindByName(overlay.Source)
+		if err != nil {
+			return err
+		}
+	} else if len(overlay.Source) > 0 {
+		sourcedir += "/" + overlay.Source
 	}
-	return CopyTree(sourcedir, context.rootdir)
+
+	return CopyTree(sourcedir, context.rootdir+overlay.Destination)
```
sjoerdsimons: Join paths using proper go function not concatination

unpack_action.go
```diff
@@ -50,9 +50,15 @@ func (pf *UnpackAction) Verify(context *DebosContext) error {
 
 func (pf *UnpackAction) Run(context *DebosContext) error {
 	pf.LogStart()
-	infile, err := context.searchpath.FindByName(pf.File)
-	if err != nil {
-		return err
+
+	//Trying to get a filename from exports first
+	infile, found := context.exports[pf.File]
+	if !found {
```
sjoerdsimons: This should be an error rather then searching

overlay_action.go
```diff
@@ -1,15 +1,26 @@
...
 }
 
 func (overlay *OverlayAction) Run(context *DebosContext) error {
 	overlay.LogStart()
-	sourcedir, err := context.searchpath.FindByName(overlay.Source)
-	if err != nil {
-		return err
+	//Trying to get a filename from exports first
+	sourcedir, found := context.exports[overlay.Origin]
+	if !found {
```
sjoerdsimons: This should be an error rather then falling back onto a search dir

download_action.go
```diff
@@ -115,7 +117,15 @@ func (d *DownloadAction) Run(context *DebosContext) error {
...
 		targetdir := filename + ".d"
-		return UnpackTarArchive(filename, targetdir, d.Compression, "--no-same-owner", "--no-same-permissions")
+		err := UnpackTarArchive(filename, targetdir, d.Compression, "--no-same-owner", "--no-same-permissions")
+		if err != nil {
+			return err
+		}
+		exportSymbol = targetdir
+	}
+
+	if len(d.Name) > 0 {
```
sjoerdsimons: Name should be mandatory for download actions as you can't do anything with them unless you have the name

download_action.go
```diff
@@ -102,6 +103,7 @@ func (d *DownloadAction) Run(context *DebosContext) error {
 		return fmt.Errorf("Incorrect filename is provided for '%s'", d.Url)
 	}
 	filename = path.Join(context.scratchdir, filename)
+	exportSymbol := filename
```
sjoerdsimons: nitpick i find that name confusing :) maybe make it originPath or somesuch?

overlay_action.go
```diff
@@ -1,16 +1,31 @@
...
 	overlay.LogStart()
-	sourcedir := path.Join(context.recipeDir, overlay.Source)
-	return CopyTree(sourcedir, context.rootdir)
+	origin := context.recipeDir
+
+	//Trying to get a filename from exports first
+	if len(overlay.Origin) > 0 {
+		source, found := context.origins[overlay.Origin]
+		if !found {
+			return fmt.Errorf("Empty export '%s' found.", overlay.Origin)
```
sjoerdsimons: Error message probalby should be `Origin not found '%s'`

overlay_action.go
```diff
@@ -1,16 +1,31 @@
...
+		}
+		origin = source
+	}
+
+	sourcedir := path.Join(origin, overlay.Source)
+
+	return CopyTree(sourcedir, path.Join(context.rootdir, overlay.Destination))
```
sjoerdsimons: todo item, ensure destination isn't causing the destination to point outside of the rootdir

unpack_action.go
```diff
@@ -51,7 +51,12 @@ func (pf *UnpackAction) Verify(context *DebosContext) error {
 
 func (pf *UnpackAction) Run(context *DebosContext) error {
 	pf.LogStart()
-	infile := path.Join(context.artifactdir, pf.File)
+
+	//Trying to get a filename from exports first
+	infile, found := context.origins[pf.File]
```
sjoerdsimons: We shouldn't overload File to mean origin. Also the file to unpack could be in a downloaded tarball, so you stil lwant Origin

So this chunk should instead add an Origin key - File is optional when the Origin is a named action which produces a file (e.g. download without unpack set) or when given it's the path of the file to unpack within the origin (by default the artifact dir)

raw_action.go
```diff
@@ -11,22 +11,25 @@ import (
 
 type RawAction struct {
 	BaseAction `yaml:",inline"`
+	Origin     string // there te source comes from
 	Offset     string
-	Source     string
-	Path       string
```
sjoerdsimons: We need to keep this in to not break current recipes

raw_action.go
```diff
@@ -11,22 +11,25 @@ import (
...
+	Source     string // relative path inside of origin
 }
 
 func (raw *RawAction) Verify(context *DebosContext) error {
-	if raw.Source != "filesystem" {
-		return errors.New("Only suppport sourcing from filesystem")
+	if len(raw.Origin) <= 0 || len(raw.Source) <= 0 {
```
sjoerdsimons: A string can't be smaller then 0 :) so that's an odd check

raw_action.go
sjoerdsimons: We need to keep backwards compatibility with Source + Path usage. As suggested in a previous commit, you can use the Verify action to detect that usage (iotw Path is not "") and set Source = Path and Origin = filesystem afterwhich things would flow.

Would make sense to also print a note sayig that usage is deprecated

download_action.go
d4s: done

overlay_action.go
d4s: done

overlay_action.go
d4s: done

raw_action.go
d4s: done

raw_action.go
d4s: Pretty good old habbit ;)
No influence to generated code so I'd left it as is.

unpack_action.go
d4s: done

download_action.go
```diff
@@ -0,0 +1,88 @@
...
+		log.Printf("No output filename is provided for '%s'", d.Url)
+		// Trying to guess the name from URL Path
+		filename = path.Base(url.Path)
+	} else {
+		filename = path.Base(d.Filename)
+	}
+	if len(filename) == 0 {
+		return fmt.Errorf("Incorrect filename is provided for '%s'", d.Url)
+	}
+	filename = path.Join(context.scratchdir, filename)
```
sjoerdsimons: There is a potential here for clashing things if the same filename ends up being used for two download actions, but that can be fixed in a later patch

unpack_action.go
```diff
@@ -12,12 +13,45 @@ type UnpackAction struct {
...
+		command = append(command, unpackTarOpt)
+	}
+	command = append(command, "-f", infile, "-C", destination)
+
+	return Command{}.Run("unpack", command...)
+}
+
+func (pf *UnpackAction) Verify(context *DebosContext) error {
+	unpackTarOpt := tarOptions(pf.Compression)
+	if len(pf.Compression) > 0 && len(unpackTarOpt) <= 0 {
```
sjoerdsimons: I find using <= 0 really weird when it's unsigned return values :)

unpack_action.go
```diff
@@ -12,12 +13,45 @@ type UnpackAction struct {
...
+	unpackTarOpts := map[string]string{
+		"gz":    "-z",
+		"bzip2": "-j",
+		"xz":    "-J",
+	} // Trying to guess all other supported formats
+
+	return unpackTarOpts[compression]
+}
+
+func UnpackTarArchive(infile, destination, compression string, options ...string) error {
```
sjoerdsimons: todo for later, move this into some utility file so that we don't have actiosn calling helpers from other actions :)

download_action.go
```diff
@@ -77,6 +79,7 @@ func (d *DownloadAction) validateUrl() (*url.URL, error) {
 func (d *DownloadAction) Verify(context *DebosContext) error {
 
 	_, err := d.validateUrl()
+
```
sjoerdsimons: spurious newline?

debos.go
```diff
@@ -270,6 +271,12 @@ func main() {
 		panic(err)
 	}
 
+	// Initialise origins map
+	context.origins = make(map[string]string)
+	context.origins["artifactdir"] = context.artifactdir
+	context.origins["filesystem"] = context.rootdir
+	context.origins["recipedir"] = context.recipeDir
```
sjoerdsimons: Can we make this artifacts, and recipe, dropping the dir suffix as they're all directories :)

unpack_action.go
```diff
@@ -51,12 +51,7 @@ func (pf *UnpackAction) Verify(context *DebosContext) error {
 
 func (pf *UnpackAction) Run(context *DebosContext) error {
 	pf.LogStart()
-
-	//Trying to get a filename from exports first
-	infile, found := context.origins[pf.File]
-	if !found {
-		infile = path.Join(context.artifactdir, pf.File)
-	}
+	infile := path.Join(context.artifactdir, pf.File)
```
sjoerdsimons: This should be in a seperate patch

sjoerdsimons: Well actually tin the patch that introduced the issue

raw_action.go
sjoerdsimons: Please squash those patches together, we shouldn't break something in a patch and then fix it in the next as that will make bisecting nasty

raw_action.go
sjoerdsimons: Yeah but it's weird for the reader so better change it :)

raw_action.go
```diff
@@ -11,22 +11,25 @@ import (
...
+		return errors.New("'origin' and 'source' properties can't be empty")
 	}
-
```
sjoerdsimons: spurious line removal?

unpack_action.go
```diff
@@ -51,7 +56,23 @@ func (pf *UnpackAction) Verify(context *DebosContext) error {
 
 func (pf *UnpackAction) Run(context *DebosContext) error {
 	pf.LogStart()
-	infile := path.Join(context.artifactdir, pf.File)
+	var origin string
+
+	if len(pf.Origin) > 0 {
+		//Trying to get a filename from origins first
+		val, found := context.origins[pf.Origin]
```
sjoerdsimons: Why use a temporary variable here? Just assign directly to origin

unpack_action.go
```diff
@@ -51,7 +56,23 @@ func (pf *UnpackAction) Verify(context *DebosContext) error {
...
+		if !found {
+			return fmt.Errorf("Origin not found '%s'", pf.Origin)
+		}
+		origin = val
+	} else {
+		origin = context.artifactdir
+	}
+
+	infile, err := RestrictedPath(origin, pf.File)
```
sjoerdsimons: This probably means that function is another candidate to move to a utility file

download_action.go
d4s: fixed

download_action.go
d4s: done

debos.go
d4s: done.

download_action.go
d4s: done

unpack_action.go
d4s: done

raw_action.go
d4s: done

raw_action.go
d4s: fixed

raw_action.go
d4s: done

unpack_action.go
d4s: done. in all places with len()

unpack_action.go
d4s: Yes. During refactoring.

download_action.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// Function for downloading single file object with http(s) protocol
+func DownloadHttpUrl(url, filename string) error {
+	log.Printf("Download started: '%s' -> '%s'\n", url, filename)
+
+	// TODO: Proxy support?
+
+	// Check if file object already exists.
+	// Ignore check if flag 'forcedRewrite' is set
```
sjoerdsimons: Outdated comment

download_action.go
```diff
@@ -0,0 +1,88 @@
...
+	fi, err := os.Stat(filename)
+	if !os.IsNotExist(err) && !fi.Mode().IsRegular() {
+		return fmt.Errorf("Failed to download '%s': '%s' is not a regular file\n", url, filename)
```
sjoerdsimons: That error should indicate that filename already exists and isn't a regular file (otherwise it's quite unclear what happened)

overlay_action.go
```diff
@@ -1,16 +1,31 @@
...
 
 func (overlay *OverlayAction) Run(context *DebosContext) error {
 	overlay.LogStart()
-	sourcedir := path.Join(context.recipeDir, overlay.Source)
-	return CopyTree(sourcedir, context.rootdir)
+	origin := context.recipeDir
+
+	//Trying to get a filename from exports first
+	if len(overlay.Origin) > 0 {
+		source, found := context.origins[overlay.Origin]
```
sjoerdsimons: nitpick source temporary variable is pointless

download_action.go
d4s: done

download_action.go
d4s: done

overlay_action.go
d4s: done

PR #2, opened 2017-09-04
Refactoring

d4s:
Split single 'main' package.


sjoerdsimons:
Did a first skim, seems to be looking good. That said the actions files should probably be in package actions ? Also would be nice if you could be a big more verbose in your commit message to explain w hat is going where and the rationale behind it :). Also i assume you tested each commit to make sure there is no breakage at any point ?

d4s:
Fixes in new version. Squashed some commits.

PR #10, opened 2017-09-18
recipe: parse configuration in 'recipe' package

d4s:
Recipe configuration file parsing has been moved from main package of 'debos'
tool to package 'recipe'. This change combines all things related to
'debos' configuration file in the single package allowing to re-use parsing
in other tools and simplify tests creation.
Package 'recipe' is now responsible for templating mechanism and YAML parsing.

### DIFF
--- a/recipe/recipe.go
+++ b/recipe/recipe.go
@@ -1,9 +1,13 @@
 package recipe
 
 import (
+	"bytes"
+	"fmt"
 	"github.com/go-debos/debos"
 	"github.com/go-debos/debos/actions"
-	"log"
+	"gopkg.in/yaml.v2"
+	"path"
+	"text/template"
 )
 
 /* the YamlAction just embed the Action interface and implements the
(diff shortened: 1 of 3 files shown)


recipe/recipe_test.go
```diff
@@ -0,0 +1,182 @@
...
+	"testing"
+)
+
+type testRecipe struct {
+	recipe string
+	err    string
+}
+
+// Test if incorrect file has been passed
+func TestParse_file(t *testing.T) {
```
sjoerdsimons: TestParseErrors_ ? it's a bit of an odd naming

recipe/recipe_test.go
```diff
@@ -0,0 +1,182 @@
...
+	runTest(t, testSector)
+}
+
+func runTest(t *testing.T, test testRecipe, templateVars ...map[string]string) recipe.Recipe {
+	file, err := ioutil.TempFile(os.TempDir(), "recipe")
+	assert.Empty(t, err)
+	defer os.Remove(file.Name())
+
+	file.WriteString(test.recipe)
+	file.Close()
```
sjoerdsimons: Fwiw you can put test fixtures in a testdata subdirectory rather then putting them in code and writing them out

d4s: I thought about that on early stage.
Since there are not a lot of things to test it is useful to keep the samples with the code. From my point of view for sure.

recipe/recipe.go
```diff
@@ -1,3 +1,66 @@
+/*
+Package 'recipe' implements actions mapping to YAML recipe.
+
+Recipe syntax
+
+Recipe is composed with 3 conditional parts:
+
+- definitions and expressions for text template engine
+(https://golang.org/pkg/text/template)
```
sjoerdsimons: This is weird. the templating isn't a part of the recipe it's a pre-processing step; The recipe is a Yaml file that gets processed through the templating engine.

recipe/recipe.go
```diff
@@ -1,3 +1,66 @@
...
+
+Mandatory properties for receipt:
+
+- architecture -- target architecture
+
+- actions -- at least one action should be listed
+
+Supported actions
+
+- apt -- https://godoc.org/github.com/go-debos/debos/actions#hdr-Apt_Action
```
sjoerdsimons: Can we use relative urls in some way so you don't have to hardcode godoc.org ?

recipe/recipe.go
d4s: Does it matter for the end user?
Just imagine you know nothing about internals of debos and Go templating -- you'll see some strange things like '{{ var := value }}' in Yaml file -- which is a part of recipe because it is written in recipe file. May be it have sense to rephrase?

recipe/recipe.go
d4s: Not sure if it is possible. At least 'scheme://host' part is needed to show the link.

recipe/recipe_test.go
d4s: done

recipe/recipe.go
sjoerdsimons: saddest face, ah well

recipe/recipe.go
sjoerdsimons: I think templating is a quite well understood concept :) 

And yes it's mostly about rephrasing; I was quite confused when i read that paragraph, as i see the templating as just a method to generate the "actual" recipe file which has the requirements you describe further. 

  Having some basic examples of the go templating syntax is quite useful indeed. fwiw, we probalby want to document the `{{- $ospack := or .ospack "loqui-ospack.tar.gz" -}}` trick  for settings defaults as that's incredibly useful but also took some time to figure out :)

d4s: Hm... I think rephrasing should be a part of #14 in that case

PR #12, opened 2017-09-26
Rework of archiver for supporting different archive types

### LINKED ISSUE #8: Ability to unpack '.deb' packages to use files without package install

d4s:
Common approach for working with archives is introduced.
Any supported archive type must to satisfy 'Archiver' interface.

Example of file unpacking:
```
archive, _ := debos.NewArchive("file.ext")
archive.Unpack("destination")
```
   
Three archive types are supported:
- tar (with and without compression)
- zip
- debian packages

Fixes #8

### DIFF
--- a/actions/download_action.go
+++ b/actions/download_action.go
@@ -37,13 +37,58 @@ func (d *DownloadAction) validateUrl() (*url.URL, error) {
 	return url, nil
 }
 
+func (d *DownloadAction) validateFilename(context *debos.DebosContext, url *url.URL) (filename string, err error) {
+	if len(d.Filename) == 0 {
+		// Trying to guess the name from URL Path
+		filename = path.Base(url.Path)
+	} else {
+		filename = path.Base(d.Filename)
+	}
+	if len(filename) == 0 {
+		return "", fmt.Errorf("Incorrect filename is provided for '%s'", d.Url)
+	}
+	filename = path.Join(context.Scratchdir, filename)
+	return filename, nil
+}
+
+func (d *DownloadAction) archive(filename string) (debos.Archive, error) {
+	archive, err := debos.NewArchive(filename)
+	if err != nil {
+		return archive, err
+	}
+	switch archive.Type() {
+	case debos.Tar:
+		if len(d.Compression) > 0 {
+			if err := archive.AddOption("tarcompression", d.Compression); err != nil {
+				return archive, err
+			}
+		}
+	default:
+	}
+	return archive, nil
+}
+
 func (d *DownloadAction) Verify(context *debos.DebosContext) error {
+	var filename string
 
 	if len(d.Name) == 0 {
 		return fmt.Errorf("Property 'name' is mandatory for download action\n")
 	}
-	_, err := d.validateUrl()
-	return err
+
+	url, err := d.validateUrl()
+	if err != nil {
+		return err
+	}
+	filename, err = d.validateFilename(context, url)
+	if err != nil {
+		return err
+	}
+	if d.Unpack == true {
+		if _, err := d.archive(filename); err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 func (d *DownloadAction) Run(context *debos.DebosContext) error {
@@ -55,16 +100,10 @@ func (d *DownloadAction) Run(context *debos.DebosContext) error {
 		return err
 	}
 
-	if len(d.Filename) == 0 {
-		// Trying to guess the name from URL Path
-		filename = path.Base(url.Path)
-	} else {
-		filename = path.Base(d.Filename)
+	filename, err = d.validateFilename(context, url)
+	if err != nil {
+		return err
 	}
-	if len(filename) == 0 {
-		return fmt.Errorf("Incorrect filename is provided for '%s'", d.Url)
-	}
-	filename = path.Join(context.Scratchdir, filename)
 	originPath := filename
 
 	switch url.Scheme {
@@ -78,8 +117,13 @@ func (d *DownloadAction) Run(context *debos.DebosContext) error {
 	}
 
 	if d.Unpack == true {
+		archive, err := d.archive(filename)
+		if err != nil {
+			return err
+		}
+
 		targetdir := filename + ".d"
-		err := debos.UnpackTarArchive(filename, targetdir, d.Compression, "--no-same-owner", "--no-same-permissions")
+		err = archive.RelaxedUnpack(targetdir)
 		if err != nil {
 			return err
 		}
--- a/archiver.go
+++ b/archiver.go
@@ -1,33 +1,217 @@
 package debos
 
 import (
-	"log"
+	"fmt"
 	"os"
+	"path/filepath"
+	"strings"
 )
 
-func TarOptions(compression string) string {
+type ArchiveType int
+
+// Supported types
+const (
+	_ ArchiveType = iota // Guess archive type from file extension
+	Tar
+	Zip
+	Deb
+)
+
+type ArchiveBase struct {
+	file    string // Path to archive file
+	atype   ArchiveType
+	options map[interface{}]interface{} // Archiver-depending map with additional hints
+}
+type ArchiveTar struct {
+	ArchiveBase
+}
+type ArchiveZip struct {
+	ArchiveBase
+}
+type ArchiveDeb struct {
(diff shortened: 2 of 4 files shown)


archiver.go
```diff
@@ -1,33 +1,173 @@
...
+		return []string{}
+	}
+
+	return archiver.CommandUnpack(destination, options...)
+}
+
+func (arc *Archive) Unpack(destination string, options ...string) error {
+
+	command := arc.CommandUnpack(destination, options...)
+	//log.Printf("unpack: %s\n", command)
```
sjoerdsimons: Drop this one :)

archiver.go
```diff
@@ -1,33 +1,173 @@
...
+	return fmt.Errorf("SetCompression is not supported")
+}
+
+func (arc *Archive) GetType() ArchiveType { return arc.Type }
+
+/*
+CommandUnpack construct string for extracting archive file to destination.
+Not intended for direct usage -- mostly for internals and tests.
+*/
+func (arc *Archive) CommandUnpack(destination string, options ...string) []string {
```
sjoerdsimons: I'm confused why this needs to be implemented.

archiver.go
d4s: Done

archiver.go
d4s: very useful for unit tests. And to have only one 'Unpack' implementation since the difference is in command line for supported archive types.

archiver.go
```diff
@@ -1,33 +1,173 @@
...
+			atype = Zip
+		default:
+			//FIXME: guess Tar maybe?
+			atype = Tar
+		}
+	} else {
+		atype = arcType[0]
+	}
+
+	common := Archive{}
```
sjoerdsimons: Why create this first rather then first creating the relative concrete implementation and then setting the values

archiver.go
```diff
@@ -1,33 +1,172 @@
...
+
+type Archive struct {
+	File        string // Path to archive file
+	Type        ArchiveType
+	compression string
+}
+
+type Archiver interface {
+	CommandUnpack(destination string, options ...string) []string
+	GetType() ArchiveType
```
sjoerdsimons: Fwiw getters prefixed with Get isn't iodiomatic go. Also shouldn't the interface be Unpacker rahter then Archiver ? it doesn't archive anything :)

archiver.go
sjoerdsimons: You can still hvae the central Unpack implementation in your base class even without this fucntion so that's not relevant.

I'll check the usage in unit tests in a bit to see how it gets used, but it seems very odd to me thusfar

archiver.go
d4s: May be I am not so good in Go yet ;) But from my findings I'll need a reflection in other case. Or do the same things for any supported

actions/download_action.go
```diff
@@ -78,8 +78,23 @@ func (d *DownloadAction) Run(context *debos.DebosContext) error {
...
+		if len(d.Compression) > 0 {
+			if err := archive.SetCompression(d.Compression); err != nil {
+				return err
+			}
+		}
 		targetdir := filename + ".d"
-		err := debos.UnpackTarArchive(filename, targetdir, d.Compression, "--no-same-owner", "--no-same-permissions")
+		switch archive.GetType() {
+		case debos.Tar:
+			err = archive.Unpack(targetdir, "--no-same-owner", "--no-same-permissions")
```
sjoerdsimons: Hrm, this requires the user to know that underlying to it all is unix tar. It would be better to have give the Tar archive class a method to turn of the  usage of same persmissions/owner and call that before calling unpack in gneeral.

archiver.go
d4s: Yet. But if we extend the interface with Pack() -- it would be.
Let me know if Unpacker is still preferable.

archiver.go
```diff
@@ -1,33 +1,172 @@
...
+	SetCompression(compression string) error
+	Unpack(destination string, options ...string) error
```
sjoerdsimons: Allowing the caller to add extra options to the unpack command breaks your abstraction. There is no reason for the user to know underneath just unix commands are being called especailly as this may change at some point.

archiver.go
```diff
@@ -1,33 +1,172 @@
...
+)
+
+type Archive struct {
+	File        string // Path to archive file
+	Type        ArchiveType
+	compression string
+}
+
+type Archiver interface {
+	CommandUnpack(destination string, options ...string) []string
```
sjoerdsimons: CommandUnpack should not be part of the interface if your interface users aren't meant to call it. Implment it onthe base archive class instead (same as a common unpack).

archiver_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// New archive
+	// Expect Tar by default
+	archive, err := debos.NewArchive("test.tar.gz")
+	assert.NotEmpty(t, archive)
+	assert.Empty(t, err)
+
+	// Type must be Tar by default
+	arcType := archive.GetType()
+	assert.Equal(t, debos.Tar, arcType)
```
sjoerdsimons: May be useful to also do a typeassertion here

archiver_test.go
```diff
@@ -0,0 +1,128 @@
...
+	// Guess Deb
+	archive, err := debos.NewArchive("test.deb")
+	assert.NotEmpty(t, archive)
+	assert.Empty(t, err)
+
+	// Type must be guessed as Deb
+	arcType := archive.GetType()
+	assert.Equal(t, debos.Deb, arcType)
+
+	// Force Zip type
```
sjoerdsimons: copy & paste error?

archiver.go
```diff
@@ -1,33 +1,172 @@
...
+	GetType() ArchiveType
+	SetCompression(compression string) error
```
sjoerdsimons: Does SetCompression make sense in the interface? seems only useful for the Tar type?

archiver.go
sjoerdsimons: I think it would be more idiomatic go if you then had a Pack interface next to the Unpack one; Like there are diferent interfaces for Reader and Writer

archiver.go
sjoerdsimons: Yeah, i don't see in the unit test commit how this is useful. It looks like you're only testing explicit implementations so this isn't needed; So the code in here should never be used it can just raise and exception

archiver.go
d4s: But they are. Test is the user for example.
The main idea was to have own method for command line for each archive type, so  any new type has it's own implementation.
In case if base type will deal with all types of archive -- it have no sense to keep all other types.

archiver.go
sjoerdsimons: Yes you need to first work out hwihc type to instantiate indeed. youcan just do
`switch atype {
  case Tar:
   archiver = &ArchiveTar{}
  ...
}
archiver.File = file
...
`

archiver.go
sjoerdsimons: No what i mean is that there is no point of the *interface* which is exposed to the user of the code to have the method. 

The Tests can do a type assertion to get at the specific implementation of a type if they want to test CommandUnpack.

Note that this doesn't mean you need a commaon CommandUnpack, see for example how the BaseAction has some dummy functions which get overriden in the Specific actions to do the specialisations.

actions/download_action.go
d4s: hm... good point. Do you think it have sense to remove possibility to extend command line at all?

archiver_test.go
d4s: Yep :(

archiver.go
d4s: I can't -- archiver is an interface 'Archiver', so: "archiver.File undefined (type Archiver has no field or method File)"
PS it also was my first assumption to use write the code in that way

archiver.go
d4s: Hm... yes.

archiver.go
sjoerdsimons: Oh crap yes, you're right. Guess you can only do it via a type assertion then if you want to go that route. Anyway this is fine, i just thought it come be more elegant :)

archiver_test.go
d4s: Not need due rework. Check Archive structure.

PR #6, opened 2017-09-15
 actions: add description of actions 

d4s:
Add description of each available action and properties allowed by actions.
Used 'godoc' format to keep actions documentation in one place with implementation.

PR #26, opened 2017-10-06
actions: debootstrap: add 'merged-usr' property

### LINKED ISSUE #17: Make the --merged-usr option for debootstrap configurable
Some old packages (`libpng12-0_1.2.54-1ubuntu1`, I'm looking at you) return puzzling errors with a merged `/usr`.

To make `debos` usable in those cases it should be possible to disable passing the `--merged-user` option to `deboostrap`.

d4s:
Add new property 'merged-usr' for 'debootstrap' action.
By default it is set to 'true' for usage filesystem with
merged '/usr'. If set to 'false' then classic FHS will be used.

Fixes #17.

### DIFF
--- a/actions/debootstrap_action.go
+++ b/actions/debootstrap_action.go
@@ -17,6 +17,14 @@ type DebootstrapAction struct {
 	Variant        string
 	KeyringPackage string `yaml:"keyring-package"`
 	Components     []string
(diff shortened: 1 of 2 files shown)


d4s:
Blocked by #6 -- need to extend the description of debootstrap action.

PR #31, opened 2017-10-13
debug: run interactive shell on error

### LINKED ISSUE #30: Fall into interactive debug shell if action fails

d4s:
Launch interactive shell in context of failed action for issues investigation and debug.

Add options to debos:
--debug-shell -- to start a shell in case of any failure
--shell <path> -- use selected shell instead of default bash

Fixes #30 

### DIFF
--- a/cmd/debos/debos.go
+++ b/cmd/debos/debos.go
@@ -13,12 +13,14 @@ import (
 	"github.com/sjoerdsimons/fakemachine"
 )
 
-func bailOnError(err error, a debos.Action, stage string) {
+func bailOnError(context debos.DebosContext, err error, a debos.Action, stage string) {
 	if err == nil {
 		return
 	}
 
-	log.Fatalf("Action `%s` failed at stage %s, error: %s", a, stage, err)
+	log.Printf("Action `%s` failed at stage %s, error: %s", a, stage, err)
+	debos.DebugShell(context)
+	os.Exit(1)
 }
 
 func main() {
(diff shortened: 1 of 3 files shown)


cmd/debos/debos.go
```diff
@@ -13,12 +13,14 @@ import (
...
-func bailOnError(err error, a debos.Action, stage string) {
+func bailOnError(context debos.DebosContext, err error, a debos.Action, stage string) {
 	if err == nil {
 		return
 	}
 
-	log.Fatalf("Action `%s` failed at stage %s, error: %s", a, stage, err)
+	log.Printf("Action `%s` failed at stage %s, error: %s", a, stage, err)
+	debos.DebugShell(context)
+	os.Exit(1)
```
sjoerdsimons: Should this now be changed away from calling Exit() directly similar to  #33  ?

cmd/debos/debos.go
```diff
@@ -13,12 +13,9 @@ import (
 	"github.com/sjoerdsimons/fakemachine"
 )
 
-func bailOnError(err error, a debos.Action, stage string) {
-	if err == nil {
-		return
-	}
-
-	log.Fatalf("Action `%s` failed at stage %s, error: %s", a, stage, err)
+func bailOnError(context debos.DebosContext, err error, a debos.Action, stage string) {
```
sjoerdsimons: This is mostly a nitpick; but the bailOnError is no longer bailing on error, so should really have a different name now.

also wondering if you could make it set the exitcode rather then your error block having to be 3 lines each time, but i don't really see a elegant solution for that unless you pass a reference to exitcode and make it say (completely untested):
```go
if checkError (err, a, stage, *exitcode) {
  return
}
```

I guess it does save a bit of redundant typing.

d4s: hm... I thought about something like this but it is no so elegant as I wanted to have.
The most elegant solution would be to get rid of functional part from main() to other function. Something like:
```Go
func do_stuff() int {
  defer func(){ /*cleanup*/}()
   // the most of the functional code here
  if err != nil { return 1 }
  return 0
}
func main() {
 os.Exit(do_stuff())
}
```

PR #32, opened 2017-10-16
commands: disable services for commands running in chroot

d4s:
Disabled services start/stop for commands running in chroot.
This also prevents services from start during packages install.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -125,6 +125,13 @@ func (cmd Command) Run(label string, cmdline ...string) error {
 		exe.Env = append(os.Environ(), cmd.extraEnv...)
 	}
 
+	// Disable services start/stop for commands running in chroot
+	if cmd.ChrootMethod != CHROOT_METHOD_NONE {
+		services := ServiceHelper{cmd.Chroot}
+		services.Deny()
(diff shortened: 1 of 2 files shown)


d4s:
Removed option from 'apt' action.

commands.go
```diff
@@ -125,6 +125,12 @@ func (cmd Command) Run(label string, cmdline ...string) error {
 		exe.Env = append(os.Environ(), cmd.extraEnv...)
 	}
 
+	// Disable services start/stop for commands running in chroot
+	if cmd.ChrootMethod != CHROOT_METHOD_NONE {
+		ServicePolicyHelper(cmd.Chroot, DenyServices)
```
sjoerdsimons: Given the only usage of the Helper is to deny things; seems more elegant to follow the pattern of the qemuHelper; iotw instatiate an object to Deny services and defer a Cleanup function on it. Or if you prefer do something like:
```go
s := ServicePolicyHelper(cmd.chroot)
s.Deny()
defer s.Allow()
```
In your current code your one function is just a dispatch for two different functions essentially :smile:

PR #33, opened 2017-10-20
Allow to finish all deferred functions on os.Exit()

d4s:
os.Exit() call do not run deferred functions.
This leads to incorrect cleanup of resources and files in certain cases.

### DIFF
--- a/cmd/debos/debos.go
+++ b/cmd/debos/debos.go
@@ -29,21 +29,30 @@ func main() {
 		TemplateVars  map[string]string `short:"t" long:"template-var" description:"Template variables"`
 	}
 
+	var exitcode int = 0
+	// Allow to run all deferred calls prior to os.Exit()
+	defer func() {
+		os.Exit(exitcode)
+	}()
+
 	parser := flags.NewParser(&options, flags.Default)
 	args, err := parser.Parse()
 
 	if err != nil {
 		flagsErr, ok := err.(*flags.Error)
 		if ok && flagsErr.Type == flags.ErrHelp {
-			os.Exit(0)
+			return
 		} else {
 			fmt.Printf("%v\n", flagsErr)
-			os.Exit(1)
+			exitcode = 1
+			return
 		}
 	}
 
 	if len(args) != 1 {
-		log.Fatal("No recipe given!")
+		log.Println("No recipe given!")
+		exitcode = 1
+		return
 	}
 
 	file := args[0]
@@ -108,10 +117,10 @@ func main() {
 			bailOnError(err, a, "PreMachine")
 		}
 
-		ret := m.RunInMachineWithArgs(args)
+		exitcode = m.RunInMachineWithArgs(args)
 
-		if ret != 0 {
-			os.Exit(ret)
+		if exitcode != 0 {
+			return
 		}
 
 		for _, a := range r.Actions {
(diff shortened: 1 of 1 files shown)


cmd/debos/debos.go
```diff
@@ -29,21 +37,32 @@ func main() {
 		TemplateVars  map[string]string `short:"t" long:"template-var" description:"Template variables"`
 	}
 
+	// Allow to run all deffer calls prior to os.Exit()
+	defer func() {
+		switch p := recover(); p {
+		case code0, code1:
+			os.Exit(p.(int))
```
em-: I found this use of `panic()` even in successful cases a bit surprising: is this somehting idiomatic in golang?

cmd/debos/debos.go
```diff
@@ -29,21 +37,32 @@ func main() {
...
+		default:
+			panic(p)
+		}
```
em-: Should we add a match for `nil` so we don't end up calling `panic(nil)` when no `panic()` has otherwise been called?

cmd/debos/debos.go
```diff
@@ -18,9 +18,17 @@ func bailOnError(err error, a debos.Action, stage string) {
...
-	log.Fatalf("Action `%s` failed at stage %s, error: %s", a, stage, err)
+	log.Printf("Action `%s` failed at stage %s, error: %s", a, stage, err)
+	panic(code1)
 }
 
+type exitcode uint
+
+const (
+	code0 exitcode = 0
+	code1          = 1
```
em-: I'm not sure we need these, but I'd at least try to find more meaningful names here.

cmd/debos/debos.go
```diff
@@ -111,7 +130,7 @@ func main() {
 		ret := m.RunInMachineWithArgs(args)
 
 		if ret != 0 {
-			os.Exit(ret)
+			panic(code1)
```
em-: Mh, is `RunInMachineWithArgs()` returning the exit code of the launched command?

Is it worth communicating it back untranslated to the `debos` caller as the previous code did, or it would just be confusing?

cmd/debos/debos.go
```diff
@@ -29,21 +37,32 @@ func main() {
...
+	}()
+
 	parser := flags.NewParser(&options, flags.Default)
 	args, err := parser.Parse()
 
 	if err != nil {
 		flagsErr, ok := err.(*flags.Error)
 		if ok && flagsErr.Type == flags.ErrHelp {
-			os.Exit(0)
+			panic(code0)
```
em-: Isn't a `return` enough here?

cmd/debos/debos.go
```diff
@@ -120,7 +139,7 @@ func main() {
 		}
 
 		log.Printf("==== Recipe done ====")
-		os.Exit(0)
+		panic(code0)
```
em-: I'm really not a fan of using `panic()` when there's no error.

Either we use a bare `return` here, relying on the fact that the process exitcode is `0` unless `os.Exit()` is called, or we [add a wrapper around `main()](https://github.com/anacrolix/torrent/blob/master/cmd/torrentfs/main.go#L72), which would probably also be a good place where to put the `recover()` deferred function.

cmd/debos/debos.go
d4s: In fact no. Rethinked panic() usage and use different approach.

cmd/debos/debos.go
d4s: Useful for panic( but do not need it anymore

PR #34, opened 2017-10-26
debos: allow to use external image for '/scratch'

### LINKED ISSUE #29: Allow to use disk image on /scratch instead of tmpfs-based
Currently have 'Out of memory' for medium and large sized OS.

d4s:
Create temporary image to mount to '/scratch' directory.
Fixes problems with preparing OS packs requires a lot of disk space for
packages.

Fixes #29

### DIFF
--- a/cmd/debos/debos.go
+++ b/cmd/debos/debos.go
@@ -6,7 +6,9 @@ import (
 	"log"
 	"os"
 	"path"
+	"path/filepath"
 
+	"github.com/docker/go-units"
 	"github.com/go-debos/debos"
 	"github.com/go-debos/debos/recipe"
 	"github.com/jessevdk/go-flags"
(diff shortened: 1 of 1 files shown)


sjoerdsimons:
Uhm no;  This is broken in a variaty of ways..

Maybe most importantly you should prepare the space image file *outside* of the fakemachine and pass it into the fakemachine as an image so it's a normal filesystem, but more importantly such that you don't go through the 9p protocol to write data to it..

I also think the size is key here rather then the location (which should be optional).

Though more generally i think a lot of this should live in fakemachine as it's generally useful to get a "bigger then memory" scratch space.

d4s:
In addition: current commit breaks work of action 'image-partition' since it has a hardcoded image name '/dev/vda'

sjoerdsimons:
Closing this one, #39 has a better solution for this (though needs some other branches to be merged first :)

PR #37, opened 2017-10-28
actions: image-partition: add alias 'vfat' for 'fat32'

d4s:
'vfat' filesystem type is widespread name for FAT FS in *nix-world.
Add alias 'vfat' to 'fat32' filesystem type.

Add support of partitions creation without filesystem.

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -302,6 +302,10 @@ func (i ImagePartitionAction) Run(context *debos.DebosContext) error {
 			}
 		}
 
+		// Give a chance for udevd to create proper symlinks
+		debos.Command{}.Run("udevadm", "udevadm", "settle", "-t", "5",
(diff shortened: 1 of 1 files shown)


actions/image_partition_action.go
```diff
@@ -45,7 +45,10 @@ configuration (below) and label the filesystem located on this partition.
 
 - fs -- filesystem type used for formatting.
 
-NB: the FAT (vfat) filesystem is named 'fat32' in configuration file.
+'none' fs type should be used for partition without filesystem.
+
+NB: the FAT (vfat) filesystem have an alias for 'fat32' in configuration file
```
sjoerdsimons: *has an alias

Also i don't really see how that's relevant for the user; The user can use either fat32 or vfat as both names tend to be used.

d4s:
Moved udevadm call to dedicated commit.

PR #40, opened 2017-10-31
actions: image-partition: allow to use disks by ID

d4s:
Naming scheme by disk ID provided by udevd uses '-part' suffix for partitions.
Allow to use correct partition names in case if disk path is
'/dev/disk/by-id/'.


d4s:
Unfortunately PR #39 includes #38 as well.
So merge breaks 'image-partition' action. Current commit restore functionality.

PR #41, opened 2017-11-01
README.md: add links to actions and examples

zumbi:
Add useful links to documentation and examples to use the tool.

Signed-off-by: Héctor Orón Martínez <zumbi@debian.org>


zumbi:
Updated

d4s:
Not sure if second commit is valid for debos itself -- we have no any action based on 'bmaptool'.
In recipe from example 'bmaptool' called directly with action 'Run' meaning we may call any tool we want -- even 'dd' for copying onto SD-card =)

zumbi:
Right, however I added that bit in case someone is missing why bmap file is created and how it is used

PR #42, opened 2017-11-07
actions: run: allow to start scripts with arguments

d4s:
Scripts passed to "Run" action are strings containing full path to script
with arguments. Debos fails to run it tr to start the whole string as a script.

Add possibility to start scripts with additional arguments.
Correct mount points for scripts with arguments in chrooted environment.

### DIFF
--- a/actions/run_action.go
+++ b/actions/run_action.go
@@ -34,9 +34,9 @@ package actions
 
 import (
 	"errors"
-	"fmt"
 	"github.com/go-debos/fakemachine"
 	"path"
+	"strings"
 
 	"github.com/go-debos/debos"
 )
@@ -64,8 +64,11 @@ func (run *RunAction) PreMachine(context *debos.DebosContext, m *fakemachine.Mac
 	}
 
 	run.Script = debos.CleanPathAt(run.Script, context.RecipeDir)
+	// Expect we have no blank spaces in path
+	scriptpath := strings.Split(run.Script, " ")
+
 	if !run.PostProcess {
-		m.AddVolume(path.Dir(run.Script))
+		m.AddVolume(path.Dir(scriptpath[0]))
 	}
 
 	return nil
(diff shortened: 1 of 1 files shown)


actions/run_action.go
```diff
@@ -86,14 +90,19 @@ func (run *RunAction) doRun(context debos.DebosContext) error {
...
 		if run.Chroot {
-			cmd.AddBindMount(path.Dir(run.Script), "/script")
-			cmdline = []string{fmt.Sprintf("/script/%s", path.Base(run.Script))}
+			scriptpath, err := debos.GetCommandFromString(run.Script)
+			if err != nil {
+				return err
+			}
+			scriptpath = path.Dir(scriptpath)
+			cmd.AddBindMount(scriptpath, "/script")
+			cmdline = []string{strings.Replace(run.Script, scriptpath, "/script", 1)}
```
sjoerdsimons: That will break if the directory name of the script and some other parameters overlap; please make sure you only replace the script name not the full command line

actions/run_action.go
```diff
@@ -86,14 +90,19 @@ func (run *RunAction) doRun(context debos.DebosContext) error {
...
 		} else {
 			cmdline = []string{run.Script}
 		}
 		label = path.Base(run.Script)
 	} else {
-		cmdline = []string{"sh", "-c", run.Command}
+		cmdline = []string{run.Command}
```
sjoerdsimons: Why this change?

commands.go
```diff
@@ -66,6 +67,26 @@ func (w *commandWrapper) flush() {
 	w.out(true)
 }
 
+/*
+GetCommandFromString extracts command path from single string
+containing full path to command and arguments.
+*/
+func GetCommandFromString(command string) (string, error) {
```
sjoerdsimons: Unless go has helpers just get the first word from the string rather then trying to be clever. The support for commands with spaces in them is impossible to end well (unless you enforce escaping).

commands.go
```diff
@@ -111,8 +130,12 @@ func (cmd Command) Run(label string, cmdline ...string) error {
 			options = append(options, "--bind", b)
 
 		}
-		options = append(options, cmdline...)
 	}
+	if len(cmdline) == 1 {
+		// Single command OR command with options passed as single string
+		options = append(options, "sh", "-c")
```
sjoerdsimons: I don't think this class should try to guess if it should use sh -c, the caller should do that.

actions/run_action.go
d4s: There is a limit '1' for replacement in `strings.Replace(..,..,1)`

actions/run_action.go
d4s: we need the same for scripts. Added shell below.

PR #54, opened 2017-12-07
Ostree deploy

d4s:
Allow to use ostree_deploy action without the target image.

### DIFF
--- a/actions/ostree_deploy_action.go
+++ b/actions/ostree_deploy_action.go
@@ -73,7 +73,7 @@ func (ot *OstreeDeployAction) setupFSTab(deployment *ostree.Deployment, context 
 	deploymentDir := fmt.Sprintf("ostree/deploy/%s/deploy/%s.%d",
 		deployment.Osname(), deployment.Csum(), deployment.Deployserial())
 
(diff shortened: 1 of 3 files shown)


actions/ostree_deploy_action.go
```diff
@@ -93,18 +93,21 @@ func (ot *OstreeDeployAction) setupFSTab(deployment *ostree.Deployment, context
 
 func (ot *OstreeDeployAction) Run(context *debos.DebosContext) error {
 	ot.LogStart()
-	/* First deploy the current rootdir to the image so it can seed e.g.
-	 * bootloader configuration */
-	err := debos.Command{}.Run("Deploy to image", "cp", "-a", context.Rootdir+"/.", context.ImageMntDir)
-	if err != nil {
-		return fmt.Errorf("rootfs deploy failed: %v", err)
+
+	if _, err := os.Stat(context.ImageMntDir); err == nil {
```
sjoerdsimons: Could use a comment explaining this is to handle cases where we didn't partition an image. Do we need to run os.Stat here though rather then simply checking if ImageMntDir is set (e.g. non-""), If ImageMntDir is set but doesn't exist we have bigger problems tbh

d4s: done

PR #55, opened 2017-12-13
Support specifying a partition in the raw action

sjoerdsimons:


actions/raw_action.go
```diff
@@ -93,9 +96,29 @@ func (raw *RawAction) Run(context *debos.DebosContext) error {
...
+		if devicePath == "" {
+			return fmt.Errorf("Failed to find partition named %s", raw.Partition)
+		}
+
+		target, err = os.OpenFile(devicePath, os.O_WRONLY, 0)
+		if err != nil {
+			return fmt.Errorf("Failed to open partition file %v", err)
+		}
+	} else {
+		target, err = os.OpenFile(context.Image, os.O_WRONLY, 0)
```
d4s: I think this should be rewritten a bit since it is almost the same as 'target' initialization above for `devicePath`.

PR #57, opened 2017-12-21
Bind mount disks for command

sjoerdsimons:


commands.go
```diff
@@ -66,8 +67,37 @@ func (w *commandWrapper) flush() {
 	w.out(true)
 }
 
-func NewChrootCommand(chroot, architecture string) Command {
-	return Command{Architecture: architecture, Chroot: chroot, ChrootMethod: CHROOT_METHOD_NSPAWN}
+func realpath(path string) (string, error) {
```
d4s: realpath() should be in filesystem.go

sjoerdsimons: done

PR #63, opened 2018-01-29
minor TODO update

stappersg:
There is a version 1.0 release out and repository has been renamed.

PR #66, opened 2018-03-01
v1.0 is out, already renamed

stappersg:
now is the TODO list aware of it


stappersg:
Probably I was to quick with closing #63

PR #68, opened 2018-03-07
Support authenticated and unauthenticated repositories

zumbi:


zumbi:
Allow me a bit more of testing

PR #71, opened 2018-03-07
debootstrap: allow repository verification

zumbi:
By default, debos does not verify repositories. This change allows
to pass keyring option to debootstrap, so it can verify signed
repositories. If no keyring file is passed, it fallbacks to previous
state were it does not check repository authenticity.

This change does not resolve how to pass keyring file within
fakemachine environment. Using an overlay previous to debootstrap
run causes overlay to be removed.

Signed-off-by: Héctor Orón Martínez <hector.oron@collabora.co.uk>

### DIFF
--- a/actions/debootstrap_action.go
+++ b/actions/debootstrap_action.go
@@ -10,6 +10,7 @@ Yaml syntax:
    components: <list of components>
    variant: "name"
    keyring-package:
+   keyring-file:
 
 Mandatory properties:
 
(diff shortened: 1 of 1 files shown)


smcv:
I've tried to fix what @sjoerdsimons' review said on #86.

> So imho that should be a seperate option; I'm on the fence whether it should default to true or false.

I used `check-gpg`, default true, but it would be very easy to swap it to false.

smcv:
The one thing that was previously in this PR, but isn't resolved by #86, is the addition of an `allow-unauthenticated` flag to the `apt` action.

If there is a real use-case for that, please open a separate PR for it. (I can't think of a good use-case, though; if you have repositories that are unsigned and you're relying on implicit authentication via an assumed-to-be-secure LAN or HTTPS or something, having them appear in `sources.list` like `deb [trusted=yes] https://mirror suite component` is usually a better approach.)

PR #76, opened 2018-04-08
Pack action: Add support for xz and bzip2

ana:
Pack action must still be rewritten in pure go. In the meantime, this patch allows to have quickly support for bzip2 and xz.

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -388,6 +388,19 @@ func (i ImagePartitionAction) Cleanup(context *debos.DebosContext) error {
 	return nil
 }
 
+func (i ImagePartitionAction) PostMachineCleanup(context *debos.DebosContext) error {
+	image := path.Join(context.Artifactdir, i.ImageName)
+	/* Remove the image in case of any action failure */
+	if context.State != debos.Success {
+		if _, err := os.Stat(image); !os.IsNotExist(err) {
+			if err = os.Remove(image); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
 func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 	if len(i.GptGap) > 0 {
 		log.Println("WARNING: special version of parted is needed for 'gpt_gap' option")


actions/pack_action.go
```diff
@@ -30,10 +30,32 @@ type PackAction struct {
...
+		}
+	}
+
+	return nil
+}
+
 func (pf *PackAction) Run(context *debos.DebosContext) error {
 	pf.LogStart()
 	outfile := path.Join(context.Artifactdir, pf.File)
+	opt := map[string]string{
```
d4s: Please use map only for compression option instead of all passed options.
Also some versions of `tar` would be confused by option 'f' following the other option instead of file name.

actions/pack_action.go
```diff
@@ -30,10 +30,32 @@ type PackAction struct {
 	File             string
 }
 
+
+func (pf *PackAction) Verify(context *debos.DebosContext) error {
+
+	if len(pf.File) == 0 {
```
d4s: Please leave the `gz` as default option

actions/pack_action.go
```diff
@@ -30,10 +30,32 @@ type PackAction struct {
...
+		return fmt.Errorf("Please add a 'file' property.")
+	}
+
+	if len(pf.Compression) > 0 {
+		if pf.Compression != "gz" && pf.Compression != "xz" &&  pf.Compression != "bzip2" {
```
d4s: here you need to use a map for compression type as well and walk instead of just comparing with names.
Please put the map somewhere outside the functions, so it could be used in both places.

d4s: Please leave the gz as default option

ana:
The third commit should address all the point listed in the review.

sjoerdsimons:
@ana Please squash your changes as appropriate. There is no point in having the master history contain patches that reworked previous patches in the same merge request, that adds no value.

Also please work on your commit messages; Having commits with no commit body message is always a bad sign.

actions/pack_action.go
```diff
@@ -30,10 +31,38 @@ type PackAction struct {
...
 }
 
+
+func (pf *PackAction) Verify(context *debos.DebosContext) error {
+
+	if len(pf.File) == 0 {
+		return fmt.Errorf("Please add a 'file' property.")
+	}
+
+	archive, err := debos.NewArchive(pf.File)
```
sjoerdsimons: Would make sense to add a NewTarArchive(compression) function rather then using NewArchive (which mostly looks at which type of archive it should be).

actions/pack_action.go
```diff
@@ -30,10 +31,38 @@ type PackAction struct {
...
+	return nil
+}
+
 func (pf *PackAction) Run(context *debos.DebosContext) error {
 	pf.LogStart()
 	outfile := path.Join(context.Artifactdir, pf.File)
 
-	log.Printf("Compression to %s\n", outfile)
-	return debos.Command{}.Run("Packing", "tar", "czf", outfile, "-C", context.Rootdir, ".")
+	archive, err := debos.NewArchive(outfile)
```
sjoerdsimons: Same as a above :)

archiver.go
```diff
@@ -172,6 +177,34 @@ func (deb *ArchiveDeb) RelaxedUnpack(destination string) error {
 	return deb.Unpack(destination)
 }
 
+// Helper function for packing with external tool
+func pack(command []string) error {
+	return Command{}.Run("Pack", command...)
+}
+
+func (arc *ArchiveBase) Pack(destination string) error {
```
sjoerdsimons: Please group this with the other ArchiveBase functions

archiver.go
```diff
@@ -172,6 +177,34 @@ func (deb *ArchiveDeb) RelaxedUnpack(destination string) error {
 	return deb.Unpack(destination)
 }
 
+// Helper function for packing with external tool
+func pack(command []string) error {
+	return Command{}.Run("Pack", command...)
```
sjoerdsimons: Any reason to not just do this inline  as it's only called from one place

archiver.go
```diff
@@ -172,6 +177,34 @@ func (deb *ArchiveDeb) RelaxedUnpack(destination string) error {
...
+
+func (arc *ArchiveBase) Pack(destination string) error {
+	return fmt.Errorf("Pack is not supported for '%s'", arc.file)
+}
+
+func (tar *ArchiveTar) Pack(destination string) error {
+	command := []string{"tar"}
+	command = append(command, "-c")
+
+	if compression, ok := tar.options["tarcompression"]; ok {
```
sjoerdsimons: Somewhat orthogonally to your commit i really dislike the base archiver class having an options map that then gets abused with subclass specific strings.. It would be far nicer for the ArchiveTar struct to simply have fields for these things directly.  But not a blocker for this merge request :)

archiver.go
```diff
@@ -172,6 +177,34 @@ func (deb *ArchiveDeb) RelaxedUnpack(destination string) error {
...
+		if unpackTarOpt := tarOptions(compression.(string)); len(unpackTarOpt) > 0 {
+			command = append(command, unpackTarOpt)
+		} else {
+			// defaults to "gz"
+			command = append(command, "-z")
```
sjoerdsimons: This doesn't default to gz; this uses gz as a fallback e.g. if compression for some reason gets set to lzo4 you end up  compressing as gz.

archiver.go
sjoerdsimons: If compression wasn't set it needs to default to gz in this case; Though it'd be nicer if one could rely on tarcmpression to always be set (or really just a filed in TarArchive) and have it default to gz.

PR #79, opened 2018-05-18
Fix details

elboulangero:
A few fixes I made while getting started with debos. Just details..

### DIFF
--- a/filesystem.go
+++ b/filesystem.go
@@ -48,7 +48,13 @@ func CopyFile(src, dst string, mode os.FileMode) error {
 		os.Remove(tmp.Name())
 		return err
 	}
-	return os.Rename(tmp.Name(), dst)
+
+	if err = os.Rename(tmp.Name(), dst); err != nil {
+		os.Remove(tmp.Name())
+		return err
+	}
+
+	return nil
 }
 
 func CopyTree(sourcetree, desttree string) error {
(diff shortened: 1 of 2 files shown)


elboulangero:
Thanks for the link, I gave it a try and ended up with simpler commit messages, hope you like it ;)

elboulangero:
Any news on that? Is there anything else you want to change?

sjoerdsimons:
Did your ocmmit messages actually change?

elboulangero:
Yes, I don't know if there's a way to see that in GitLab though.

sjoerdsimons:
Well even https://github.com/elboulangero/debos/tree/fix-details still has commit messages with only a subject so...

elboulangero:
OK I didn't really understand that it was about more detailed commit messages. I pushed again a revised version. Thanks.

filesystem.go
```diff
@@ -63,7 +69,10 @@ func CopyTree(sourcetree, desttree string) error {
 		target := path.Join(desttree, suffix)
 		switch info.Mode() & os.ModeType {
 		case 0:
-			CopyFile(p, target, info.Mode())
+			err := CopyFile(p, target, info.Mode())
+			if err != nil {
+				log.Panicf("Failed to copy file %s: %v", p, err)
```
sjoerdsimons: This should return err instead of panicking; potentially extending the error message to make it more clear what when wrong

Issue #81, opened 2018-05-30
YAML comments should be ignored when executing the templates


elboulangero:
Consider the following snippet in a recipe:

```
{{ $kver := "4.9.0" }}
#{{ $kver := "4.16.0" }}
```
Currently the value of kver will be set to `4.16.0`, since the Golang templating engine is not aware of the YAML syntax. But this is definitely not what an user expects, comments should be ignored.

gctucker:
Use Go template comments maybe?
https://golang.org/pkg/text/template/#hdr-Actions

gctucker:
`#` is not a comment in the Go template language, but this `{{/* */}}` is:
```
# The line below should not appear in the YAML output:
{{/* {{ $arch := "riscv" }} */}}
```

I've verified it with a [debos recipe](https://gitlab.collabora.com/gtucker/jenkins-debian/commit/bb8e63847de345c64d15c591ddbb99b9e4c6971d#cd98844e25dd40239b1a7a14d31531e896e2a9ae_4_4).

elboulangero:
@gtucker, thanks for the example. Actually I understand that I can use Go template comments, but that's not the point I want to raise. I opened this issue because I care about the usability of the tool.

It appears than recipe files are both YAML format **AND** Go template language at the same time (smcv said it better above). So when you look at a debos recipe, what do you see? Maybe some users see a YAML file, some other users see Go templating at work. Some educated users see both and know how it works under the hood, but you can't expect that from new users.

Personally, the first debos recipes I've seen looked and behaved very much like a YAML file, and I had absolutely no idea what Go templating was. So I used it like a YAML file, and it was just fine, until I happened to comment a line containing a template, and then I lost an hour or two figuring out what the hell was going on. Then I became a bit more of an educated user :)

Every new user is different, sure, but I think that there's something confusing that could be improved here.

smcv:
> It appears than recipe files are both YAML format **AND** Go template language at the same time

No, they are a Go template that *generates* YAML. The difference between that and what I quoted is important.

You can't write a machine-readable language correctly unless you know what it is. It's the same as how you need to know whether you are writing Javascript or Javascript-embedded-in-HTML before you can write a less-than comparison using `<` (if in Javascript) or `&lt;` (if in Javascript embedded in HTML); if you get it wrong (in either direction!) then your code won't work.

If new users think the recipes are YAML, the solution is for the documentation to be clearer about the recipes being Go templates that generate YAML.

[...]

PR #82, opened 2018-05-30
Remove YAML comments before template substitution

elboulangero:
This fix the issue #81 

Please not that I have no skill with regex and no skill with go, and this commit uses both ;) Use at your own risk (but it works on my machine!)

Also we should probably add a test case.

Cheers

### DIFF
--- a/recipe/recipe.go
+++ b/recipe/recipe.go
@@ -69,7 +69,9 @@ import (
 	"github.com/go-debos/debos"
 	"github.com/go-debos/debos/actions"
 	"gopkg.in/yaml.v2"
+	"io/ioutil"
 	"path"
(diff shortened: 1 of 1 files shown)


elboulangero:
Hmm, however this patch has a nasty side-effect: because it removes the comment lines, the text that is fed to the YAML parser is different from the text in the original file, and so do the line numbers. So when the YAML parser reports something like: `yaml: line 57: mapping values are not allowed in this context`, actually the error could be on another line.

sjoerdsimons:
Nope that's fundamentall broken; The processing should be text file -> go templating -> yaml, doing anything else is fragile.  This seems like more of a documentation issue rather then anything else

PR #86, opened 2018-06-08
debootstrap: Allow repository verification

### LINKED ISSUE #21: actions: debootstrap: remove hardcoded '--no-check-gpg'
Add correct property for debootstrap action.

smcv:
Until now, debos has not verified apt repositories. Add a keyring option
to debootstrap, so that it can verify signed repositories against the
given keyring. If no keyring file is passed, debootstrap will attempt to
download packages securely by either using a default keyring installed
on the host system, or downloading metadata from a HTTPS mirror. A new
check-gpg option can be used to restore the previous behaviour in which
it does not check repository authenticity at all.

Heavily based on a patch from Héctor Orón Martínez (#71)
and review from Sjoerd Simons. Fixes #21.

### DIFF
--- a/actions/debootstrap_action.go
+++ b/actions/debootstrap_action.go
@@ -10,12 +10,15 @@ Yaml syntax:
    components: <list of components>
    variant: "name"
    keyring-package:
+   keyring-file:
 
 Mandatory properties:
 
 - suite -- release code name or symbolic name (e.g. "stable")
 
 Optional properties:
+
+- check-gpg -- verify GPG signatures on Release files, true by default
 
 - mirror -- URL with Debian-compatible repository
 
@@ -25,7 +28,9 @@ Optional properties:
 Example:
  components: [ main, contrib ]
 
-- keyring-package -- keyring for packages validation. Currently ignored.
+- keyring-package -- keyring for package validation.
+
+- keyring-file -- keyring file for repository validation.
 
 - merged-usr -- use merged '/usr' filesystem, true by default.
 */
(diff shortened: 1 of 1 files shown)


actions/debootstrap_action.go
```diff
@@ -79,12 +88,19 @@ func (d *DebootstrapAction) RunSecondStage(context debos.DebosContext) error {
...
-	cmdline := []string{"debootstrap", "--no-check-gpg"}
+	cmdline := []string{"debootstrap"}
 
 	if d.MergedUsr {
 		cmdline = append(cmdline, "--merged-usr")
 	}
 
+	if !d.CheckGpg {
+		cmdline = append(cmdline, fmt.Sprintf("--no-check-gpg"))
+	} else if d.KeyringFile != "" {
```
sjoerdsimons: for extra brownie points; check in the verification these optoins aren't set such that they're in conflict

actions/debootstrap_action.go
```diff
@@ -25,7 +28,9 @@ Optional properties:
 Example:
  components: [ main, contrib ]
 
-- keyring-package -- keyring for packages validation. Currently ignored.
+- keyring-package -- keyring for package validation.
+
+- keyring-file -- keyring file for repository validation.
```
sjoerdsimons: Should probably  document that this is relative to the recipe dir if a non-absolute path is given

smcv: I'll do a follow-up commit. See also #85.

actions/debootstrap_action.go
smcv: I made `check-gpg: false` take precedence over `keyring-file` on the basis that it's still true that there is a "correct" keyring file, you're just not checking against it this time. But if you think that's too much of a trap I can make it fail validation.

PR #91, opened 2018-06-14
commands, debootstrap: Don't assume the host architecture is amd64

smcv:
This complements <https://github.com/go-debos/fakemachine/pull/30>.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -8,9 +8,80 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"runtime"
 )
 
 type ChrootEnterMethod int
+
+// Return the qemu-user executable name (without a -static suffix)
+// to run binaries of the given Debian architecture, or nil if the
+// current architecture can run those binaries without using qemu.
+func GetQemuUser(debianArch string) (string, error) {
+	switch debianArch {
+	case "i386":
+		if runtime.GOARCH == "386" || runtime.GOARCH == "amd64" {
+			return "", nil
+		}
+		return "qemu-i386", nil
+	case "amd64":
+		if runtime.GOARCH == "amd64" {
+			return "", nil
+		}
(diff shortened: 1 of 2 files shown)


commands.go
```diff
@@ -8,10 +8,81 @@ import (
...
+	"runtime"
 )
 
 type ChrootEnterMethod int
 
+// Return the qemu-user executable name (without a -static suffix)
+// to run binaries of the given Debian architecture, or nil if the
+// current architecture can run those binaries without using qemu.
+func GetQemuUser(debianArch string) (string, error) {
+	switch debianArch {
```
d4s: Please rework this to use table-like data structure instead of switch-case approach.

commands.go
```diff
@@ -171,15 +242,14 @@ func newQemuHelper(c Command) qemuHelper {
...
-	case "arm64":
-		q.qemusrc = "/usr/bin/qemu-aarch64-static"
-	case "amd64", "i386":
-		/* Dummy, no qemu */
-	default:
-		log.Panicf("Don't know qemu for Architecture %s", c.Architecture)
+	qemu, err := GetQemuUser(c.Architecture)
+
+	if err != nil {
+		log.Panicf("%v", err)
```
d4s: No panic here please!

smcv: It's what the current code does in the equivalent situation. Sorry, I don't really know Go; is changing the signature of newQemuHelper so it can throw an error a valid thing to do, or would that be an ABI break?

commands.go
smcv: Sorry, I don't actually know Go. If someone who knows the language well enough to design APIs can put together a better API, that would be much appreciated.

sjoerdsimons:
superseded by newer prs

PR #94, opened 2018-06-19
Add format-image action

elboulangero:
format-image creates an image file and formats it with a filesystem. It
is meant to be followed by the 'filesystem-deploy' action.

This action is useful to prepare images that can be copied "as is" on a
block device, for example one could prepare a rootfs image and install
it with a command such as:

    dd if=rootfs.img of=/dev/sda1

Signed-off-by: Arnaud Rebillout <arnaud.rebillout@collabora.com>

### DIFF
--- a/actions/format_image_action.go
+++ b/actions/format_image_action.go
@@ -8,6 +8,7 @@ Yaml syntax:
    imagename: image_name
    imagesize: size
    fs: filesystem
+   label: label
    blocksize: 4096
 
 Mandatory properties:
@@ -17,6 +18,8 @@ Mandatory properties:
 - imagesize -- generated image size in human-readable form, examples: 100MB, 1GB, etc.
 
 - fs -- filesystem type used for formatting.
+
+- label -- volume label of the filesystem.
 
 Optional properties:
 
@@ -43,6 +46,7 @@ type FormatImageAction struct {
 	ImageName        string
 	ImageSize        string
 	FS               string
+	Label            string
 	BlockSize        int
 	size             int64
 	usingLoop        bool
@@ -54,13 +58,13 @@ func (i FormatImageAction) format(context debos.DebosContext) error {
 	cmdline := []string{}
 	switch i.FS {
 	case "vfat":
-		cmdline = append(cmdline, "mkfs.vfat")
+		cmdline = append(cmdline, "mkfs.vfat", "-n", i.Label)
 	case "btrfs":
 		// Force formatting to prevent failure in case if partition was formatted already
-		cmdline = append(cmdline, "mkfs.btrfs", "-f")
+		cmdline = append(cmdline, "mkfs.btrfs", "-L", i.Label, "-f")
 	case "none":
 	default:
-		cmdline = append(cmdline, fmt.Sprintf("mkfs.%s", i.FS))
+		cmdline = append(cmdline, fmt.Sprintf("mkfs.%s", i.FS), "-L", i.Label)
 	}
 
 	if len(cmdline) != 0 {
@@ -144,6 +148,10 @@ func (i *FormatImageAction) Verify(context *debos.DebosContext) error {
 		i.FS = "vfat"
 	case "":
 		return fmt.Errorf("Missing fs type")
+	}
+
+	if i.Label == "" {
+		return fmt.Errorf("Image without a name")
 	}
 
 	size, err := units.FromHumanSize(i.ImageSize)
--- a/actions/format_image_action.go
+++ b/actions/format_image_action.go
@@ -33,7 +33,6 @@ import (
 	"github.com/docker/go-units"
 	"github.com/go-debos/fakemachine"
 	"os"
-	"os/exec"
 	"path"
 	"strings"
 	"syscall"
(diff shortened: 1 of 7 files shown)


elboulangero:
I didn't add support for the filesystem label here, should I add it? Optional or mandatory option?

Also, for things to work we need this kind of snipper

```
  - action: format-image
  ...

  - action: filesystem-deploy
    setup-fstab: false
    setup-kernel-cmdline: false
```

We MUST set `setup-fstab` and `setup-kernel-cmdline` to false explicitly, because `format-image` doesn't prepare anything related to the fstab and kernel cmdline. And because `filesystem-deploy` has these two options set to true by default, it fails.

sjoerdsimons:
Arguably it should though; what happens if you don't set those fields? debos should really do the right thing by default.

actions/format_image_action.go
```diff
@@ -0,0 +1,157 @@
...
+type FormatImageAction struct {
+	debos.BaseAction `yaml:",inline"`
+	ImageName        string
+	ImageSize        string
+	FS               string
+	BlockSize        int
+	size             int64
+}
+
+func (i FormatImageAction) format(context debos.DebosContext) error {
```
sjoerdsimons: You'll probably want to factor this out form image-partition instead so that we don't have two copies of the code doing the same thing.

actions/format_image_action.go
```diff
@@ -0,0 +1,157 @@
...
+		return err
+	}
+
+	context.Image = image
+	*args = append(*args, "--internal-image", image)
+	return nil
+}
+
+func (i *FormatImageAction) PreNoMachine(context *debos.DebosContext) error {
+	// XXXX Couldn't test this function.
```
sjoerdsimons: Don't add todo's like this.. PreNoMachine is when you run it under sudo without fakemachine support (e.g. no /dev/kvm support(..

That said probably worth refactoring this fout of image-partition again rather then copying it here

actions/format_image_action.go
```diff
@@ -0,0 +1,157 @@
...
+		cmd := debos.Command{}
+		if err := cmd.Run(label, cmdline...); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (i FormatImageAction) PreMachine(context *debos.DebosContext, m *fakemachine.Machine,
```
sjoerdsimons: and ditto; this is copied from partition -image isn't it.

actions/format_image_action.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func (i FormatImageAction) Cleanup(context debos.DebosContext) error {
+	syscall.Unmount(context.ImageMntDir, 0)
+
+	return nil
+}
+
+func (i *FormatImageAction) Verify(context *debos.DebosContext) error {
+	// XXXX Should we really support 'fat32' as in image_partition_action.go?
```
sjoerdsimons: yeah you should otherwise it's going to be confusing

sjoerdsimons: Wonder how much of this you could share with imagepartitoin

elboulangero:
What happens is that debos tries to setup the fstab, and seeing that nothing was prepared beforehand, it fails:

```
return errors.New("Fstab not generated, missing image-partition action?")
```

Maybe we could just remove this check?

Note that for the kernel command line, there's no such check, so there's nothing to do.

actions/format_image_action.go
elboulangero: Yep there's a good deal that can be factorized, but I'm not familiar with Go so I didn't give it a try yet. I can do that and issue a second PR (updating this one will be confusing if I move too much code).

elboulangero:
Edited to remove `XXXX` comments and fix missing loop device in `PreNoMachine`.

elboulangero:
I just pushed additional commits for code factorization.

- I put common code in `image.go` but I realize now that this could also live in the existing `filesystem.go`
- I don't see a nice way of factorizing the code in `Verify`, I don't see much to factorize actually.

elboulangero:
An additional thought: in the implementation I propose here, the user has to say how big the image should be (for example `1G`).

The pros of that is that it keeps the implementation simple (in-line with what's already done in `image-partition`).

The cons is that the result is not so optimal: if you end up filling this image with only 500M of data, then why would you want your image to be 1G? Wouldn't you want to have it exactly the right size? Probably, but not sure.

For the user it's not a big problem though, for different reasons:
- you can shrink the image file, it's not difficult. For an ext4 filesystem it just takes two commands.
- you can compress the image file, which works well. I tried on a 1G and 2G images (containing the same amount of data), and the compressed files end up with roughly the same size.

So for now, I'd be in favor of keeping things simple, ie. the way I've implemented it in the PR, and let the user shrink the image by himself in a postprocess step if he wants to.

elboulangero:
Right now I'm actually not sure I need this action.

Another way to get a `block device image` is to extract it out of the disk image after it's created. I realize now that it can work even better than having a `format-image` action.

So let's put that on hold for now. Thanks.

sjoerdsimons:
@elboulangero As we're now a month or so later lets close this one for now; If it turns out you have a good use-case for this please reopen it ;)

PR #95, opened 2018-06-22
ostree-commit: Disable fsync to reduce I/O load

em-:
Debos is likely to be used on ephemeral repositories, so using `fsync`
slows down things without much benefit.


sjoerdsimons:
Seems this one has gone stale; closing (please reopen to restart the discussion)

Issue #61, opened 2018-01-18
how to INSTALL


stappersg:
Hi,

How to install  debos?

Yes, I'm new to the go-lang world.

In the GCC world there is the good tradition of providing an INSTALL file.

Please provide an eqivalent of the

```
./configure
make
sudo make install
```

dance, preferable in INSTALL   :-)

See also https://en.wikipedia.org/wiki/README#As_a_generic_term

stappersg:
Things like `go build`, `go get` or `go install` don't work ...

zumbi:
Hello,
  You can clone the repository and build it. FYI, you might be able to install using Debian packages now.
Regards

stappersg:
On Tue, Mar 06, 2018 at 05:26:17PM +0000, Hector Oron wrote:
> Hello,
>   You can clone the repository and build it.

Yes, my question is how to build debos.

I do know how to clone a git repository,
but don't known how to build an executable from go-lang source.


A websearch on `go build` did bring up https://golang.org/pkg/go/build/
That page lacks an equivalent of "make".



> FYI, you might be able to install using Debian packages now.

The INSTALL in the subject is a reference to the file INSTALL
that is usually ship with C source  tar ball.


Cheers
Geert Stappers

P.S.
Thank you for packaging debos for debian

stappersg:
First there where errors like
```
No package 'ostree-1' found
Package ostree-1 was not found in the pkg-config search path.
Perhaps you should add the directory containing `ostree-1.pc'
to the PKG_CONFIG_PATH environment variable
No package 'glib-2.0' found
Package gobject-2.0 was not found in the pkg-config search path.
Perhaps you should add the directory containing `gobject-2.0.pc'
to the PKG_CONFIG_PATH environment variable
No package 'gobject-2.0' found
Package glib-2.0 was not found in the pkg-config search path.
Perhaps you should add the directory containing `glib-2.0.pc'
to the PKG_CONFIG_PATH environment variable
```

Got those errors fixed by
```
sudo apt install libglib2.0-dev libostree-dev
```
Then:
```
stappers@bob:~/debosbuild$ go get -u github.com/go-debos/debos/cmd/debos
stappers@bob:~/debosbuild$ 
stappers@bob:~/debosbuild$ ls
stappers@bob:~/debosbuild$ time go get -u github.com/go-debos/debos/cmd/debos

real	0m6.180s
user	0m1.294s
sys	0m0.388s
stappers@bob:~/debosbuild$ ls
[...]

PR #97, opened 2018-06-28
Document installation under Debian

ngiger:
Found the needed help via https://github.com/go-debos/debos/issues/61.

Tested this on Debian Testing and Stable  (stretch)


ana:
Hi! Thanks for your update. For the record, the package debos is also available in both testing and sid and the same package should work in stretch.

PR #99, opened 2018-07-03
Cleanup

d4s:
Reworked cleanup logic allow to remove artifacts created by Per(No)Machine() methods.
Cleanup is now done in reverse order instead of listed in recipe file.

### DIFF
--- a/action.go
+++ b/action.go
@@ -6,6 +6,14 @@ import (
 	"log"
 )
 
+type DebosState int
+
+// Represent the current state of Debos
+const (
+	Success DebosState = iota
+	Failed
+)
+
 // Mapping from partition name as configured in the image-partition action to
 // device path for usage by other actions
 type Partition struct {
@@ -27,6 +35,7 @@ type DebosContext struct {
 	Architecture    string
 	DebugShell      string
 	Origins         map[string]string
+	State           DebosState
 }
 
 type Action interface {
@@ -35,8 +44,13 @@ type Action interface {
 	PreMachine(context *DebosContext, m *fakemachine.Machine, args *[]string) error
 	PreNoMachine(context *DebosContext) error
 	Run(context *DebosContext) error
-	Cleanup(context DebosContext) error
-	PostMachine(context DebosContext) error
+	// Cleanup() method gets called only if the Run for an action
+	// was started and in the same machine (host or fake) as Run has run
+	Cleanup(context *DebosContext) error
+	PostMachine(context *DebosContext) error
+	// PostMachineCleanup() gets called for all actions if Pre*Machine() method
+	// has run for Action. This method is always executed on the host with user's permissions.
+	PostMachineCleanup(context *DebosContext) error
 	String() string
 }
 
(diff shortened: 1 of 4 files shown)


cmd/debos/debos.go
```diff
@@ -200,8 +207,11 @@ func main() {
 
 	if !fakemachine.InMachine() {
 		for _, a := range r.Actions {
+			// Stack PostCleanup methods
+			defer a.PostCleanup(&context)
```
sjoerdsimons: I find that using defer here makes it very hard to reason about when the actions are ran. I'd much rather see this be done explicitly.

cmd/debos/debos.go
```diff
@@ -218,22 +228,21 @@ func main() {
...
-			return
-		}
-	}
 
-	for _, a := range r.Actions {
-		err = a.Cleanup(context)
-		if exitcode = checkError(context, err, a, "Cleanup"); exitcode != 0 {
+		// This does not stop the call of stacked Cleanup methods for other Actions
+		// Stack Cleanup methods
+		defer a.Cleanup(&context)
```
sjoerdsimons: Same; i've got a much harder time reasoning about this.

action.go
```diff
@@ -35,8 +44,9 @@ type Action interface {
 	PreMachine(context *DebosContext, m *fakemachine.Machine, args *[]string) error
 	PreNoMachine(context *DebosContext) error
 	Run(context *DebosContext) error
-	Cleanup(context DebosContext) error
-	PostMachine(context DebosContext) error
+	Cleanup(context *DebosContext) error
+	PostCleanup(context *DebosContext) error
```
sjoerdsimons: If you were to split out the passing by reference of all the things in a sepeate commit this would have been so so much easier to review

action.go
```diff
@@ -57,8 +67,9 @@ func (b *BaseAction) PreMachine(context *DebosContext,
 }
 func (b *BaseAction) PreNoMachine(context *DebosContext) error { return nil }
 func (b *BaseAction) Run(context *DebosContext) error          { return nil }
-func (b *BaseAction) Cleanup(context DebosContext) error       { return nil }
-func (b *BaseAction) PostMachine(context DebosContext) error   { return nil }
+func (b *BaseAction) Cleanup(context *DebosContext) error      { return nil }
```
sjoerdsimons: Ok lets see if i got this right; As i can't really work it out entirely from your commit message

* Cleanup() gets called only if the Run  for an action was started and in the same machine (host or fake) as Run has run
* PostCleanup() gets call for all actions regardless and always on the host
* Due to all the deferring both *Cleanup methods get called in reverse order; Cleanup always happens first (either on host or fake).

You can't use PostMachine as that doesn't get called on failure.

For some reason this doesn't sit right with me (mainly because the order is getting a bit complex ;)).. I think it would help if we name things so it's a bit more clear when it gets called? I found it quite hard to work out what hte differences between the two cleanup methods are.

Maybe make it `CleanupRun` and `PostMachineCleanup` ?

cmd/debos/debos.go
sjoerdsimons: Looking at it some more; I do find it awkward but not using defers would mean more refactoring then is worth atm, so ignore those commetns for now

PR #106, opened 2018-08-16
add --http-proxy

petersenna:


petersenna:
On my setup the proxy is on the host, and from the host pointing to localhost will work fine. For making it work inside debos, I need to specify the host ip instead of localhost. For my setup it is more convenient to use --http-proxy than configuring an environment variable specific to debos.

sjoerdsimons:
So you're using it to overwrite; which is fine. But i don't see a reason to force people to do that (arguable if we're nice debos should warn folks about using localhost as the proxy)...  Also why do you have a proxy in this century :p

petersenna:
I don't understand what you mean by overwrite. debos is a special case in which using localhost does not work, which is particular to debos. I live in a small town, loads of cows around here. :-)

PR #109, opened 2018-08-20
Add support to environment variables

petersenna:
This change makes debos read a predefined set of environment variables
from the host and propagates it to fakemachine. The set of environment
variables is defined by environ_vars on cmd/debos/debos.go. Currently
the list of environment variables is configured to the proxy environment
variables as documented here[1]:

  http_proxy, https_proxy, ftp_proxy, rsync_proxy, all_proxy, no_proxy

While the elements of environ_vars are in lower case, for each element
both lower and upper case variants are probed on the host, and if found
propagated to fakemachine. So if the host has the environment variables
HTTP_PROXY and no_proxy defined, both will be propagated to fakemachine
respecting the case.

This patch also adds --environ-var and -e command line options that can
be used to specify, overwrite, and unset environment variables for
fakemachine with the syntax:

  $ debos -e ENVIRONVAR:VALUE ...

To unset an enviroment variable, or in other words, to prevent an
environment variable to be propagated to fakemachine, use the same
syntax without a value. debos accept multiple -e simultaneously.

[1] - https://wiki.archlinux.org/index.php/proxy_settings

This change depends on:

  8b7eea7096cd015771cf9862647b7cd04bb41904 - machine.go: Add support to
  environment variables

on fakemachine repository.

Signed-off-by: Peter Senna Tschudin <peter.senna@collabora.com>


petersenna:
Go also uses environment variables for proxy which made the download action work without any further modification. The only place where the variables were not configured was inside the chroot and the latest patch made the missing changes.

petersenna:
Yes, doing so will affect all downloads, as even go itself uses the
environment variables for proxy. I added documentation describing the known
issues and a few more details.

smcv:
A version rebased on current master: https://github.com/smcv/debos/tree/http_proxy

Issue #9, opened 2017-09-15
Create a standard docker image with debos to push to docker hub


d4s:

macpijan:
Hello. I was struggling to run the debos on Ubuntu and I ended up trying to develop a Debian docker container for it. You can find my work here: https://github.com/3mdeb/debos/blob/add-dockerfile/docker/Dockerfile

Unfortunately, build for arch other than the native (`amd64`) fails at:
`Debootstrap (stage 2) | chroot: failed to run command '/debootstrap/debootstrap': Exec format error`

I also wrote a short post to document my triage: https://3mdeb.com/os-dev/our-first-look-at-debos-new-debian-images-generator/

If you have a clue why does it fail, I will be glad to fix and contribute the Dockerfile. `debos` seems like a nice project, but limiting the usage to Debian-only host seems a little bit limiting. Having a portable container could allow me (and possibly others) to take advantage of this tool more. Thanks.

PR #110, opened 2018-08-21
docker: add basic docker support

macpijan:
Signed-off-by: Maciej Pijanowski <maciej.pijanowski@3mdeb.com>

The problem with building for non-host arch as described in the:
https://github.com/go-debos/debos/issues/9 was resolved. Basic
arm64 example image building was tested on the Ubuntu 18.04.
This may enable more users to take advantage of the debos, as
using it on non-Debian distros seems either impossible or not
trivial.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -29,7 +29,10 @@ RUN apt-get update && \
     ca-certificates \
     systemd-container \
     binfmt-support \
-    patch \
+    parted \
+    dosfstools \
+    e2fsprogs \
+    bmap-tools \
     # fakemachine runtime dependencies
     qemu-system-x86 \
     qemu-user-static \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -14,7 +14,6 @@ RUN apt-get update && \
 
 RUN go get -d github.com/go-debos/debos/cmd/debos
 WORKDIR /go/src/github.com/go-debos/debos/
-# RUN dep ensure
 RUN GOOS=linux go build -a cmd/debos/debos.go
 
 ### second stage - runner ###
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -47,7 +47,8 @@ RUN apt-get update && \
 # provide 1.0.100.
 RUN printf "deb http://httpredir.debian.org/debian stretch-backports main \ndeb-src http://httpredir.debian.org/debian stretch-backports main" > /etc/apt/sources.list.d/backports.list && \
     apt-get update && \
-    apt-get -t stretch-backports install -y debootstrap && \
+    apt-get -t stretch-backports install -y --no-install-recommends \
+    debootstrap && \
     rm -rf /var/lib/apt/lists/*
 
 COPY --from=builder /go/src/github.com/go-debos/debos/debos /usr/bin/debos
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,53 @@
+### first stage - builder ###
+FROM golang:1.10 as builder
+
+MAINTAINER Maciej Pijanowski <maciej.pijanowski@3mdeb.com>
+
+ENV HOME=/scratch
+
+# install debos build dependencies
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    libglib2.0-dev \
+    libostree-dev \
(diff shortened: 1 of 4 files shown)


petersenna:
little detail: no need to enable non-free for installing debootstrap from backports.

sjoerdsimons:
Fwiw https://gitlab.collabora.com/docker/debos is the basic docker file we tend to use; typical our recipes only have one component for debootstrap so that isn't relaly a problem in practise.

For your run script; debos really should not be run priviledged, running unpriviledged is part of the point. 

But yes debos having an official docker image to use is a good feature to have

macpijan:
I did not know such Dockerfile exists - it could be helpful.
You are totally right about the `privileged` mode. So far I have not managed to run it unprivileged due to the:

```
2018/08/21 19:17:06 Debootstrap (stage 2) | W: Failure trying to run:  mount -t proc proc /proc
2018/08/21 19:17:06 Debootstrap (stage 2) | W: See //debootstrap/debootstrap.log for details
2018/08/21 19:17:06 debootstrap.log | gpgv: Signature made Tue Aug 21 14:20:20 2018 UTC
2018/08/21 19:17:06 debootstrap.log | gpgv:                using RSA key A1BD8E9D78F7FE5C3E65D8AF8B48AD6246925553
2018/08/21 19:17:06 debootstrap.log | gpgv: Good signature from "Debian Archive Automatic Signing Key (7.0/wheezy) <ftpmaster@debian.org>"
2018/08/21 19:17:06 debootstrap.log | gpgv: Signature made Tue Aug 21 14:20:20 2018 UTC
2018/08/21 19:17:06 debootstrap.log | gpgv:                using RSA key 126C0D24BD8A2942CC7DF8AC7638D0442B90D010
2018/08/21 19:17:06 debootstrap.log | gpgv: Good signature from "Debian Archive Automatic Signing Key (8/jessie) <ftpmaster@debian.org>"
2018/08/21 19:17:06 debootstrap.log | gpgv: Signature made Tue Aug 21 14:20:20 2018 UTC
2018/08/21 19:17:06 debootstrap.log | gpgv:                using RSA key 16E90B3FDF65EDE3AA7F323C04EE7237B7D453EC
2018/08/21 19:17:06 debootstrap.log | gpgv: Good signature from "Debian Archive Automatic Signing Key (9/stretch) <ftpmaster@debian.org>"
2018/08/21 19:17:06 debootstrap.log | mount: /proc: permission denied.
2018/08/21 19:17:06 Action `debootstrap` failed at stage Run, error: exit status 1
```

I have not investigated it any further yet and posted the version which actually let me to successfully build the image.

sjoerdsimons:
That'll be because you don't seem to be exposing /dev/kvm in your docker container; see e.g. https://gitlab.apertis.org/infrastructure/apertis-image-recipes/blob/master/README for an exmaple docker line on how to run debos. 

If there isn't kvm it'll try to run in a different mode which doesn't work so well in docker (there should be a note at the start of your output about that).. Tbh i'm surprised you could build images this way as i wouldn't expect loop devices to be exposed in your docker to do an image build? Or did you just bulid a rootfs tarball?

macpijan:
Yes, sorry. My first example recipe was: https://github.com/go-debos/debos/blob/master/doc/examples/example.yaml which in fact produces just a tarball, not a full image. 
So the image phase would need the loop devices to be exposed. Also, some binaries like `parted` seem to be missing in my Dockerfile:
`Action image-partition failed at stage Run, error: exec: "parted": executable file not found in $PATH`
Thanks for the link. I will try to take advantage of the `kvm` device and allow for full image building.

macpijan:
Some results:
* After adding the `parted` and filesystem-related packages I was able to build the full disk image (https://github.com/go-debos/debos-recipes/blob/master/debian/arm64/image-rpi3/debimage-rpi3.yaml) with my container
* I dropped the privileged mode and exposed the `kvm` device according to your suggestion - I am still able to build the rpi3 image with the current container and `run` command.

macpijan:
I am not sure what would be the plan for having this debos docker image.
For now, I am successfully using this container: https://github.com/3mdeb/debos-docker for a few debos use-cases I have.

sjoerdsimons:
@macpijan Sorry for the tardy reply; @fdanis-oss has been picking up your nice work for #150 so closing this one now; Thanks for your work!

PR #119, opened 2018-10-05
Pass through http_proxy, etc. from caller's environment

smcv:
This allows images to be built using a caching proxy like apt-cacher-ng, without its address having to be in the inputs or ending up hard-coded into the output.

As with all my PRs, this is probably not very idiomatic; I don't actually know Go.


smcv:
#109 is probably a better change than this, although it needs a corresponding change in fakemachine.

PR #122, opened 2018-10-15
Wip/sjoerd/sync no fakemachine mode

sjoerdsimons:
When building without fakemachine we've seen some issues where not all data was correctly flushed (or synced) into the image file before the post-processing actions ran. This caused issues with e.g. invalid checksums as well as bootloaders not being correctly on the image. The issues should be resolved with the various fixes in this PR

### DIFF
--- a/cmd/debos/debos.go
+++ b/cmd/debos/debos.go
@@ -25,6 +25,23 @@ func checkError(context *debos.DebosContext, err error, a debos.Action, stage st
 	return 1
 }
 
+func do_run(r recipe.Recipe, context *debos.DebosContext) int {
+	for _, a := range r.Actions {
+		err := a.Run(context)
+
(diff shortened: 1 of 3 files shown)


cmd/debos/debos.go
```diff
@@ -226,17 +243,9 @@ func main() {
...
-
-		// This does not stop the call of stacked Cleanup methods for other Actions
-		// Stack Cleanup methods
-		defer a.Cleanup(&context)
-
-		// Check the state of Run method
-		if exitcode = checkError(&context, err, a, "Run"); exitcode != 0 {
-			return
-		}
+	exitcode = do_run(r, &context)
```
d4s: Just a nitpick: may I ask you to compact this a bit? Something like.
if exitcode = do_run(r, &context); exitcode !=0 {
  return 0;
}

Issue #129, opened 2018-11-18
`pack` action packs /dev/ tree


cceckman:
I'm trying to do an incremental build using `debos`; however, I'm getting some un-unarchiveable archives from `pack`. A small example:

```shell
∵ go get -u -v github.com/go-debos/debos/cmd/debos 2>/dev/null
∴ 0 cceckman:master…Downloads
∵ cat example.yaml
{{- $image := or .image "debian.tgz" -}}

architecture: arm64

actions:
  - action: debootstrap
    suite: "stretch"
    components:
      - main
    mirror: https://deb.debian.org/debian
    variant: minbase

  - action: pack
    file: {{ $image }}
    compression: gz
∴ 0 cceckman:master…Downloads
∵ sudo $(which debos) -t image:img.tar.gz example.yaml >/dev/null 2>&1
∴ 0 cceckman:master…Downloads
∵ tar -xvf img.tar.gz >/dev/null
tar: ./dev/ptmx: Cannot mknod: Operation not permitted
tar: ./dev/tty: Cannot mknod: Operation not permitted
tar: ./dev/null: Cannot mknod: Operation not permitted
tar: ./dev/random: Cannot mknod: Operation not permitted
tar: ./dev/zero: Cannot mknod: Operation not permitted
tar: ./dev/full: Cannot mknod: Operation not permitted
tar: ./dev/urandom: Cannot mknod: Operation not permitted
tar: Exiting with failure status due to previous errors
∴ 2 cceckman:master…Downloads
```

Note that `tar` exits with status `2`; this prevents (a later invocation of) `debos` from unpacking.

It looks like the archive produced by the `pack` action is including parts of the `dev` tree. This seems not-really-right; I get `/dev` being present (as a mountpoint target), but not populated with files.

Are there some directories that should be excluded from the [pack action](https://github.com/go-debos/debos/blob/f8d7e4804f19e963682aa93d23737735ec7d7103/actions/pack_action.go#L33)? Or (very likely) are there more subtleties to the `/dev` system that I'm missing? :-)

d4s:
According to your log -- `debos` works as expected.
Your issue is related to `tar` usage -- your host OS didn't allow non-root user to create device nodes (please use `sudo` for that).
It have no effect to later `debos` usage since it works in `qemu` virtual machine and have all needed "privileges".

[...]

PR #130, opened 2018-11-18
Exclude /dev's contents from `pack` command.

cceckman:
Changes the `tar` invocation used by `pack` to exclude the contents of
the `./dev` subtree. In the fakemachine / chroot, these are devices
(`urandom` etc) that aren't actually packable- or rather, aren't
unpackable.

Fixes go-debos/debos#129.


d4s:
Please check my [comment](https://github.com/go-debos/debos/issues/129#issuecomment-450203753)

PR #132, opened 2018-11-20
apt: allow unathenticated packages installation

zumbi:
When using unsigned repositories apt action is unable to install
packages, this option allows a setting to install from unauthenticated
repositories.

Signed-off-by: Héctor Orón Martínez <hector.oron@collabora.com>


zumbi:
Any update on this one?

sjoerdsimons:
I'm not too big a fan of this; but we have something similar to it for debootstrap so i guess it's ok

Issue #46, opened 2017-11-15
actions: image-partition: Add build-time mount points


d4s:
Add mountpoint which are mounted only in build time, e.g. do not need to add them into fstab and need to remove mount points on rootfs after umounting.

Thread974:
Second that!

PR #135, opened 2018-12-27
image_partition_action: add 'buildtime' option for mountpoint

d4s:
If option is set to true -- mount point is added only during the image
creation without additional entry in `/etc/fstab' on target.
Mount point directory is wiped out from the image.
Defaults to false

Fixes: #46


actions/image_partition_action.go
```diff
@@ -81,6 +82,10 @@ should be mounted.
 Optional properties:
 
 - options -- list of options to be added to appropriate entry in fstab file.
+- buildtime -- if set to true then mount point is added only during the image
+creation without additional entry in `/etc/fstab' on target. Mount point directory
+is wiped out from the image.
+Defaults to false.
```
sjoerdsimons: The removal seems a bit unexpected. May make sense to document this a bit better and provide some hint on how to do it (e.g. use `/media/temporarybadger` not `/media` or `/mnt` directly)

I've also pondered of whether instead of the user having to specifiy a mountpoint in these cases that should be left unset and debos will create one; but that then needs to communicate to user scripts (or overlays) the location and that seems more haphazard..

So I *think* this way is probably the best trade-off; but a bit more clear documentation will be good.

actions/image_partition_action.go
```diff
@@ -81,6 +82,11 @@ should be mounted.
 Optional properties:
 
 - options -- list of options to be added to appropriate entry in fstab file.
+- buildtime -- if set to true then mount point is added only for the image
+creation without additional entry in `/etc/fstab' on target.
+Mount point directory will be removed from the image, so need to define `mountpoint` path
```
sjoerdsimons:
```suggestion
The mountpoints directory will be removed from the image, so it is recommended to define a `mountpoint` path
```

actions/image_partition_action.go
```diff
@@ -81,6 +82,11 @@ should be mounted.
 Optional properties:
 
 - options -- list of options to be added to appropriate entry in fstab file.
+- buildtime -- if set to true then mount point is added only for the image
```
sjoerdsimons:
```suggestion
- buildtime -- if set to true then the mountpoint only used during the debos run.
```

actions/image_partition_action.go
```diff
@@ -81,6 +82,11 @@ should be mounted.
...
+creation without additional entry in `/etc/fstab' on target.
```
sjoerdsimons:
```suggestion
No entry in `/etc/fstab' will be created.
```

PR #142, opened 2019-02-07
cmd: Add options to print recipe used for build

fdanis-oss:
Add 'verbose' option to log the actions used for the build


d4s:
Please try reflect to list all fields and their values instead of listing all fields manually.

sjoerdsimons:
Also this is really a developer feature from my pov. May make sense to start adding `-v`'s for more verbose logging for this type of stuff.

fdanis-oss:
All fields listing done using reflection. The new verbose option is used to log all fields.

PR #143, opened 2019-02-10
ostree-commit: add support of metadata and collections

d4s:
Allow to add metadata key-value pairs into ostree commit.

Parameter `collection-id` allows bindings to collection refs (collection id + branch). This add metadata information into commit: `ostree.collection-binding` and `ostree.ref-binding` (binding only to created branch).

Requires libostree minimal version 2018.6.
Depends on changes proposed in https://github.com/sjoerdsimons/ostree-go/pull/1

### DIFF
--- a/actions/ostree_commit_action.go
+++ b/actions/ostree_commit_action.go
@@ -8,6 +8,9 @@ Yaml syntax:
    repository: repository name
    branch: branch name
    subject: commit message
+   metadata:
+     key: value
+     vendor.key: somevalue
 
 Mandatory properties:
 
@@ -22,10 +25,13 @@ type (https://ostree.readthedocs.io/en/latest/manual/repo/#repository-types-and-
 Optional properties:
 
 - subject -- one line message with commit description.
+
+- metadata -- key-value pairs of meta information to be added into commit.
 */
 package actions
 
 import (
+	"fmt"
 	"log"
 	"os"
 	"path"
@@ -40,6 +46,7 @@ type OstreeCommitAction struct {
 	Branch           string
 	Subject          string
 	Command          string
+	Metadata         map[string]string
 }
 
 func emptyDir(dir string) {
@@ -54,7 +61,7 @@ func emptyDir(dir string) {
 	for _, f := range files {
 		err := os.RemoveAll(path.Join(dir, f))
 		if err != nil {
-	                log.Fatalf("Failed to remove file: %v", err)
+			log.Fatalf("Failed to remove file: %v", err)
 		}
 	}
 }
@@ -77,6 +84,11 @@ func (ot *OstreeCommitAction) Run(context *debos.DebosContext) error {
 
 	opts := otbuiltin.NewCommitOptions()
 	opts.Subject = ot.Subject
+	for k, v := range ot.Metadata {
+		str := fmt.Sprintf("%s=%s", k, v)
+		opts.AddMetadataString = append(opts.AddMetadataString, str)
+	}
+
 	ret, err := repo.Commit(context.Rootdir, ot.Branch, opts)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


actions/ostree_commit_action.go
```diff
@@ -8,6 +8,10 @@ Yaml syntax:
    repository: repository name
    branch: branch name
    subject: commit message
+   collection-id: org.apertis.example
+   metadata:
+     - key = value
```
sjoerdsimons: Why isn't this just a map in yaml?

actions/ostree_commit_action.go
```diff
@@ -75,8 +117,22 @@ func (ot *OstreeCommitAction) Run(context *debos.DebosContext) error {
...
 	opts := otbuiltin.NewCommitOptions()
 	opts.Subject = ot.Subject
+	for k, v := range ot.meta {
+		str := fmt.Sprintf("%s=%s", k, v)
+		opts.AddMetadataString = append(opts.AddMetadataString, str)
+	}
+
+	if ot.CollectionID != "" {
+		opts.CollectionID = ot.CollectionID
+		opts.RefBindings = append(opts.RefBindings, ot.Branch)
```
sjoerdsimons: refbindings should be configurable as extra brances to add on top of the current one. e.g. for being able to switch between a development and production branch.

Also is it correct to do reference binding *only* when there is a collection id and if there is one then *enforce* it ? that seems like encoding a lot of policy.

d4s: From my findings "ref-binding" should be inserted in case if "collection-id" added.

Add following logic:
If "ref-bindings:" is absent but "collection-id:" is set, then implicitly add only current branch.  
if "ref-bindings:" parameter is set explicitly for "ostree-commit" action, then set only listed branches from that parameter despite of "collection-id".

actions/ostree_commit_action.go
```diff
@@ -48,13 +66,15 @@ func emptyDir(dir string) {
...
 	if err != nil {
+		//TODO: remove
 		log.Fatal(err)
 	}
 
 	for _, f := range files {
 		err := os.RemoveAll(path.Join(dir, f))
 		if err != nil {
-	                log.Fatalf("Failed to remove file: %v", err)
+			//TODO: remove
```
sjoerdsimons: THis should really be a seperate commit

actions/ostree_commit_action.go
```diff
@@ -22,10 +29,18 @@ type (https://ostree.readthedocs.io/en/latest/manual/repo/#repository-types-and-
 Optional properties:
 
 - subject -- one line message with commit description.
+
+- collection-id -- Collection ID ref binding (require libostree 2018.6).
```
sjoerdsimons: *requires

actions/ostree_commit_action.go
```diff
@@ -22,10 +29,18 @@ type (https://ostree.readthedocs.io/en/latest/manual/repo/#repository-types-and-
...
+
+- ref-binding -- enforce that the commit was retrieved from one of the branch names in this array.
+  If 'collection-id' is set and 'ref-binding' is empty, the branch name will be used implicitly.
```
sjoerdsimons: unclear what `the branch name` refers to; will default to the branch name.

Though makes me wonder if ref-binding is set should the branch name be added by default if not already in the list? Are there cases where we do not want to add it?

d4s: tbh I prefer to have a possibility to override implicit things.

actions/ostree_commit_action.go
```diff
@@ -54,7 +72,7 @@ func emptyDir(dir string) {
 	for _, f := range files {
 		err := os.RemoveAll(path.Join(dir, f))
 		if err != nil {
-	                log.Fatalf("Failed to remove file: %v", err)
+			log.Fatalf("Failed to remove file: %v", err)
```
sjoerdsimons: sprurious change?

PR #144, opened 2019-02-11
cmd: Add options to print final recipe used for build

fdanis-oss:


recipe/recipe.go
```diff
@@ -159,6 +159,10 @@ func (r *Recipe) Parse(file string, templateVars ...map[string]string) error {
 		return err
 	}
 
+	if (dump) {
+		fmt.Println(data)
```
sjoerdsimons: Nitpick; would be nice to print a header when doing this.

fdanis-oss: Done

PR #146, opened 2019-02-19
Add new Recipe action

fdanis-oss:


cmd/debos/debos.go
```diff
@@ -204,16 +204,6 @@ func main() {
 		return
 	}
 
-	if options.Verbose {
-		log.Println("Internal actions dump:")
-		DumpActions(reflect.ValueOf(r).Interface(), 0)
```
sjoerdsimons: I wonder if it's not the verify of the recipe action that should dump the actions after verify seems late.

actions/recipe_action.go
```diff
@@ -116,7 +119,7 @@ func (recipe *RecipeAction) Run(context *debos.DebosContext) error {
 	recipe.LogStart()
 
 	for _, a := range recipe.Actions.Actions {
-		if err := a.Run(context); err != nil {
+		if err := a.Run(&recipe.context); err != nil {
```
sjoerdsimons: This is too little; another Run action can change e.g. the rootdir which I don't think you'd catch this way

actions/recipe_action.go
```diff
@@ -38,42 +38,45 @@ type RecipeAction struct {
 	Inherits         string
 	Variables        map[string]string
 	Actions          Recipe
-	templateVars     map[string]string
+	context          debos.DebosContext
```
sjoerdsimons: I don't think just using another DebosContext makes sense as values in the context can change during Runs.

cmd/debos/debos.go
```diff
@@ -117,10 +117,10 @@ func DumpActionFields(iface interface{}, depth int) {
 			case reflect.Slice:
 				s := reflect.ValueOf(f.Interface())
 				if s.Len() > 0 && s.Index(0).Kind() == reflect.Struct {
-					log.Printf("    %s:\n", entries.Type().Field(i).Name)
+					log.Printf("%s  %s:\n", tab, entries.Type().Field(i).Name)
```
sjoerdsimons: Your commit doesn't explain what this does; Should go into a seperaet MR as well maybe?

actions/recipe_action.go
```diff
@@ -6,7 +6,6 @@ Include a recipe.
 Yaml syntax:
  - action: recipe
    recipe: path to recipe
-   inherit: bool
```
sjoerdsimons: Squash this into your first commit please

actions/recipe_action.go
```diff
@@ -60,15 +61,19 @@ func (recipe *RecipeAction) Verify(context *debos.DebosContext) error {
 
 	// Add Variables to template vars
 	for k, v := range recipe.Variables {
+		if k == "architecture" {
```
sjoerdsimons: This is weird. Also the commit message doesn't explain much of what this is intending to achieve.

actions/recipe_test.go
```diff
@@ -1,7 +1,7 @@
 package actions_test
 
 import (
-	"github.com/go-debos/debos/recipe"
+	"github.com/go-debos/debos/actions"
```
sjoerdsimons: This should be combined in the patch that moves things around so we don't have random commits that fail tests

actions/recipe_action.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/go-debos/debos"
+	"github.com/go-debos/fakemachine"
+)
+
+type RecipeAction struct {
+	debos.BaseAction `yaml:",inline"`
+	Recipe           string
+	Inherits         string
+	Variables        map[string]string
+	Actions          Recipe
```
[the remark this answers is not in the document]
fdanis-oss: ok

actions/recipe.go
```diff
@@ -57,17 +57,19 @@ Supported actions
...
 */
-package recipe
+package actions
 
 import (
 	"bytes"
 	"fmt"
 	"github.com/go-debos/debos"
-	"github.com/go-debos/debos/actions"
+//	"github.com/go-debos/debos/actions"
```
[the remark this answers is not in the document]
fdanis-oss: Removed

actions/recipe_action.go
fdanis-oss: Squashed

actions/recipe_test.go
fdanis-oss: Squashed into commit moving recipe/recipe.go to actions/recipe.go

actions/recipe_action.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	return nil
+}
+
+func (recipe *RecipeAction) Run(context *debos.DebosContext) error {
+	recipe.LogStart()
+
+	for _, a := range recipe.Actions.Actions {
+		if err := a.Run(context); err != nil {
```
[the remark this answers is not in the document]
sjoerdsimons: As discussed for now just re-using the main context is fine; This just means there is a limitation that the including recipes need to be in the same directory, which is fine for a first merge and we can ttweak it later.

cmd/debos/debos.go
sjoerdsimons: any thoughts on this?

fdanis-oss: I think dump of the actions should be done after all recipes have been loaded, so after Verify has been called for all recipes.
This dumps the final list of all actions to do.
If this is done in Verify, it can be intermixed with the recipe print.

sjoerdsimons: On the other hand doing it this way  means the dump might not be printend at all of the Verify of an action fails, which is rather annoying as being able to use this to help debug is rather useful... I dont mind so much if the output would be the following:

<raw recipe of the root recipe>
<dump of the fileds of the root recip structs>
<raw recipe for first reecipe action>
<dump of fields first recipe action>
etc

the main key would be ensure the output is clear for which recipe the dumps are

PR #147, opened 2019-02-19
image-partition: add `fsck` flag

d4s:
Enable filesystems check by default by setting `fs_passno` in fstab.
Allow to disable filesystem check on partition by setting flag `fsck`
to `false`.

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -42,6 +42,7 @@ Yaml syntax for partitions:
 	   start: offset
 	   end: offset
 	   flags: list of flags
+	   fsck: bool
 
 Mandatory properties:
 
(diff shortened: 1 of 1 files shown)


actions/image_partition_action.go
```diff
@@ -139,6 +144,7 @@ type Partition struct {
 	End    string
 	FS     string
 	Flags  []string
+	NoFsck bool "no_fsck"
```
sjoerdsimons: All other mutliple word option have a `-` iirc

actions/image_partition_action.go
```diff
@@ -42,6 +42,7 @@ Yaml syntax for partitions:
 	   start: offset
 	   end: offset
 	   flags: list of flags
+	   no_fsck: bool
```
sjoerdsimons: Why is this a negation? seems better to just call it fsck or enable-fsck or somesuch and default to true. negative options are confusing :)

d4s: fsck is enabled by default, so need to switch it off explicitly.
will add `fsck: bool` then

actions/image_partition_action.go
```diff
@@ -82,6 +86,7 @@ should be mounted.
 Optional properties:
 
 - options -- list of options to be added to appropriate entry in fstab file.
+
```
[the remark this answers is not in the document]
d4s: yes. moved to separate commit.

PR #150, opened 2019-03-12
Debos docker

fdanis-oss:
Add Dockerfile for debos

Debos container build is split in 2 phases:
- 1st stage builds the debos executable
- 2nd stage provides debos and its dependencies

This 2 stages build allows to avoid to embed the go toolchain in the final image.


docker/Dockerfile
```diff
@@ -0,0 +1,66 @@
+# Global ARGs shared by all stages
+ARG DEBIAN_FRONTEND=noninteractive
+ARG GOPATH=/usr/local/go
+
+### first stage - builder ###
+FROM debian:stretch-slim as builder
```
sjoerdsimons: I'd move to buster at this stage :)

docker/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+
+ARG DEBIAN_FRONTEND
+ARG GOPATH
+
+# Dependencies to build debos
+# - libostree: Latest debos needs version from stretch-backports at least
+RUN printf "deb http://httpredir.debian.org/debian stretch-backports main \ndeb-src http://httpredir.debian.org/debian stretch-backports main" > /etc/apt/sources.list.d/backports.list && \
```
sjoerdsimons: Moving to buster would avoid this hoop

docker/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+    rm -rf /var/lib/apt/lists/*
+
+# Build debos
+COPY . $GOPATH/src/github.com/go-debos/debos
+WORKDIR $GOPATH/src/github.com/go-debos/debos/cmd/debos
+RUN go get -d ./... && \
+    go install
+
+### second stage - runner ###
+FROM debian:stretch-slim as runner
```
sjoerdsimons: Move to buster recommended

docker/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+Debos needs virtualization to be enabled on the host and shared with the container.
+
+Check that `kvm` is enabled by running ```ls /dev/kvm```
+
+## Usage
+To build `recipe.yaml`:
+```
+cd <PATH_TO_RECIPE_DIR>
+docker run --rm --device /dev/kvm --group-add $(getent group kvm | cut -d: -f3) -w /recipes -u $(id -u) -i -v $(pwd):/recipes -t debos/debos-docker debos <RECIPE.yaml>
```
sjoerdsimons: kvm group is specific o some operating systems; I'd overall recomend making `/dev/kvm` world writable or add it to uaccess (which is the default anyway on most systems);  So the `--group-add` should probably be dropped and replaced by a note in the kvm setup that it should be ensured to be writable by the user running in the docker container.

recipe/recipe_test.go
```diff
@@ -33,7 +33,7 @@ func TestParse_incorrect_file(t *testing.T) {
 
 	for _, test := range tests {
 		r := recipe.Recipe{}
-		err = r.Parse(test.filename)
+		err = r.Parse(test.filename, false)
```
sjoerdsimons: This is probably worth a seperate MR as it's unrelated to the dockerization

docker/recipes.test.yml
```diff
@@ -0,0 +1,14 @@
+version: '3.4'
```
sjoerdsimons: The commit message for this one is misleading;  All actions can be done in non-fakemachine mode, what you probably mean is that it doesn't use actions which act on iage files which either needs fakemachine *or* access to loop devices

docker/Dockerfile
```diff
@@ -1,56 +1,57 @@
...
-WORKDIR /go/src/github.com/go-debos/debos/
+WORKDIR $GOPATH/src/github.com/go-debos/debos/
 RUN GOOS=linux go build -a cmd/debos/debos.go
 
 ### second stage - runner ###
-FROM debian:stretch-slim as runner
+FROM debian:buster-slim as runner
 
-ARG DEBIAN_FRONTEND=noninteractive
+ARG DEBIAN_FRONTEND
```
sjoerdsimons: Why unseting this one ?

docker/Dockerfile
```diff
@@ -1,56 +1,57 @@
...
+ARG GOPATH
```
sjoerdsimons: ditto

docker/Dockerfile
```diff
@@ -20,38 +20,42 @@ RUN apt-get update && \
...
         e2fsprogs \
+        btrfs-progs \
         bmap-tools \
         # fakemachine runtime dependencies
         qemu-system-x86 \
         qemu-user-static \
         busybox \
         linux-image-amd64 \
-        systemd \
-        dbus && \
```
sjoerdsimons: Why dropping dbus?

docker/README.md
```diff
@@ -0,0 +1,34 @@
+# debos-docker
+
+Docker container for ['debos' tool](https://github.com/go-debos/debos).
+
+## Installation
+```
+docker pull debos/debos-docker
```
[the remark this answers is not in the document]
sjoerdsimons: Can we just call the main one debos such that `docker pull debos` works?

docker/Dockerfile
fdanis-oss: This doesn't unset DEBIAN_FRONTEND but allows the stage to use the global ARG

docker/Dockerfile
fdanis-oss: Based on the other docker try (https://gitlab.collabora.com/docker/debos/blob/master/docker/Dockerfile), dbus doesn't seem to be requested

sjoerdsimons: Explain that in your commit message; I get very wary of spurious changes which are not explained

docker/README.md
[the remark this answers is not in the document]
fdanis-oss: I will replace it by debos/debos

docker/Dockerfile
```diff
@@ -0,0 +1,63 @@
+# Global ARGs shared by all stages
```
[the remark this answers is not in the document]
sjoerdsimons: Debian Buster uses uaccess for `/dev/kvm` so the logged in user can access /dev/kvm; so no. Also my recommendation would be to make kvm world writable otehrwise (like fedora does)

I'm also not a fan of complex command-lines with more bits then you may need. Especially when some things would be tunable for your host system..

For the LABELS please set a seperate patch ? :) The long option are also fine for me fwiw in the example run.

docker/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+        bmap-tools \
+        # fakemachine runtime dependencies
+        qemu-system-x86 \
+        qemu-user-static \
+        busybox \
+        linux-image-amd64 \
+        systemd && \
+    rm -rf /var/lib/apt/lists/*
+
+COPY --from=builder $GOPATH/bin/debos /usr/local/bin/debos
```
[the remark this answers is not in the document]
sjoerdsimons: I'm a bit in two minds about that because it may make it non-obvious that really it's just launching debos and you should add command line arguments to debos after the container run.. Maybe it's more obvious for people using docker containers more, but i've never seen a docker container being used like that?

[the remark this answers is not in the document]
sjoerdsimons: Discussion with some folks it does seem like it's a common way for putting docker containers together; So yes sounds like a good suggestions. Probably worht doing it before merging so we don't end up with the container behaviour changing on updates

docker/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+        dosfstools \
+        e2fsprogs \
+        btrfs-progs \
+        bmap-tools \
+        # fakemachine runtime dependencies
+        qemu-system-x86 \
+        qemu-user-static \
+        busybox \
+        linux-image-amd64 \
+        systemd && \
```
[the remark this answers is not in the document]
sjoerdsimons: Seems like a good plan but not a blocker for merging from my pov.

PR #153, opened 2019-03-20
Ability to override path and architecture for "recipe" action

fdanis-oss:


actions/recipe_test.go
```diff
@@ -314,7 +314,8 @@ actions:
 }
 
 func runTestWithSubRecipes(t *testing.T, test testSubRecipe, templateVars ...map[string]string) actions.Recipe {
-	var context debos.DebosContext
+	var common debos.CommonContext
+	var context = debos.DebosContext { &common, "", "" }
```
sjoerdsimons: Unless you want to use common this can be `context := debos.DebosContext { &debos.CommonContext{}, "", ""}`

actions/recipe_action.go
```diff
@@ -58,19 +62,26 @@ func (recipe *RecipeAction) Verify(context *debos.DebosContext) error {
 
 	// Add Variables to template vars
 	for k, v := range recipe.Variables {
+		/* By default included recipe shared the same arhitecture with its
```
sjoerdsimons: * Architecture

actions/recipe_action.go
```diff
@@ -58,19 +62,26 @@ func (recipe *RecipeAction) Verify(context *debos.DebosContext) error {
...
+		   parent. Overwrite the local context "architecture" in case this
+		   is set explicitly in Variables options */
+		if k == "architecture" {
+			recipe.context.Architecture = v
+		}
+
```
sjoerdsimons: This probably deserved splitting out in its own commit; With a better explanation what it's trying to do as i'm not entirely clear what the semantics of this are..

This seems to try and assert that the architecture is the same as the parent architecture unless it was explicitly set in the Variables for the recipe?

actions/recipe_action.go
```diff
@@ -1,7 +1,21 @@
 /*
 Recipe Action
 
-Include a recipe.
+This action includes the recipe actions at the given path, and can optionally
```
sjoerdsimons:
```suggestion
This action includes the recipe at the given path, and can optionally
```

actions/recipe_action.go
```diff
@@ -1,7 +1,21 @@
...
+overrides or adds new template variables.
```
sjoerdsimons:
```suggestion
override or set template variables.
```

actions/recipe_action.go
```diff
@@ -1,7 +1,21 @@
...
+
+To allow recipes to operate both stand-alone and when included, template
+variable "included_recipe" is added and set to "true" to indicate that the
+recipe is part of a longer chain.
+This allows a recipe to skip the unpack step if it's not the first to be
+executed or to optionally skip a pack action when it's not the last.
```
sjoerdsimons: I'm not clear on how this is meant to work from this documentation? This seems to imply that a recipe can know it's "not the first" or if it's "not the last" when all it gets is a boolean saying it was included.

You probably just want to indicate it's up to the recipe writes on how to use this (I keep going back and forward, but i think in my recipe collection i wouldn't end up using this one but be explicit per recipe in enabling/disabling optional steps)

actions/recipe_action.go
```diff
@@ -1,7 +1,21 @@
...
+
+For compatibility, both the parent recipe and all included recipes have
```
sjoerdsimons:
```suggestion
To ensure compatibility, both the parent recipe and all included recipes have
```

actions/recipe_action.go
```diff
@@ -1,7 +1,21 @@
...
+to be for the same architecture. This is passed to the included recipe as
```
sjoerdsimons:
```suggestion 
to be for the same architecture. For convenience the parent architecture is passed in the
```

actions/recipe_action.go
fdanis-oss:
```
To allow recipes to operate both stand-alone and when included, template
variable "included_recipe" is added and set to "true" to indicate that the
recipe is part of a longer chain.
This allows a recipe to skip or add some actions, if it's not the first to be
executed, by surrounding them with "included_recipe" test. For example, the
pack action of an included recipe can be skipped when its build result is
directly used by its parent recipe.
```
Do you think this is clearer?

sjoerdsimons: No because it doesn't address the core of my complaint.  You're giving examples that cannot be implemented with the provided extra variable as said variable doesn't contain the required information.

sjoerdsimons:
Please rebase this on the latest master; I cherry-picked the good commits out leaving just the documentation commit really.

PR #165, opened 2019-05-29
actions: Support for local packages installation

### LINKED ISSUE #157: Add a new action to install a debian package
It could be useful to be able to install a locally built debian package to the image.

fdanis-oss:
This add a new 'source' option to 'apt' action, allowing to install local
packages from the 'source' directory in addition to the repository.

Fix #157

Signed-off-by: Frédéric Danis <frederic.danis@collabora.com>

### DIFF
--- a/actions/apt_action.go
+++ b/actions/apt_action.go
@@ -7,6 +7,7 @@ Yaml syntax:
  - action: apt
    recommends: bool
    unauthenticated: bool
+   local-packages-dir: directory
    packages:
      - package1
      - package2
@@ -20,10 +21,18 @@ Optional properties:
 - recommends -- boolean indicating if suggested packages will be installed
 
 - unauthenticated -- boolean indicating if unauthenticated packages can be installed
+- local-packages-dir -- directory containing local packages to be installed,
+located in the recipe directory ($RECIPEDIR)
 */
 package actions
 
 import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"strings"
+
 	"github.com/go-debos/debos"
 )
 
@@ -32,6 +41,7 @@ type AptAction struct {
 	Recommends       bool
 	Unauthenticated  bool
 	Packages         []string
+	LocalPackagesDir string `yaml:"local-packages-dir"` // external path containing local packages
 }
 
 func (apt *AptAction) Run(context *debos.DebosContext) error {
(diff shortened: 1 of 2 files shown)


andhe:
bikeshedding: I think "source" may be a bit unfortunate name/keyword for this as it's not obvious what it would mean without a description. eg. it could (incorrectly) be interpretted as source code, source packages, etc. Is there any other more obvious keyword that fits better perhaps? Just thinks it's useful to consider the question now, because once it's in changing it would break recipies which would be bad....

fdanis-oss:
yes, you're right, what about "local-packages-dir"?

actions/apt_action.go
```diff
@@ -49,6 +56,49 @@ func (apt *AptAction) Run(context *debos.DebosContext) error {
 	aptOptions = append(aptOptions, "install")
 	aptOptions = append(aptOptions, apt.Packages...)
 
+	if apt.LocalPackagesDir != "" {
+		localpackagesdir := path.Join(context.RecipeDir, apt.LocalPackagesDir)
+		destination, err := debos.RestrictedPath(context.Rootdir, "/usr/local/debs")
```
obbardc, line 61: is `/usr/local/debs` the best place for this ? would a temp directory be better?
i think it is worth allowing the user to optionally override this path from the recipe; for instance the rootfs may be too small.

actions/apt_action.go
```diff
@@ -49,6 +56,49 @@ func (apt *AptAction) Run(context *debos.DebosContext) error {
...
+		if err != nil {
+			return err
+		}
+		defer os.RemoveAll(destination)
```
obbardc, line 65: do the defers run at the end of the program or the end of the action? (my go is not good)
i think `destination` should be cleaned up after the action has completed. 
subsequent `apt` actions would also have this directory available, best to delete it after the action rather than after the recipe

actions/apt_action.go
```diff
@@ -49,6 +56,49 @@ func (apt *AptAction) Run(context *debos.DebosContext) error {
...
+		if err != nil {
+			return err
+		}
+
+		currentDir, _ := os.Getwd()
+		os.Chdir(destination)
+
+		// apt-ftparchive tries to read /etc/apt/apt.conf.d, add this path temporarily to prevent warnings
+		os.MkdirAll("/etc/apt/apt.conf.d", 0755)
+		defer os.RemoveAll("/etc/apt/apt.conf.d/")
```
obbardc, line 77: if this is only a warning i wonder if this should be a separate patch since debootstrap action also moans about this
also, this is in the middle of a `chdir`, for clarity move it above

docker/Dockerfile
```diff
@@ -54,6 +54,7 @@ LABEL org.label-schema.docker.cmd 'docker run \
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
         apt-transport-https \
+        apt-utils \
```
obbardc, line 57: also add `Recommends: apt-utils` to https://salsa.debian.org/go-team/packages/debos/-/blob/master/debian/control

actions/apt_action.go
```diff
@@ -49,6 +56,49 @@ func (apt *AptAction) Run(context *debos.DebosContext) error {
...
+			return err
+		}
+		err = debos.Command{}.Run("apt", "sh", "-c", "apt-ftparchive release . > Release")
+		if err != nil {
+			return err
+		}
+
+		os.Chdir(currentDir)
+
+		locallist, err := debos.RestrictedPath(context.Rootdir, "/etc/apt/sources.list.d/local.list")
```
obbardc, line 90: `local.list` is quite generic

actions/apt_action.go
```diff
@@ -49,6 +56,49 @@ func (apt *AptAction) Run(context *debos.DebosContext) error {
...
+		if err != nil {
+			return err
+		}
+		defer os.RemoveAll(locallist)
+		cmd := fmt.Sprintf("echo 'deb [trusted=yes] file:///usr/local/debs/ ./' > %s", locallist)
```
obbardc, line 95: again, subsequent `apt` actions would also have this configuration available, best to delete it after the action rather than after the recipe

actions/apt_action.go
```diff
@@ -49,6 +56,49 @@ func (apt *AptAction) Run(context *debos.DebosContext) error {
...
+		err = debos.Command{}.Run("apt", "sh", "-c", cmd)
```
obbardc, line 96: can this be done as a direct go file write rather than a command?

actions/apt_action.go
```diff
@@ -20,10 +21,15 @@ Optional properties:
 - recommends -- boolean indicating if suggested packages will be installed
 
 - unauthenticated -- boolean indicating if unauthenticated packages can be installed
+- local-packages-dir -- directory containing local packages to be installed
```
obbardc, line 24: it's not entirely clear without thinking about it where this directory lives - could make it clear that it is under `$RECIPEDIR`

actions/apt_action.go
obbardc: got it

sjoerdsimons:
What feels incredibly awkward with this action for me is the amount of hakcs it has to do to the root filesystem we're building to "add" the local packages directory; 

And then you have to specify the pakcage to install in your normal apt run; however i don't think you generate a policy file to ensure the local file:// archive gets priority so if the os you build against has a newer version there is a chance it install the non-local package..

Would probably be better to simplify this and rather then extending the `apt` actions have a new action which only install a list of named files (with glob?) using `apt install <files>` which iirc installs dependencies from other apt sources automatically

that should also support the typical `origin` keyword one could easily install packages downloaded externally

fdanis-oss:
afaik `apt install <files>` will only install dependencies from other apt sources, but not from the local directory, and force user to list all packages and their dependencies in the command

obbardc:
Continuing the alternative implementation under #220 so closing this one.

PR #166, opened 2019-06-03
image-partition: Enable filesystem features on fs creation

a-wai:
We might want to enable specific features during a filesystem creation,
as it can be difficult and/or cause issues to enable them afterwards.

This commit introduces a new 'features' field to partition definitions,
and makes sure the corresponding features are enabled when creating the
filesystem.

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -41,6 +41,7 @@ Yaml syntax for partitions:
 	   fs: filesystem
 	   start: offset
 	   end: offset
+	   features: list of filesystem features
 	   flags: list of flags
 	   fsck: bool
 
@@ -61,6 +62,9 @@ For 'start' and 'end' properties offset can be written in human readable
 form -- '32MB', '1GB' or as disk percentage -- '100%'.
 
 Optional properties:
+
+- features -- list of additional filesystem features which need to be enabled
+for partition.
 
 - flags -- list of additional flags for partition compatible with parted(8)
 'set' command.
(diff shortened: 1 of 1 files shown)


actions/image_partition_action.go
```diff
@@ -179,6 +184,21 @@ func (p *Partition) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+func appendFeatures(cmdline []string, features []string) []string {
+	if len(features) > 0 {
+		var tmp strings.Builder
+		for _, o := range features {
+			fmt.Fprintf(&tmp, "%s,", o)
+		}
+		options := tmp.String()
+		// Strip final ","
+		options = options[:tmp.Len()-1]
+		cmdline = append(cmdline, "-O", options)
```
sjoerdsimons: You can use strings.Join to concatinate the list with ',' as seperator; that probalby cuts away half of the code :)

sjoerdsimons: Also can Features not be nil ? I'd expect taht to be the default

a-wai: > You can use strings.Join to concatinate the list with ',' as seperator; that probalby cuts away half of the code :)

Indeed, thanks for the hint!

> Also can Features not be nil ? I'd expect taht to be the default

I'm not sure what you mean here: Features is a member of the `Partition` struct, can it be nil in a valid `Partition` instance ?

gportay:
What is the status of that PR?

PR #168, opened 2019-06-14
Move target bind mount for scripts to tmp

### LINKED ISSUE #151: Cleanup the directory "/script" from target
Action run starting the script in chroot adds the "/script" to target image/tarball.
Need to cleanup that directory in case if it has been created by Debos.

d4s:
`/script` directory is left in target rootfs. Moving the target bind
mount to `/tmp` allows to avoid additional step for removing this directory
for all projects calling own scripts in chrooted environment.

Fixes #151

### DIFF
--- a/actions/run_action.go
+++ b/actions/run_action.go
@@ -98,8 +98,8 @@ func (run *RunAction) doRun(context debos.DebosContext) error {
 		script[0] = debos.CleanPathAt(script[0], context.RecipeDir)
 		if run.Chroot {
 			scriptpath := path.Dir(script[0])
-			cmd.AddBindMount(scriptpath, "/script")
-			script[0] = strings.Replace(script[0], scriptpath, "/script", 1)
+			cmd.AddBindMount(scriptpath, "/tmp/script")
+			script[0] = strings.Replace(script[0], scriptpath, "/tmp/script", 1)
 		}
 		cmdline = []string{strings.Join(script, " ")}
 		label = path.Base(run.Script)


actions/run_action.go
```diff
@@ -98,8 +98,8 @@ func (run *RunAction) doRun(context debos.DebosContext) error {
 		script[0] = debos.CleanPathAt(script[0], context.RecipeDir)
 		if run.Chroot {
 			scriptpath := path.Dir(script[0])
-			cmd.AddBindMount(scriptpath, "/script")
-			script[0] = strings.Replace(script[0], scriptpath, "/script", 1)
+			cmd.AddBindMount(scriptpath, "/tmp/script")
+			script[0] = strings.Replace(script[0], scriptpath, "/tmp/script", 1)
```
sjoerdsimons: This does assume the /tmp in the chroot is on a tmpfs otherwise it'll still leave it around (though in a less awkward location);  Is this guaranteed ? (Otherwise making the cmd functionality explicitly clean up afterwards would be nicer)

d4s: This change is related for `systemd-nspawn` only.
And there is [a commit](https://github.com/systemd/systemd/commit/bbb99c30d01a8bcdc27fb151cc6376a7877a6b07) which explicitly mount `tmpfs` to `/tmp` for container.

sjoerdsimons: Could you note that in the commit message or a comment please? otherwise looks fine

d4s: Done

d4s:
Known problem: fail with systemd 232 (check https://github.com/systemd/systemd/issues/4789).

PR #170, opened 2019-06-22
pack,unpack: Save and restore xattrs and file capabilities in tarballs

em-:
Add options to `tar` to tell it to save and restore extended attributes:
    
* `--xattr`: enable extended attributes support
* `--xattrs-include=*.*`: tell `tar` to include every extended attribute
  since by default `tar` only stores attributes in the `user.*` namespace
    
This fixes the `pack` and `unpack` actions when dealing with tools like
`ping` which on modern distributions have been switched to use file
capabilities to do privileged operations without being setuid, as they are
based on extended attributes.
    
This relies on `tar` being GNU Tar >= 1.27, released in 2013 and shipped since
Debian Jessie.

Tested with the recipe below:
```
architecture: amd64

actions:
  - action: run
    description: Create test file
    chroot: false
    command: touch "${ROOTDIR}/testfile"

  - action: run
    description: Set file capabilities on test file
    chroot: false
    command: setcap cap_sys_pacct=p "${ROOTDIR}/testfile"

  - action: pack
    file: testpack.tar.gz

  - action: run
    description: Clean up test file
    chroot: false
    command: rm "${ROOTDIR}/testfile"

  - action: unpack
    file: testpack.tar.gz

  - action: run
    description: Dump the unpacked file capabilities
    chroot: false
    command: getcap "${ROOTDIR}/testfile"

  - action: run
    description: Check that the file capabilities are set
    chroot: false
    command: getcap "${ROOTDIR}/testfile" | grep -c cap_sys_pacct+p
```

### DIFF
--- a/actions/pack_action.go
+++ b/actions/pack_action.go
@@ -35,5 +35,7 @@ func (pf *PackAction) Run(context *debos.DebosContext) error {
 	outfile := path.Join(context.Artifactdir, pf.File)
 
 	log.Printf("Compressing to %s\n", outfile)
-	return debos.Command{}.Run("Packing", "tar", "czf", outfile, "-C", context.Rootdir, ".")
+	return debos.Command{}.Run("Packing", "tar", "czf", outfile,
+		"--xattrs", "--xattrs-include=*.*",
+		"-C", context.Rootdir, ".")
 }
--- a/TODO
+++ b/TODO
@@ -41,8 +41,6 @@ TODO
 * Make actions using (host) commands check their existance early
 
 
-* Ensure we copy xattrs?
-
 * Fix race in qemu-helper (if qemu-user-static gets installed in the system
   chroot  things will get confused)
 
--- a/archiver.go
+++ b/archiver.go
@@ -99,6 +99,8 @@ func (tar *ArchiveTar) Unpack(destination string) error {
 	}
 	command = append(command, "-C", destination)
 	command = append(command, "-x")
+	command = append(command, "--xattrs")
+	command = append(command, "--xattrs-include=*.*")
 
 	if compression, ok := tar.options["tarcompression"]; ok {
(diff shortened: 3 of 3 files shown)


actions/pack_action.go
```diff
@@ -35,5 +35,7 @@ func (pf *PackAction) Run(context *debos.DebosContext) error {
 	outfile := path.Join(context.Artifactdir, pf.File)
 
 	log.Printf("Compressing to %s\n", outfile)
-	return debos.Command{}.Run("Packing", "tar", "czf", outfile, "-C", context.Rootdir, ".")
+	return debos.Command{}.Run("Packing", "tar", "czf", outfile,
+		"--xattrs", "--xattrs-include=security.capability",
```
sjoerdsimons: should it really only include security.capability? I can imagine recipes creating others as well.

em-: I initially saved all the xattrs, because why not.

Then it happened that I'm running with SELinux enabled and then all the SELinux labels from my system were captured as well, which is rather undesirable.

So I decided to play safe and include only the thing I specifically care about in this round. I suppose some configurability would be needed for more complicated use-cases, but I tried to stay away from that rabbit hole.

sjoerdsimons: That's surprising as debos will build in either RAM on the VM or an extra image (scratch image) ; i wonder what's adding the selinux labels?

Did you test without fakemachine maybe?

em-: IIRC I did some tests with `go test` and that's probably where the xattrs got mixed up. They indeed don't appear when using fakemachine from, say, the Apertis container.

em-: Updated the MR to use `*.*` and save all xattrs.

sjoerdsimons:
Can you run go fmt on this :) otehrwise looks good

sjoerdsimons:
apparently you did; i just suck at go standard formatting :)

Issue #48, opened 2017-11-15
commands: systemd-nspawn changes resolv.conf in target environment


d4s:
During call of any command/script in chroot environment 'systemd-nspawn' copies resolv.conf from host's environment.

PR #186, opened 2019-12-04
Protect resolv.conf from changes caused by systemd-nspawn

d4s:
Usage of `systemd-nspawn` for starting commands and scripts in container have side effect causing the copying of 'resolv.conf' from the build host by default.
The similar behavior is used by OS scripts from `debootstrap` -- they do the copy of    resolv.conf from the build host.

Such behavior prepare potentially incorrect configuration of the target rootfs.

- Remove 'resolv.conf' file after the `debootstrap` calls
- Protect the `resolv.conf` for Run action according https://github.com/go-debos/debos/pull/186#issuecomment-562064172

Fixes: #48


sjoerdsimons:
I don't like all the options needed for this; I'd really prefer if it could do the right thing without needing options.

The annoying thing with resolv.conf is that we open need hostname resolving to work during the image builds (which users expects) but at the same time it will be expected that all resolv.conf changes will actually stick.. However these details should really be something debos handles so it just does the right thing (we can add options if the right thing isn't obvious or if there are exceptions where you do not want it to do the typical right thing).. For extra fun resolv.conf can be both a file and a symlink!

Now the good thing is that for both problematic actions we can expect what happens the filesystem ends up with a copy of the resolv.conf of the host. For the run action we can actually make it even more predictable by having debos generate the resolvconf file and pass (`--resolv-conf=off`) to nspawn.

So what i think we should do is:
For the  debootstrap action: Always remove the resolv.conf at the end

For the run action (chrooted):
* save the resolv.conf information in the filesystem (e.g symlink destination or raw file content)
* Let debos copy the host resolv.conf and add some unique cookie in it (e.g. a "generated by debos" sring")
* run nspawn with `--resolv-conf=off`
* After the run if the filesystem still has the debos generated (e.g. check checksum) file put back the
  original otherwise leave it alone (the user changed it)

That way we should always be doing the right way; I do not think users relying on the resolv.conf as copied from the host as side-effect is valid.

sjoerdsimons:
oh btw this should be done in the chroot helper not in the run action as other action can have the same effect (e.g. apt)

d4s:
Agree for debootstap -- that's more predictable and saving of resolv.conf is not needed at this stage.

> I don't like all the options needed for this; I'd really prefer if it could do the right thing without needing options.

initially it was done in similar way -- no need to use any option to protect resolv.conf from changes for chrooted commands. Option for Run allows to remove the "protection" flag from resolv.conf, i.e. do not save/restore that file.

Not sure for the proposed Run logic. There is no problem to detect if the file has changed during the script/cmd call. But it wouldn't work properly in case of several commands in a raw. For instance:
- the 1-st Run action change the content of `resolv.conf` -- here we have some changed host's resolv.conf or completely new version (doesn't matter)
- the 2-nd action will use the host's version again instead of changed version, since debos will save the original version and copy the host's version instead

The magic string looks redundant for changes detection -- since we do a copy of host's file at the beginning -- it is enough to compare files (or checksums)  after the command execution to detect if user has changed the content of the resolv.conf file.

@sjoerd how about a different approach with globally changed behavior:
debos protect the original resolv.conf file and uses the host's file until the original version doesn't changed by any script/command called in chrooted environment.
After that we always use the changed resolv.conf in chrooted actions and do not copy host's version anymore -- assume the user will take care about the resolv.conf from that time.

Add the global flag CustomResolvConf to indicate what user provided (or changed) the target resolv.conf, and for each command we do:
- if flag is set -- just warn the user about the custom version and leave resolv.conf as is
- else:
  - save the original resolv.conf version
  - copy host's version inside the chrooter rootfs
  - ... chrooted command/script execution
  - check if file has been changed and either trigger the CustomResolvConf flag to indicate we will use custom version, or restore saved.

> oh btw this should be done in the chroot helper not in the run action as other action can have the same effect (e.g. apt)

It is done in that way from the beginning -- the option in Run action is needed in fact as transitional variable to pass the value from recipe to command executor. However with my initial approach I miss the corresponding option for the Apt action.

sjoerdsimons:
> Not sure for the proposed Run logic. There is no problem to detect if the file has changed during the script/cmd call. But it wouldn't work properly in case of several commands in a raw. For instance:
> 
> * the 1-st Run action change the content of `resolv.conf` -- here we have some changed host's resolv.conf or completely new version (doesn't matter)
> * the 2-nd action will use the host's version again instead of changed version, since debos will save the original version and copy the host's version instead

Sorry i don' t understand the issue here; For every action where debos does an nspawn it would do the same dance for resolv.conf (save original if any, put host in place, do the action, put the original back if applicable). It doesn't matter how many actions you run; the invariant stays in place.

> The magic string looks redundant for changes detection -- since we do a copy of host's file at the beginning -- it is enough to compare files (or checksums) after the command execution to detect if user has changed the content of the resolv.conf file.

The reason for the "magic string" suggestion was to be able to detect the corner-case where the user changes resolv.conf to the same as the host version; Which is not entirely unlikely as there are some common ones (e.g. the resolved stub). Adding a magic string in (e.g. `# temporary resolv.conf from debos`) makes this corner basically impossible to hit.

> @sjoerd how about a different approach with globally changed behavior:
> debos protect the original resolv.conf file and uses the host's file until the original version doesn't changed by any script/command called in chrooted environment.
> After that we always use the changed resolv.conf in chrooted actions and do not copy host's version anymore -- assume the user will take care about the resolv.conf from that time.

Too error-prone; the resolv.conf the user puts in may not be suitable for the debos environment.

> Add the global flag CustomResolvConf to indicate what user provided (or changed) the target resolv.conf, and for each command we do:

Seems awkward for the user; so lets not do that.

d4s:
> > Not sure for the proposed Run logic. There is no problem to detect if the file has changed during the script/cmd call. But it wouldn't work properly in case of several commands in a raw. For instance:
> > 
> > * the 1-st Run action change the content of `resolv.conf` -- here we have some changed host's resolv.conf or completely new version (doesn't matter)
> > * the 2-nd action will use the host's version again instead of changed version, since debos will save the original version and copy the host's version instead
> 
> Sorry i don' t understand the issue here; For every action where debos does an nspawn it would do the same dance for resolv.conf (save original if any, put host in place, do the action, put the original back if applicable). It doesn't matter how many actions you run; the invariant stays in place.

The problem is in custom file.
In case if user want to use own `resolv.conf` with custom nameserver during the Run action what he need to do?

With approach you've described it looks like it wouldn't be possible:
On the first run the `resolv.conf` has been changed; we detect that change and leave changed version as is with added nameserver.
But on the second run we save this changed file as "original" and substitute with host's version. So all changes added on the previous step including custom nameserver would be lost. And the second chrooted run will use the host's config again.
Even worse -- there is a corner case with changing the 'resolv.conf' twice in raw -- in fact the first changed file became an 'original' at the second run, so one of two changes would be lost.

> The reason for the "magic string" suggestion was to be able to detect the corner-case where the user changes resolv.conf to the same as the host version; Which is not entirely unlikely as there are some common ones (e.g. the resolved stub). Adding a magic string in (e.g. `# temporary resolv.conf from debos`) makes this corner basically impossible to hit.

right, thanks for the hint about this corner case.

> > @sjoerd how about a different approach with globally changed behavior:
> > debos protect the original resolv.conf file and uses the host's file until the original version doesn't changed by any script/command called in chrooted environment.
> > After that we always use the changed resolv.conf in chrooted actions and do not copy host's version anymore -- assume the user will take care about the resolv.conf from that time.
> 
> Too error-prone; the resolv.conf the user puts in may not be suitable for the debos environment.

So what? If user do changes in `resolv.conf` -- it is expected what he could do a mistake in network configuration. To my mind this is something what should be managed by user, so if he will made an error -- it's his own error.
btw that is why I initially propose an additional parameter for the Run action -- hint for the debos which resolv.conf version user prefer for the action -- host's (by default) or custom from chrooted environment.

> > Add the global flag CustomResolvConf to indicate what user provided (or changed) the target resolv.conf, and for each command we do:
> Seems awkward for the user; so lets not do that.

Sorry, I probably wasn't clear enough -- that flag is not for the user. This is for Debos internals, i.e. we always work with host's resolv.conf until user explicitly did a change of the file in chrooted environment. Use only user's changed version after that.

d4s:
That's what I mean -- please check the result of actions on steps 4-6
```
2019/12/11 00:40:03 ==== 1. should not exists after debootsrap ====
2019/12/11 00:40:03 ==== 2. check if file is copied from the host and containd additional string by debos ====
2019/12/11 00:40:03 cat /etc/resolv.conf | # Automatically generated by Debos
2019/12/11 00:40:03 cat /etc/resolv.conf | domain amber.lxd
2019/12/11 00:40:03 cat /etc/resolv.conf | search amber.lxd
2019/12/11 00:40:03 cat /etc/resolv.conf | nameserver 10.0.103.1
2019/12/11 00:40:03 ==== 3. should not exists ====
2019/12/11 00:40:03 ==== 4. create own resolv.conf ====
2019/12/11 00:40:03 ==== 5. check the content of resolv.conf in chrooted environment ====
2019/12/11 00:40:03 cat "$ROOTDIR"/etc/resolv.conf | # Automatically generated by Debos
2019/12/11 00:40:03 cat "$ROOTDIR"/etc/resolv.conf | domain amber.lxd
2019/12/11 00:40:03 cat "$ROOTDIR"/etc/resolv.conf | search amber.lxd
2019/12/11 00:40:03 cat "$ROOTDIR"/etc/resolv.conf | nameserver 10.0.103.1
2019/12/11 00:40:03 ==== 6. ensure we still have own file ====
2019/12/11 00:40:03 cat "$ROOTDIR"/etc/resolv.conf | nameserver 8.8.8.8
2019/12/11 00:40:03 ==== 7. change resolv.conf in chrooted environment ====
2019/12/11 00:40:04 ==== 8. check the file contain our new changes ====
2019/12/11 00:40:04 cat "$ROOTDIR"/etc/resolv.conf | nameserver 4.4.4.4
2019/12/11 00:40:04 ==== 9. remove resolve.conf in chrooted environment ====
2019/12/11 00:40:04 ==== 10. should not exists anymore ====
2019/12/11 00:40:04 ==== 11. symlink resolv.conf ====
2019/12/11 00:40:04 ln -s /lib/systemd/resolv.conf "$ROOTDIR"/etc/resolv.conf && ls -l "$ROOTDIR"/etc/resolv.conf | lrwxrwxrwx 1 root root 24 Dec 11 00:40 /scratch/root/etc/resolv.conf -> /lib/systemd/resolv.conf
2019/12/11 00:40:04 ==== 12. check the content of resolv.conf in chrooted environment ====
2019/12/11 00:40:04 cat /etc/resolv.conf | # Automatically generated by Debos
2019/12/11 00:40:04 cat /etc/resolv.conf | domain amber.lxd
2019/12/11 00:40:04 cat /etc/resolv.conf | search amber.lxd
2019/12/11 00:40:04 cat /etc/resolv.conf | nameserver 10.0.103.1
2019/12/11 00:40:04 ==== 13. ensure we still have symlink ====
2019/12/11 00:40:04 ls -l "$ROOTDIR"/etc/resolv.conf | lrwxrwxrwx 1 root root 24 Dec 11 00:40 /scratch/root/etc/resolv.conf -> /lib/systemd/resolv.conf
2019/12/11 00:40:04 ==== 14. remove the saved copy in chroot ====
2019/12/11 00:40:04 ==== 15. should not exists anymore ====
```
The prepared [test.yaml](https://github.com/go-debos/debos/files/3947860/test.txt) file is attached

sjoerdsimons:
I don't see the user wanting to use their own resolv.conf during the run action as valid use-case; The goal for the resolv.conf wrt. chrooted run is such that the network access from the host (fakemachine) and the build system are equivalent under the assumption that network access works from the host in the first place (otherwise we have bigger issues).

If there are good use-cases for a debos run not using the standard fakemachine resolv.conf then that should probably be a global option of debos to use an alternative file; Such that it's also used for e.g. debootstrap, but to be honest i don't really see why you'd want that.. But in any case that is a property of the debos run, not of the recipe.

 So the results from the test run you're doing seem fine to me.

commands.go
```diff
@@ -152,10 +227,19 @@ func (cmd Command) Run(label string, cmdline ...string) error {
 	}
 
 	// Disable services start/stop for commands running in chroot
+	// Save the original resolv.conf and copy version from host
 	if cmd.ChrootMethod != CHROOT_METHOD_NONE {
 		services := ServiceHelper{cmd.Chroot}
 		services.Deny()
 		defer services.Allow()
+
```
sjoerdsimons: spurios newline?

commands.go
```diff
@@ -152,10 +227,19 @@ func (cmd Command) Run(label string, cmdline ...string) error {
 	}
 
 	// Disable services start/stop for commands running in chroot
+	// Save the original resolv.conf and copy version from host
```
sjoerdsimons: shouldn't that comment be in the block handling this?

commands.go
```diff
@@ -152,10 +227,19 @@ func (cmd Command) Run(label string, cmdline ...string) error {
...
 	if cmd.ChrootMethod != CHROOT_METHOD_NONE {
 		services := ServiceHelper{cmd.Chroot}
 		services.Deny()
 		defer services.Allow()
+
+	}
+
+	// We use own resolv.conf handling
```
sjoerdsimons: and this one with the nspawn call?

commands.go
```diff
@@ -152,10 +227,19 @@ func (cmd Command) Run(label string, cmdline ...string) error {
...
+	if sum, err := cmd.saveResolvConf(); err != nil {
+		return err
+	} else {
+		defer cmd.restoreResolvConf(sum)
```
sjoerdsimons: This will ignore errors when restoring the resolv.conf;  I don't think that's wanted.

actions/debootstrap_action.go
```diff
@@ -3,6 +3,8 @@ Debootstrap Action
 
 Construct the target rootfs with debootstrap tool.
 
+NB: file "/etc/resolv.conf" will be removed after execution.
```
sjoerdsimons: This probably need a bit more explanation as reading it without knowing the details this seems odd

actions/debootstrap_action.go
```diff
@@ -68,7 +70,7 @@ func NewDebootstrapAction() *DebootstrapAction {
 	// Be secure by default
 	d.CheckGpg = true
 	// Use main as default component
-	d.Components = []string {"main"}
+	d.Components = []string{"main"}
```
sjoerdsimons: Would be awesome of style changes could be doe in a seperate patch :)

d4s:
Hm... my assumption was to allow user to use own config during the chrooted call. But ok, as soon as we have such request -- it would be easy to add appropriate flag.
PS I'll add formatting changes as separate PR.

actions/debootstrap_action.go
```diff
@@ -174,6 +178,14 @@ func (d *DebootstrapAction) Run(context *debos.DebosContext) error {
 	}
 	srclist.Close()
 
+	/* Clenup resolv.conf after debootstrap */
```
sjoerdsimons:
```suggestion
	/* Cleanup resolv.conf after debootstrap */
```

actions/debootstrap_action.go
```diff
@@ -3,6 +3,10 @@ Debootstrap Action
 
 Construct the target rootfs with debootstrap tool.
 
+Please keep in mind -- file `/etc/resolv.conf` will be removed after execution.
+The most of OS scripts used by `debootstrap` do the copy of `resolv.conf` from the host,
```
sjoerdsimons:
```suggestion
Most of the OS scripts used by `debootstrap` copy `resolv.conf` from the host,
```

actions/debootstrap_action.go
```diff
@@ -3,6 +3,10 @@ Debootstrap Action
...
+and this leads to incorrect configuration, which may became the part of the created rootfs.
```
sjoerdsimons:
```suggestion
and this may lead to incorrect configuration when becoming part of the created rootfs.
```

commands.go
```diff
@@ -115,6 +117,83 @@ func (cmd *Command) AddBindMount(source, target string) {
...
+	hostconf := "/etc/resolv.conf"
+	chrootedconf := path.Join(cmd.Chroot, hostconf)
+	savedconf := chrootedconf + ".debos"
+	var sum [sha256.Size]byte
+
+	if cmd.ChrootMethod == CHROOT_METHOD_NONE {
+		return nil, nil
+	}
+
+	// It is expected what there were no resolv.conf before
```
sjoerdsimons:
```suggestion
	// There may not be an existing resolv.conf
```

commands.go
```diff
@@ -158,11 +238,25 @@ func (cmd Command) Run(label string, cmdline ...string) error {
...
 
-	err := exe.Run()
+	// Save the original resolv.conf and copy version from host
+	resolvsum, err := cmd.saveResolvConf()
+	if err != nil {
+		return err
+	}
+
+	if err = exe.Run(); err != nil {
+		return err
```
sjoerdsimons: Previously we'd run flush and cleanup in case Run errored out; Now it bails immediately? Was that change intentional ?

d4s: Nice catch! added defer for both

Issue #190, opened 2019-12-22
Debos docker error: open /scratch/root/etc/resolv.conf: no such file or directory


forslund:
This is probably a user error but I can't figure out the cause of this error...

We're switching to the official docker image from the old one by @JBBgameich 
After installing a group of packages the recipe is terminated with the error message

```
2019/12/21 12:08:29 Action `recipe` failed at stage Run, error: open /scratch/root/etc/resolv.conf: no such file or directory
```
[Full logs](https://termbin.com/5tr0)

I wonder if it's the new automatic restore of the resolve.conf that's the issue? Maybe We're not running some action that's needed?

Out recipe can be found [here](https://github.com/MycroftAI/mycroft-devices/tree/new-docker-image)

And we launch the docker container with

```
docker run --rm --device /dev/kvm --user $(id -u)  --group-add=$(getent group kv
m | cut -d : -f 3) --workdir /recipes --mount "type=bind,source=$(pwd),destinati
on=/recipes" --security-opt label=disable godebos/debos mycroft-mark2-rpi4.yml -
m 5324M
```

d4s:
Hello, thank you for reporting the issue -- I'm able to reproduce it with your instructions! Will check the root cause at the first glance it looks like #186 related indeed.

d4s:
Hm... Thank you @forslund!
Definitely have a corner case here -- after the packages installation during "Base packages" stage I see the state of `resolv.conf` is:
```
root@root:~# ls -l /etc/resolv.conf 
lrwxrwxrwx 1 root root 39 Dec 22 16:21 /etc/resolv.conf -> ../run/systemd/resolve/stub-resolv.conf
root@root:~# ls -l ../run/systemd/resolve/stub-resolv.conf
/usr/bin/ls: cannot access '../run/systemd/resolve/stub-resolv.conf': No such file or directory
```
Some package/script/whatever set the `resolv.conf` as a symlink to `stub-resolv.conf` which exists only during the call of `systemd-nspawn`. After the execution it is pointed to non-existing file and fail too strict check on action exit.

forslund:
That's some impressive detective work and quick response!

Let me know if you need any testing done.

d4s:
@forslund thanks, no need for additional testing yet. It is a real corner case which opened a black hole for another corner case ;)
Will propose the fix today or tomorrow. Will ensure I'm able to build your recipe before submitting.

d4s:
[...]

PR #192, opened 2019-12-23
commands: check if the chrooted resolv.conf is a symlink

d4s:
If the resolv.conf in chrooted environment changed to symlink and point to
non-existing file, the os.Stat() fail the check of file existence.
Use os.Lstat() call instead and add additional checks during the resolv.conf
restore phase.

Fixes: #190


d4s:
Tested on Apertis builds and build described in #190

fdanis-oss:
This is OK for me

PR #198, opened 2020-02-07
Reverse the exitcode initialization value

gportay:
Fakemachine is subject to panic and causes Debos to exit success due to
the current logic of the exitcode. For example, the fakemachine function
CopyFileTo() panics if the file is missing.

In the case of a panic, the function never returns. Thus, the exitcode
cannot be set to 1 and Debos exits with 0.

This commit reverses the logic of the exitcode: it is initialized to 1
(i.e. failure), and it is set to 0 (i.e. success) only if Debos has
reached the end of the things it has to do (or for the help message).

Signed-off-by: Gaël PORTAY <gael.portay@collabora.com>

### DIFF
--- a/cmd/debos/debos.go
+++ b/cmd/debos/debos.go
@@ -87,7 +87,7 @@ func main() {
 		"no_proxy",
 	}
 
-	var exitcode int = 0
+	var exitcode int = 1
 	// Allow to run all deferred calls prior to os.Exit()
 	defer func() {
 		os.Exit(exitcode)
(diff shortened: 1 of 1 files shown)


sjoerdsimons:
The code makes sense to me; would need some testing of the various failure and test cases and with both kvm and non-fakemachine  runs

gportay:
> The code makes sense to me; would need some testing of the various failure and test cases and with both kvm and non-fakemachine runs

I can do some testing if needed. Just tell me how to test it.

obbardc:
superseded by #278

Issue #178, opened 2019-09-05
Filesystem-deploy should contain more information


sangorrin:
I think that the action filesystem-deploy should contain information about how to deploy the chroot contents into the partitions (mountpoints should be used only to define fstab). For example, something like this:

```
  - action: filesystem-deploy
    description: Deploying filesystem onto image
    partitions:
      - name: EFI
        files: /efi
      - name: ROOTFS
        files: /
        blacklist: /efi, /var
      - name: FACTORYRESET
        files: /
      - name: VAR
        files: /var
```

Also, mountpoints should have its own action, separated from image-partition. This will allow, specifying mountpoints for multiple rootfs partitions (e.g., factory reset rootfs, A/B software updates rootfs, chrootfs, ...)

```
- action: mountpoints
  root-partitions: 
      - name: ROOTFS
        mountpoints:
          - mountpoint: /
	       partition: ROOTFS
          - mountpoint: /var
	       partition: VAR
      - name: FACTORYRESET
          - mountpoint: /
	       partition: FACTORYRESET
```

Also these options should be removed from filesystem-deploy and put into the mountpoints action so that we can generate fstab/cmdline on each root partition.
- setup-fstab: bool
- setup-kernel-cmdline: bool
- append-kernel-cmdline: arguments

Thanks,
Daniel

obbardc:
in your case with two rootfs, I would suggest mounting into seperate locations, and setting `buildtime` on one of them.

It's not practical to mount to partitions in the same location.

sangorrin:
@obbardc thanks for the suggestion. Just in case, let me explain the problem again and see if I understood you:

* I want to build an image that has 2 root partitions with the exact same contents (except the `fstab` file): ROOTFS and FACTORYRESET. (you can think of FACTORYRESET as a recovery partition in case ROOTFS stops working after an update)
* I want to be able to boot to either of those partitions, so they need to have separate `fstab` files. One of them will map ROOTFS's UUID to the mount point "/", and the other one will map FACTORYRESET's UUID to the mount point "/".

I believe that you are suggesting that I mount FACTORYRESET in a temporary location with the `buildtime` flag enabled (so that it doesn't show in ROOTFS's fstab); and then copy manually the files from ROOTFS to FACTORYRESET, and generate FACTORYRESET also manually (e.g. copy it and then use sed to change the UUID). Did I understand you correctly?.

[...]

Issue #176, opened 2019-09-05
Deploying twice does not work


sangorrin:
Hi,

It seems that you can only deploy the same file once while the actions' syntax suggests that you should be able to deploy them as many times as you want.

This is an example of an image with a redundant partition that can be used as a fallback filesystem or factory reset partition.

```
  - action: image-partition
    imagename: {{ $image }}
    imagesize: {{ $size }}
    partitiontype: gpt
    partitions:
      - name: EFI
        fs: fat32
        start: 0%
        end: 128MiB
        flags: [ esp ]
      - name: ROOTFSA
        fs: ext4
        start: 128MiB
        end: 1920MiB
      - name: ROOTFSB
        fs: ext4
        start: 1920MiB
        end: 3456MiB
    mountpoints:
      - mountpoint: /
        partition: ROOTFSA
        options: [ ro, noatime ]
      - mountpoint: /
        partition: ROOTFSB
        options: [ ro, noatime ]
      - mountpoint: /efi
        partition: EFI
        buildtime: true

  - action: filesystem-deploy
    description: Deploying filesystem onto image
```

In other words, the contents of "/" should be deployed to both partitions ROOTFSA and ROOTFSB. Unfortunately, the current behavior is that ROOTFSB remains empty.

Thanks,
Daniel

stappersg:
```
mount /dev/fsA  /
mount /dev/fsB  /
```

PR #208, opened 2020-04-30
actions/image_partition: do not allow duplicate partitions or mountpoints

obbardc:
A common user complaint is the result of mounting a filesystem on top of another
can be confusing. So do not allow duplicate mountpoints.
    
Also better not allow multiple partitions with the same name.

Resolves: #176
Resolves: #178

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -52,7 +52,8 @@ Yaml syntax for partitions:
 Mandatory properties:
 
 - name -- is used for referencing named partition for mount points
-configuration (below) and label the filesystem located on this partition.
+configuration (below) and label the filesystem located on this partition. Must be
+unique.
 
 - fs -- filesystem type used for formatting.
 
@@ -86,17 +87,18 @@ Yaml syntax for mount points:
 
 Mandatory properties:
 
-- partition -- partition name for mounting.
+- partition -- partition name for mounting. The partion must exist under `partitions`.
 
 - mountpoint -- path in the target root filesystem where the named partition
-should be mounted.
+should be mounted. Must be unique, only one partition can be mounted per
+mountpoint.
 
 Optional properties:
 
 - options -- list of options to be added to appropriate entry in fstab file.
 
 - buildtime -- if set to true then the mountpoint only used during the debos run.
-No entry in `/etc/fstab' will be created.
+No entry in `/etc/fstab` will be created.
 The mountpoints directory will be removed from the image, so it is recommended
 to define a `mountpoint` path which is temporary and unique for the image,
 for example: `/mnt/temporary_mount`.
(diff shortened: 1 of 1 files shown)


actions/image_partition_action.go
```diff
@@ -413,14 +415,24 @@ func (i ImagePartitionAction) Run(context *debos.DebosContext) error {
...
 	os.MkdirAll(context.ImageMntDir, 0755)
-	for _, m := range i.Mountpoints {
+	for idx, m := range i.Mountpoints {
 		dev := i.getPartitionDevice(m.part.number, *context)
 		mntpath := path.Join(context.ImageMntDir, m.Mountpoint)
+
+		// check if the recipe already mounted something at mountpoint
+		for _, mntpoint := range i.Mountpoints {
+			if (mntpoint.Mountpoint == m.Mountpoint && mntpoint.mounted) {
+				return fmt.Errorf("%s mount failed: mountpoint at %s already exists", m.part.Name, m.Mountpoint)
```
sjoerdsimons, line 425: This should be done in Verify instead; Such we can fail early before spending minutes in the recipe :) Also just error out if there are duplicated mountpoints whether it was mounted is irrelevant

actions/image_partition_action.go
```diff
@@ -413,14 +415,24 @@ func (i ImagePartitionAction) Run(context *debos.DebosContext) error {
...
+			}
+		}
+
 		os.MkdirAll(mntpath, 0755)
 		err := syscall.Mount(dev, mntpath, m.part.FS, 0, "")
 		if err != nil {
 			return fmt.Errorf("%s mount failed: %v", m.part.Name, err)
 		}
+
+		i.Mountpoints[idx].mounted = true
```
sjoerdsimons, line 435: IN which case you also don't need this anymore.

actions/image_partition_action.go
```diff
@@ -540,6 +542,18 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
...
+
+		// check for duplicate partition names
+		duplicates := 0
+		for j := 0; j < len(i.Partitions); j++ {
+			if i.Partitions[j].Name == p.Name {
+				duplicates++
+			}
+		}
+		if duplicates > 1 {
+			return fmt.Errorf("Partition %s already exists", p.Name)
```
sjoerdsimons, line 554: why not return direlty in the for loop?

actions/image_partition_action.go
```diff
@@ -540,6 +542,18 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
...
 		}
+
+		// check for duplicate partition names
+		duplicates := 0
+		for j := 0; j < len(i.Partitions); j++ {
+			if i.Partitions[j].Name == p.Name {
+				duplicates++
+				if duplicates > 1 {
+					return fmt.Errorf("Partition %s already exists", p.Name)
+				}
```
sjoerdsimons, line 553:
```suggestion
				return fmt.Errorf("Partition %s already exists", p.Name)
```

actions/image_partition_action.go
```diff
@@ -557,6 +571,18 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
...
 		m := &i.Mountpoints[idx]
+
+		// check for duplicate mountpoints
+		duplicates := 0
+		for j := 0; j < len(i.Mountpoints); j++ {
+			if i.Mountpoints[j].Mountpoint == m.Mountpoint {
+				duplicates++
+				if duplicates > 1 {
+					return fmt.Errorf("Mountpoint %s already exists", m.Mountpoint)
+				}
```
sjoerdsimons, line 582:
```suggestion
				return fmt.Errorf("Mountpoint %s already exists", m.Mountpoint)
```

Issue #175, opened 2019-09-05
behavior of run action after filesytem-deploy


sangorrin:
Hi,

I think that run actions that occur _after_ the filesystem-deploy action should not affect the contents of the image. In other words, once the filesystem is deployed to the loop device, changes to the chroot environment should not be deployed.

Current behavior does deploy those files. For example, the following sequence will produce an image with "/etc/packages.txt" despite I haven't deployed that file into the image.

```
  - action: filesystem-deploy
    description: Deploying filesystem onto image

  - action: run
    chroot: true
    command: dpkg --list > /etc/packages.txt
```

Is that a misunderstanding from my side?

Thanks
Daniel

obbardc:
all that `action: filesystem-deploy` does is copy the files from a temporary dir to the image which has been partitioned. it can be called at any time.
After it has been called, any actions are switched to work directly on on the deployed image.

If that answers your question, please feel free to close the ticket.

sangorrin:
@obbardc thanks for the explanation. It might be a good idea to mention that "switch" in the documentation. I believe that the behavior of the action "run" (with chroot enabled) changes significantly enough to mention it. For example, it is only after the "switch" that we are able to do stuff like `grub-install`.

PR #210, opened 2020-05-19
docs: action: filesystem-deploy: clarify what happens after deploy

obbardc:
All the filesystem-deploy action does is copy the contents from scratch into the
mounted image, then all actions after that refer to the image rather than the
scratch space. Be clearer.

Resolves: #175

### DIFF
--- a/actions/filesystem_deploy_action.go
+++ b/actions/filesystem_deploy_action.go
@@ -1,8 +1,13 @@
 /*
 FilesystemDeploy Action
 
-Deploy prepared root filesystem to output image. This action requires
-'image-partition' action to be executed before it.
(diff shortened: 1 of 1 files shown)


actions/filesystem_deploy_action.go
```diff
@@ -1,9 +1,13 @@
 /*
 FilesystemDeploy Action
 
-Deploy prepared root filesystem to output image. This action requires
+Deploy prepared root filesystem to output image by copying the files from the
+temporary scratch directory to the mounted image. This action requires
 'image-partition' action to be executed before it.
```
sjoerdsimons, line 6: maybe add that it also configures optionally various files for the image (fstab etc)

Issue #154, opened 2019-04-11
Partition tables are not updated after ImagePartition / FilesystemDeploy


elboulangero:
After deploying a filesystem, a quick look at `/dev/disk/by-id` shows that the kernel is not aware of the new partitions. This breaks tools that rely on that, for example `grub-probe` won't find the root partition as expected, and then `grub-mkconfig` won't be able to setup the kernel command-line with `root=UUID=...`. Instead it will fall back to `root=/dev/...`.

Workaround: One must manually run `partprobe`.

Suggestion: Debos could run partprobe by itself after creating the partitions?

----

Here's a quick recipe to illustrate this issue and the workaround:

```
{{- $suite := or .suite "stretch" -}}
  
architecture: amd64

actions:
  - action: debootstrap
    suite: {{ $suite }}
    components: [ main ]

  - action: apt
    packages: [ parted ]

  - action: image-partition
    imagename: {{ $suite }}.img
    imagesize: 2GB
    partitiontype: gpt
    partitions:
      - name: efi
        fs: fat32
        start: 0%
        end: 128MB
        flags: [ esp ]
      - name: boot
        fs: ext2
        start: 128MB
        end: 256MB
      - name: root
        fs: ext4
        start: 256MB
        end: 100%
    mountpoints:
      - mountpoint: /
        partition: root
      - mountpoint: /boot
        partition: boot
      - mountpoint: /efi
        partition: efi

  - action: run
    chroot: true
    command: ls -l /dev/disk/by-id

  - action: run
    chroot: true
    command: partprobe

  - action: run
    chroot: true
    command: ls -l /dev/disk/by-id
```

And the relevant part of debos output:

```
==== run ====
ls -l /dev/disk/by-id | total 0
ls -l /dev/disk/by-id | lrwxrwxrwx 1 root root 9 Apr 11 03:30 ata-QEMU_DVD-ROM_QM00003 -> ../../sr0
ls -l /dev/disk/by-id | lrwxrwxrwx 1 root root 9 Apr 11 03:30 virtio-fakedisk-0 -> ../../vda
==== run ====
==== run ====
ls -l /dev/disk/by-id | total 0
ls -l /dev/disk/by-id | lrwxrwxrwx 1 root root  9 Apr 11 03:30 ata-QEMU_DVD-ROM_QM00003 -> ../../sr0
[...]

PR #213, opened 2020-05-26
actions/image_partition: trigger udev rules after image-partition completion

obbardc:
The image is locked while partitioning and mounting, which has a side effect of
not running udev rules so there are no entries for the freshly partitioned disk
under the children of `/dev/disk`.

Currently the workaround is to call partprobe in the recipe, which creates the
entries but we should do this as part of the image-partition action so that no
magic is needed in the recipe.

Fixes: f91e2b2c2789 ("Fix partition races while partitioning")
Resolves: #154

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -255,6 +255,16 @@ func (i ImagePartitionAction) getPartitionDevice(number int, context debos.Debos
 	}
 }
 
+func (i *ImagePartitionAction) triggerDeviceNodes(context *debos.DebosContext) error {
+	err := debos.Command{}.Run("udevadm", "udevadm", "trigger", "--settle", context.Image)
+	if err != nil {
+		log.Printf("Failed to trigger device nodes")
+		return err
+	}
+
+	return nil
+}
+
 func (i ImagePartitionAction) PreMachine(context *debos.DebosContext, m *fakemachine.Machine,
 	args *[]string) error {
 	image, err := m.CreateImage(i.ImageName, i.size)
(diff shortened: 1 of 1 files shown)


actions/image_partition_action.go
```diff
@@ -250,6 +250,20 @@ func (i ImagePartitionAction) getPartitionDevice(number int, context debos.Debos
...
 }
 
+func (i *ImagePartitionAction) triggerDeviceNodes(context *debos.DebosContext) error {
+	// find the raw /dev entry
+	device, _ := filepath.EvalSymlinks(context.Image)
+
+	// trigger udev rules
+	err := debos.Command{}.Run("udevadm", "udevadm", "trigger", device)
+	if err != nil {
+		log.Printf("Failed to reload partition table")
```
sjoerdsimons, line 260: we're not reloading the partiton table here; so that's an odd error

actions/image_partition_action.go
```diff
@@ -250,6 +250,20 @@ func (i ImagePartitionAction) getPartitionDevice(number int, context debos.Debos
 	}
 }
 
+func (i *ImagePartitionAction) triggerDeviceNodes(context *debos.DebosContext) error {
+	// find the raw /dev entry
+	device, _ := filepath.EvalSymlinks(context.Image)
+
+	// trigger udev rules
+	err := debos.Command{}.Run("udevadm", "udevadm", "trigger", device)
```
sjoerdsimons, line 258: This will only trigger e.g. `/dev/vda`; You probably want to use `--parent-match`  to trigger the raw disk and all its children; Also probably wise to add `--settle` so we know for sure udev had processed all events before exiting this function

PR #220, opened 2020-06-22
actions: apt-file: initial implementation of action to install local debian packages

obbardc:
The apt-file action allows local .deb packages to be installed using the `apt`
command, much like the apt action but for local packages rather than those from
apt repositories.

Resolves: #157
Closes: #165


actions/apt_file_action.go
```diff
@@ -0,0 +1,164 @@
...
+		aptOptions = append(aptOptions, "--no-install-recommends")
+	}
+
+	if apt.Unauthenticated {
+		aptOptions = append(aptOptions, "--allow-unauthenticated")
+	}
+
+	aptOptions = append(aptOptions, "install")
+	aptOptions = append(aptOptions, pkgs...)
+	err = c.Run("apt-file", aptOptions...)
```
sjoerdsimons, line 153: Note that apt will not install .debs you point to by file if they're of an older of the same version of what's already installed.. Which is likely not the behaviour the user expects ?

obbardc: apt doesn't allow that and exits with an error, i have added a note to the doc.

actions/apt_file_action.go
```diff
@@ -0,0 +1,164 @@
...
+	Recommends       bool
+	Unauthenticated  bool
+	Origin           string
+	Packages         []string
+}
+
+func (apt *AptFileAction) Run(context *debos.DebosContext) error {
+	apt.LogStart()
+	var origin string
+	aptOptions := []string{"apt", "-oDpkg::Progress-Fancy=0", "--yes"}
```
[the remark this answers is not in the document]
obbardc: using the apt command with progress-fancy broke debos output; switched back to apt-get to be more consistent

actions/apt_file_action.go
```diff
@@ -0,0 +1,164 @@
...
+	if !apt.Recommends {
+		aptOptions = append(aptOptions, "--no-install-recommends")
+	}
+
+	if apt.Unauthenticated {
+		aptOptions = append(aptOptions, "--allow-unauthenticated")
+	}
+
+	aptOptions = append(aptOptions, "install")
+	aptOptions = append(aptOptions, pkgs...)
```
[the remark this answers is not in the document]
obbardc: I added a separate commit to abstract the apt command away (also finally got around to this just over 3 years since the original implementation.....)

actions/install_dpkg_action.go
```diff
@@ -0,0 +1,180 @@
...
+				return err
+			}
+			if len(matches) == 0 {
+				return fmt.Errorf("file(s) not found after globbing: %s", pkg)
+			}
+
+			packages = append(packages, matches...)
+		}
+	} else {
+		packages = append(packages, origin)
```
sjoerdsimons, line 139: what if the user passed a Package list?

actions/install_dpkg_action.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	/* bind mount each package into rootfs & update the list with the
+	 * path relative to the chroot */
+	for idx, pkg := range packages {
+		// check for duplicates after globbing
+		for j := idx + 1; j < len(packages); j++ {
+			if packages[j] == pkg {
+				return fmt.Errorf("duplicate package found: %s", pkg)
```
sjoerdsimons, line 148: why not dedup the list rather then erroring out ?

actions/install_dpkg_action.go
```diff
@@ -0,0 +1,180 @@
...
+			}
+		}
+
+		log.Printf("Installing %s", pkg)
```
sjoerdsimons, line 152: you're printing Installing; but there is a bunch of stuff and commands being run before things are actually installed

sjoerdsimons (changes requested):

obbardc:
I rebased onto main. Still need to fix last review comments.

README.md
```diff
@@ -41,6 +41,7 @@ Some of the actions provided by debos to customize and produce images are:
 * download: download a single file from the internet
 * filesystem-deploy: deploy a root filesystem to an image previously created
 * image-partition: create an image file, make partitions and format them
+* install-dpkg: install packages and their dependencies from local 'deb' files
```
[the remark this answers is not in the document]
obbardc: done

obbardc:
@sjoerdsimons @daissi I cleaned up the commits. Can you take a quick look in case there is anything else? I plan to address remaining bits and get this in over the next month.

actions/install_deb_action.go
```diff
@@ -0,0 +1,181 @@
+/*
+InstallDeb Action
+
+Install packages from .deb files and their dependencies to the target rootfs
+using 'apt'.
+
+Dependencies will be satisfied first from the `packages` list (i.e. locally
+available packages) and then from the target's configured apt repositories. If
+`deps` is set to false, dependencies will not be installed and an error will be
```
obbardc: Current code doesn't seem to parse this. If it's not set, we'd need to add `--nodeps`. Needs more testing. Also, should it be set by default ?

wrapper/apt_wrapper.go
```diff
@@ -0,0 +1,49 @@
...
+func NewAptCommand(context debos.Context, label string) AptCommand {
+	command := "apt-get"
+
+	apt := AptCommand{
+		Wrapper: NewCommandWrapper(context, command, label),
+	}
+
+	apt.AddEnv("DEBIAN_FRONTEND=noninteractive")
+
+	/* Don't show progress update percentages */
```
obbardc, line 21: perhaps look at adding `-o=Dpkg::Progress-Fancy=0` ?

obbardc:
@sjoerdsimons can you take a look at the action yaml properties & description before i rework the PR ? Currently not sure about setting `no-deps=false` by default, maybe it needs to be changed.

actions/install_dpkg_action.go
obbardc: changed the log message to `Adding <package> to install list` rather than `Installing <package>`

actions/install_dpkg_action.go
obbardc: added dedup to list; leaving open as may not be right...

actions/install_dpkg_action.go
obbardc: will error out in that case now

tobsan:
This is exactly what I've been looking for, I can't believe I didn't check the PR list years ago!

Code looks good to me. Tested this locally and replaced all of my ugly `tmp_install.yaml` hacks with it. Works like a charm, at least for me!

actions/install_deb_action.go
```diff
@@ -0,0 +1,190 @@
...
+using 'apt'.
+
+Dependencies will be satisfied first from the `packages` list (i.e. locally
+available packages) and then from the target's configured apt repositories. If
+`no-deps` is set to true, dependencies will not be installed.
+
+Attempting to downgrade packages which are already installed is not allowed and
+will throw an error.
+
+ # Yaml syntax:
```
tobsan, line 13: The update property is never mentioned here.

obbardc: Done, thanks!

tests/install-deb/test.yaml
```diff
@@ -0,0 +1,24 @@
+architecture: "amd64"
```
sjoerdsimons: this test doesn't seem to be added to the CI tests; which seems a shame (and yeah it would require downloading some debs before execution)

actions/install_deb_action.go
```diff
@@ -0,0 +1,191 @@
...
+
+Optional properties:
+
+- origin -- reference to named origin. Defaults to `recipe` to install from local filesystem.
+
+- recommends -- boolean indicating if suggested packages will be installed. Defaults to false.
+
+- unauthenticated -- boolean indicating if unauthenticated packages can be installed. Defaults to false.
+
+- no-deps -- boolean indicating if dependencies are not resolved by apt. Defaults to false.
```
sjoerdsimons: if dependencies are not resolved; does it mean it just fails if it they're not satisfied? 

Fwiw in all the local deb install workarounds i've done; i've never ended up using nodeps,  so i think the default is fine.

 Bit on the fence on the naming though;  I don't think we do negative options anywhere else (apart from unauthenticated, but that feels different); why  not e.g. 'install-dependencies' with a default to true?

actions/install_deb_action.go
```diff
@@ -0,0 +1,191 @@
...
+
+- update -- boolean indicating if `apt update` will be ran before installing packages. Defaults to true.
```
sjoerdsimons, line 37: I guess this should default to true unless no-deps is true (which would make it pointless)

actions/install_deb_action.go
```diff
@@ -0,0 +1,191 @@
...
+e.g. the result of a download action, the package path will be automatically
+generated from the origin contents and the `packages` property can be omitted.
+
+Optional properties:
+
+- origin -- reference to named origin. Defaults to `recipe` to install from local filesystem.
+
+- recommends -- boolean indicating if suggested packages will be installed. Defaults to false.
+
+- unauthenticated -- boolean indicating if unauthenticated packages can be installed. Defaults to false.
```
sjoerdsimons: I assumed this means extra dependencies from unauthenticated repository can be installed? As packages by themselves aren't authenticated. Making the doc a bit more clear here might be good :)

actions/install_deb_action.go
```diff
@@ -0,0 +1,191 @@
...
+- packages -- list of package files to install from the filesystem (or named
+origin). Resolves Unix-style glob patterns. If installing from a named origin,
+e.g. the result of a download action, the package path will be automatically
+generated from the origin contents and the `packages` property can be omitted.
+
+Optional properties:
+
+- origin -- reference to named origin. Defaults to `recipe` to install from local filesystem.
+
+- recommends -- boolean indicating if suggested packages will be installed. Defaults to false.
```
sjoerdsimons:
```suggestion
- recommends -- boolean indicating if recommended packages will be installed. Defaults to false.
```

actions/install_deb_action.go
```diff
@@ -0,0 +1,191 @@
...
+   packages:
+     - package_path.deb
+     - *.deb
+
+Mandatory properties:
+
+- packages -- list of package files to install from the filesystem (or named
+origin). Resolves Unix-style glob patterns. If installing from a named origin,
+e.g. the result of a download action, the package path will be automatically
+generated from the origin contents and the `packages` property can be omitted.
```
sjoerdsimons: nitpick: So sometimes it's not a mandatory property :P  Also this seems to imply you can leave it whenever you use a origin, but from the code it seems that's only for the case where the origin is a single file rather then a dir

tests/install-deb/test.yaml
obbardc: oops, I will add that back in.

I guess for the test I could debootstrap stable then use e.g. `apt download hello` in a run action before calling this action.

sjoerdsimons: mmdebstrap!  but yeah that would be a neat way

obbardc: added the tests to CI && reworked it as I said in previous commit. pushed it so we can see if it works.

need to add 2 more tests, which are detailed in comments.

actions/install_deb_action.go
obbardc: yeah; i will change this to `install-dependencies` defaulting to true.

Need to check your question & also put the answer in the action description.

actions/install_deb_action.go
obbardc: agree; will do.

actions/install_deb_action.go
obbardc: agree. need to check and make docs more clear.

actions/install_deb_action.go
obbardc: need to double check and update docs. maybe also worth bailing out if its set and origin is from download action. will have a think.

tests/install-deb/test.yaml
obbardc: can't think of a neat way to do the other tests which are marked as TODO, would you be happy if i remove them or have any bright ideas ?

actions/install_deb_action.go
obbardc: done as separate commit && updated docs

actions/install_deb_action.go
obbardc: done in separate commit.

actions/install_deb_action.go
obbardc: done, I moved it to `Optional properties` section. Maybe it is worth just having a `Properties` section, but that may be confusing compared to other actions docs.

actions/install_deb_action.go
obbardc: HMMM apt doesn't have a `--nodeps` option. I will remove all mention of it....

tests/install-deb/test.yaml
```diff
@@ -0,0 +1,64 @@
...
+architecture: {{ $architecture }}
+
+actions:
+  - action: mmdebstrap
+    description: Bootstrap Debian {{ $suite }}
+    suite: {{ $suite }}
+    variant: minbase
+
+  # TODO: Test install-deb with packages from recipe dir
+  # TODO: This may be too hard to test; since I don't want any hard-coded URLs in the test recipe.
```
sjoerdsimons: We could use dget (from devscripts)  to download to the recipe dir, this would avoid hardcoded urls.

tests/install-deb/test.yaml
```diff
@@ -0,0 +1,64 @@
...
+    description: Test hello package
+    chroot: true
+    command: hello
+
+  # TODO: Test install-deb with packages from a named origin
+  # TODO: This may be too hard to test; since I don't want any hard-coded URLs in the test recipe.
+  # TODO: Remove package & clean apt cache after
+  #- action: download
+  #  description: Install Debian package from url
+  #  url: http://ftp.us.debian.org/debian/pool/main/b/bmap-tools/bmap-tools_3.5-2_all.deb
```
sjoerdsimons: We could use a package from snapshot.debian.org for this; While it's a hardcoded url, it's also one that shouldn't change.  Otherwise there isn't a way to test this indeed; Which i wouldn't entirely be against, that said it's also probably a more fragile/unused part of the code, so likely the most fragile

tests/install-deb/test.yaml
obbardc: done

tests/install-deb/test.yaml
obbardc: done!

actions/install_deb_action.go
```diff
@@ -0,0 +1,188 @@
+/*
+InstallDeb Action
+
+Install packages from .deb files and their dependencies to the target rootfs
+using 'apt'.
+
+Dependencies will be satisfied first from the `packages` list (i.e. locally
+available packages) and then from the target's configured apt repositories. If
+`install-dependencies` is set to true, dependencies will not be installed.
```
tobsan: This sounds backwards: we **don't** install dependencies when the property is true?

actions/install_deb_action.go
```diff
@@ -0,0 +1,188 @@
...
+property can be omitted.
+
+- origin -- reference to named origin. Defaults to `recipe` to install from local filesystem.
+
+- recommends -- boolean indicating if recommended packages will be installed. Defaults to false.
+
+- unauthenticated -- boolean indicating if unauthenticated packages (e.g. dependencies) can be installed by apt. Defaults to false.
+
+- update -- boolean indicating if `apt update` will be ran before installing packages. Defaults to true.
+
```
tobsan, line 38: The optional property `install-dependencies` is not documented here.

actions/install_deb_action.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/go-debos/debos/wrapper"
+)
+
+type InstallDebAction struct {
+	debos.BaseAction `yaml:",inline"`
+	Recommends       bool
+	Unauthenticated  bool
+	Update           bool
+	Origin           string
+	Packages         []string
```
tobsan, line 88: `InstallDependencies` is described in the docs but not included in the action struct.

actions/install_deb_action.go
obbardc: removed this documentation as it was leftover, thanks!

actions/install_deb_action.go
obbardc: removed this documentation as it was leftover, thanks!

actions/install_deb_action.go
obbardc: yep, rmeoved the docs

tests/install-deb/test.yaml
```diff
@@ -0,0 +1,101 @@
...
+    description: Install Debian package (recipe dir)
+    packages:
+      - {{ $package_name }}_*.deb
+
+  - action: run
+    description: Check installed Debian package version (recipe dir)
+    chroot: true
+    command: |
+      dpkg -l "{{ $package_name }}"
+      dpkg -l "{{ $package_name }}" | awk '$1=="ii"{print $3}' | grep -qx '{{ $package_version }}'
```
sjoerdsimons: fwiw `dpkg -s` is way simpler to parse

obbardc: done; should be better

PR #221, opened 2020-06-23
Wip/obbardc/action-image-partition-label-partlabel-type

obbardc:
Allow setting filesystem label, partition label and partition type from the recipe.

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -71,6 +71,14 @@ Optional properties:
 - partlabel -- label for the partition in the GPT partition table. Defaults
 to the `name` property of the partition. May only be used for GPT partitions.
 
+- parttype -- set the partition type in the partition table. The string should
+be in a hexadecimal format (2-characters) for msdos partition tables and GUID format
+(36-characters) for GPT partition tables. For instance, "82" for msdos sets the
+partition type to Linux Swap. Whereas "0657fd6d-a4ab-43c4-84e5-0933c84b4f4f" for
+GPT sets the partition type to Linux Swap.
+For msdos partition types hex codes see: https://en.wikipedia.org/wiki/Partition_type
+For gpt partition type GUIDs see: https://systemd.io/DISCOVERABLE_PARTITIONS/
+
 - features -- list of additional filesystem features which need to be enabled
 for partition.
 
@@ -155,6 +163,7 @@ type Partition struct {
 	number    int
 	Name      string
 	PartLabel string
+	PartType  string
 	Start     string
 	End       string
 	FS        string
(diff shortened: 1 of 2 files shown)


actions/image_partition_action.go
```diff
@@ -75,6 +75,14 @@ labels (i.e. `raw`)
 - partlabel -- label for the partition in the GPT partition table. Defaults
 to the `name` property of the partition. May only be used for GPT partitions.
 
+- parttype -- set the partition type in the partition table. The string should
+be in a hexadecimal format (2-characters) for msdos partition tables and GUID format
+(36-characters) for GPT partition tables. For instance, "82" for msdos sets the
+partition type to Linux Swap. Whereas "0657fd6d-a4ab-43c4-84e5-0933c84b4f4f" for
+GPT sets the partition type to Linux Swap.
+For msdos partition types hex codes see: https://en.wikipedia.org/wiki/Partition_type
+For gpt partition type GUIDs see: https://systemd.io/DISCOVERABLE_PARTITIONS/
```
sjoerdsimons, line 84: not a blocker for this PR; but would be nice to add a nicer way for the user to do this rather then putting in magic numbers

actions/image_partition_action.go
```diff
@@ -436,6 +445,14 @@ func (i ImagePartitionAction) Run(context *debos.DebosContext) error {
 			}
 		}
 
+		if p.PartType != "" {
+			err = debos.Command{}.Run("sfdisk", "sfdisk", "--part-type", context.Image, fmt.Sprintf("%d", p.number), p.PartType)
```
sjoerdsimons, line 449: can we do this as part of the parted call rather then needing another tools ?

actions/image_partition_action.go
```diff
@@ -151,6 +155,7 @@ import (
 type Partition struct {
 	number   int
 	Name     string
+	Label    string
```
sjoerdsimons: I don't get what this adds; What does it do you cannot currently do?

actions/image_partition_action.go
obbardc: yeah, the plan is to expose the GUIDs to the recipes in a follow-up MR

actions/image_partition_action.go
obbardc: currently the name parameter is checked for uniqueness; adding an optional label parameter would allow the partitions to have the same label or a different label to the name.

for instance,
```
- name: rootfs-A
  label: rootfs
  partlabel: rootfs

- name: rootfs-B
   label: rootfs
   partlabel: rootfs
```

Can drop the patch if you don't see the benefit to it

actions/image_partition_action.go
obbardc: doesn't look like parted can set raw guids

obbardc: unfortunately some more peeking into parted, only handles setting basic guids: https://git.savannah.gnu.org/gitweb/?p=parted.git;a=blob;f=libparted/labels/gpt.c;h=93f7add2a2411f035f6c4991506871c3e1e25663;hb=HEAD#l1458

actions/image_partition_action.go
obbardc: dropped patch

PR #224, opened 2020-08-04
actions: image_partition: add F2FS support

a-wai:
F2FS requires a few tweaks to be usable within debos:
  * its command-line option for setting the fs label is different from
    other mkfs utils
  * parted doesn't recognize "f2fs" as a valid filesystem, therefore it
    shouldn't appear on the command line

With these simple changes, it becomes possible to use F2FS partitions
with debos.


obbardc:
LGTM

PR #226, opened 2020-08-13
Build Debos and run tests for each merge request and push

obbardc:
Fix the test suite and run the complete test suite in GitHub actions.

For now only supporting running on the host (more information in commit messages).


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,45 @@
+name: Debos Continuous Integration
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    name: Build Debos
+    runs-on: ubuntu-latest
```
sjoerdsimons, line 8: can we run on debian buster instead?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+      - name: Install Go compiler
+        uses: actions/setup-go@v2
```
sjoerdsimons, line 11: Using debiasn go compiler

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,45 @@
...
+        uses: actions/checkout@v2
+
+      - name: Build debos on host
+        env:
+          GOSRC: /home/runner/go/src
+          GOBIN: /home/runner/go/bin
+        run: |
+          mkdir -p ${GOSRC}/github.com/go-debos
+          ln -s $(pwd) ${GOSRC}/github.com/go-debos/debos
+          go get -v ./...
```
sjoerdsimons, line 29: A lot of this duplicates what `docker/Dockerfile` already does; Which i guess is fine but it will cause duplication going forward ; Could we do this in a smarter way ?

docker/unit-tests.test.yml
```diff
@@ -1,10 +1,10 @@
...
 
 services:
-    sut:
-        build:
-            context: ..
-            dockerfile: docker/Dockerfile
-            target: builder
-        working_dir: /usr/local/go/src/github.com/go-debos/debos/actions
-        command: go test
+  debos-unit-tests:
```
sjoerdsimons, line 4: Will this rename still trigger testing inside dockerhub?

.github/workflows/ci.yaml
obbardc: unfortunately not: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#github-hosted-runners

sjoerdsimons: hmmm in that case and given we're already running the docker container build in a (seperate?) job that builds on buster can we use this job to build and test with the latest go instead?

Then we have two actually different test when it comes to compilations and unit tests etc at least

obbardc: running these tests allow you to get annotations which show against the source code, like: https://github.com/obbardc/debos/commit/4e04945f42c2623c3a9c783b92aaf2f5ef2ac975

unfortunately github actions isn't clever enough to create annotations from the docker run: https://github.com/obbardc/debos/commit/79291ca1d8f6ce0f0288aa8ae84bb9ae3ab6c062

if you don;t think that annotations are useful then we can just drop this whole stage and just run the docker tests

docker/unit-tests.test.yml
obbardc: no, this breaks it. whoops. fixed now

.github/workflows/ci.yaml
obbardc: yep, agree i will make the build run on latest go.

.github/workflows/ci.yaml
obbardc: removed the duplication !

obbardc:
Example test runs:
- https://github.com/obbardc/debos/runs/3010837287 - all OK
- https://github.com/obbardc/debos/actions/runs/1008159202 - compile error in Debos code
- https://github.com/obbardc/debos/actions/runs/1008591484 - failed tests

sjoerdsimons:
Silly question; this PR add all the test but they're not run a part of this PR? Is that something that needs to be enable in the repository still?

obbardc:
> Silly question; this PR add all the test but they're not run a part of this PR? Is that something that needs to be enable in the repository still?

you may need to enable the actions under "Settings" > "Actions" > "Allow local actions only" which only allows actions which are part of this repo to be ran.

they're current ran in the context of my personal repo since the trigger was from a "push" to my repo.

If there was an action defined for this repository for the "pull_request" action, it would be running as part of this pull request.
Once merged, future pulls will be run using the actions defined in this repo, so nothing yet needs to be changed (hopefully?).

sjoerdsimons:
hmm it's set to "allow all action" and "require approval for first time contributors"; i'll branch this into this repo to see what happens a bit more i guess :)

Issue #18, opened 2017-09-27
Expose the --include option of deboostrap in the configuration


em-:
When deboostrapping HTTPS-based repositories the `apt-transport-https` and `ca-certificates` should be installed right from the start, and that's usually handled with the `--include` option.

In my specific case their availability is probably better ensured through a custom debootstrap script (see #16), but I wonder if there's value in exposing more deboostrap options.

sjoerdsimons:
`debootstrap (1.0.59) unstable; urgency=medium

  * Install ca-certificates as well as apt-transport-https for HTTPS
    installations.  This makes it possible to copy certificates that were
    built into the installer to /usr/local/share/ca-certificates/ and thus
    have them continue to be trusted after installation.
`

So in debian since wheezy and ubuntu since trusty :)

Fwiw I'm wary about given direct access to --include, debootstraps dependency resolver is.. lets say basic. So it's quite easy to break things if using include, the preferred way is to debootstrap and then use apt actions for everything else.

In the exceptional case (like the https transport) where you really need to install extra things during debootstrap, I think debos needs to handle that internally.. The only other example i can think of is the keyring package, but that's a bit of a chicken and egg issue that we can hopefully flush out as part of #21.

PR #227, opened 2020-08-24
actions/debootstrap_action: Add property 'include'

sangorrin:
'include' is an optional property which contains a list of packages
which will be added to download and extract lists.
This uses option '--include' of debootstrap to install extra packages.

Example:
  include: [gnupg2, sudo]
or
  include:
    - gnupg2
    - sudo

Signed-off-by: Trung Do <trung1.dothanh@toshiba.co.jp>
Signed-off-by: Daniel Sangorrin <daniel.sangorrin@toshiba.co.jp>

### DIFF
--- a/actions/debootstrap_action.go
+++ b/actions/debootstrap_action.go
@@ -15,6 +15,7 @@ Yaml syntax:
    variant: "name"
    keyring-package:
    keyring-file:
+   include: <list of packages>
 
 Mandatory properties:
 
@@ -40,6 +41,15 @@ Example:
 - keyring-file -- keyring file for repository validation.
 
 - merged-usr -- use merged '/usr' filesystem, true by default.
+
+- include -- list of packages which will be added to download and extract lists.
+
+Example:
+ include: [gnupg2, sudo]
+ or
+ include:
+   - gnupg2
+   - sudo
 */
 package actions
 
@@ -63,6 +73,7 @@ type DebootstrapAction struct {
 	Components       []string
 	MergedUsr        bool `yaml:"merged-usr"`
 	CheckGpg         bool `yaml:"check-gpg"`
+	Include          []string
 }
 
 func NewDebootstrapAction() *DebootstrapAction {
@@ -141,6 +152,11 @@ func (d *DebootstrapAction) Run(context *debos.DebosContext) error {
 
 	if d.Variant != "" {
 		cmdline = append(cmdline, fmt.Sprintf("--variant=%s", d.Variant))
+	}
+
+	if d.Include != nil {
+		s := strings.Join(d.Include, ",")
+		cmdline = append(cmdline, fmt.Sprintf("--include=%s", s))
 	}
 
 	cmdline = append(cmdline, d.Suite)


sangorrin:
Since I noticed that this functionality was mentioned in https://github.com/go-debos/debos/issues/18, I would like to stress the use case for this. For some reason, the `debootstrap` action with a keyring file created with gnupg2 works fine (I imagine that the host has gnupg2 support), but trying to install anything after that through the `apt` action will fail. The reason is that gnupg2 does not seem to be included in the packages installed by `debootstrap` (at least in Buster), and since the `apt` action does an update by default (see https://github.com/go-debos/debos/pull/228) it complains about the repository release file  (signed with gnupg2). The workaround is installing `gnupg2` through the `run` action (without `apt update`) but it is not ideal.

sjoerdsimons:
@sangorrin Can you clarify your issue a bit more; Do you have  a custom keyring that's causing this issue? I would indeed not expect a basic buster bootstrap to have a gnupg2; it will have gpgv though

sangorrin:
Hi @sjoerdsimons 

We have a Debian repository using the Aptly server with GNUPG2 support, I didn't know about gpgv.
https://github.com/urpylka/docker-aptly/commit/e4cfe4b5a212c1963bb880d99b7af4e55b06f55a

After the debootstrap action we are using this as a workaround:

```
actions:
 - action: run
   description: "Installing gnupg2 for apt-key"
   chroot: true
   command: apt-get install -y gnupg2

 - action: overlay
   description: "Adding aptly public key"
   source: ../overlays/aptly-pub-key

 - action: run
   description: "Setting up aptly key with apt-key"
   chroot: true
   command: apt-key add /usr/share/keyrings/aptly_public_key.gpg
```

Once we have that in place the apt action works fine.

The workaround works fine for the moment, so you may prefer to wait until there are more users having the same problem.

Kind regards,
Daniel

sjoerdsimons:
@sangorrin What's the format of your aptly_public_key.gpg ? (file will tell you). As long as it's a key public ring format it should work fine. If it's just a keyblock you can convert it by runing something like:

```
gpg --homedir $(pwd) --no-options --no-default-keyring --keyring $(pwd)/keyring --import aptly_public_key.gpg
gpg --homedir $(pwd) --no-options --no-default-keyring --keyring $(pwd)/keyring --export >  aptly_public_keyring.gpg
```

If you pass that to the debootstrap action as the keyring things should work fine (this is the same format as used by the debian keys).

For reference also see how we solved this in apertis' keyring package: https://gitlab.apertis.org/pkg/apertis-archive-keyring

sangorrin:
Hi @sjoerdsimons, thanks a lot for your answer!
I checked my .gpg key with `file` and the result was:
```
$ file aptly_public_key.gpg
aptly_public_key.gpg: PGP/GPG key public ring (v4) created [...] RSA (Encrypt or Sign) 3072 bits MPI=[...].
```

I tried your instructions but unfortunately, the resulting `aptly_public_keyring.gpg` was exactly the same as the original.
I will prepare a small git repository that reproduces the problem.

sjoerdsimons:
> Hi @sjoerdsimons, thanks a lot for your answer!
> I checked my .gpg key with `file` and the result was:
> 
> ```
> $ file aptly_public_key.gpg
> aptly_public_key.gpg: PGP/GPG key public ring (v4) created [...] RSA (Encrypt or Sign) 3072 bits MPI=[...].
> ```
> 
> I tried your instructions but unfortunately, the resulting `aptly_public_keyring.gpg` was exactly the same as the original.
> I will prepare a small git repository that reproduces the problem.

Did you ever get around to this?

sangorrin:
Hi @sjoerdsimons, we have a workaround but let me prepare a reproducer now that our patches to aptly docker made it into the upstream repository.

sangorrin:
Hi @sjoerdsimons 

I realized that apt-key is now deprecated and the better of adding a key is now using
```
deb [signed-by=/usr/share/keyrings/deriv-archive-keyring.gpg] https://deriv.example.net/debian/ stable main
```

I have confirmed that this works fine, so I will close the issue.

Thanks,
Daniel

PR #229, opened 2020-09-07
WIP: action/debootstrap: add certificate and private-key

d4s:
Allow to use certificate and private key for client authentication.
Those action parameters falls thru `debootstrap` for `wget` allowing
to authenticate client on server side.


obbardc (approved):
LGTM!

PR #230, opened 2020-09-08
action/debootstrap: use absolute paths for certificate and private-key

d4s:
Need to use absolute paths for certificate and private key to avoid
problems caused by incorrect current directory.

### DIFF
--- a/actions/debootstrap_action.go
+++ b/actions/debootstrap_action.go
@@ -134,11 +134,11 @@ func (d *DebootstrapAction) Run(context *debos.DebosContext) error {
 	}
 
 	if d.Certificate != "" {
-		cmdline = append(cmdline, fmt.Sprintf("--certificate=%s", d.Certificate))
+		cmdline = append(cmdline, fmt.Sprintf("--certificate=%s", path.Join(context.RecipeDir, d.Certificate)))
 	}
 
 	if d.PrivateKey != "" {
-		cmdline = append(cmdline, fmt.Sprintf("--private-key=%s", d.PrivateKey))
+		cmdline = append(cmdline, fmt.Sprintf("--private-key=%s", path.Join(context.RecipeDir, d.PrivateKey)))
 	}
 
 	if d.Components != nil {


actions/debootstrap_action.go
```diff
@@ -134,11 +134,11 @@ func (d *DebootstrapAction) Run(context *debos.DebosContext) error {
 	}
 
 	if d.Certificate != "" {
-		cmdline = append(cmdline, fmt.Sprintf("--certificate=%s", d.Certificate))
+		cmdline = append(cmdline, fmt.Sprintf("--certificate=%s", path.Join(context.Artifactdir, d.Certificate)))
```
obbardc, line 137: worth checking file exists first in `Verify` ?

actions/debootstrap_action.go
```diff
@@ -134,11 +134,11 @@ func (d *DebootstrapAction) Run(context *debos.DebosContext) error {
...
 	}
 
 	if d.PrivateKey != "" {
-		cmdline = append(cmdline, fmt.Sprintf("--private-key=%s", d.PrivateKey))
+		cmdline = append(cmdline, fmt.Sprintf("--private-key=%s", path.Join(context.Artifactdir, d.PrivateKey)))
```
obbardc, line 141: as above

actions/debootstrap_action.go
obbardc: also, does this path need escaping or wrapping in quotes since it is an argument ?

obbardc (changes requested):

actions/debootstrap_action.go
d4s: mmm... the escaping is needed for the shell. Since we are using os/exec -- arguments are passed to debootsrap as is. And with blank spaces it doesn't work since debootstrap script do not expect blank spaces :-(
`PRIVATEKEY="--private-key=$2"` <- this is the argument passed to wget as is.

d4s:
Closing this MR since the https://github.com/go-debos/debos/pull/231 contains more useful approach

PR #231, opened 2020-09-08
action/debootstrap: allow to use files outside of RecipeDir

d4s:
Added stages "Verify" and "PreMachine" allowing to use secret files
outside of recipe directory tree (both absolute and relative paths
are supported).
Verification stage ensures that configuration files from options
are existing on file system.

This is needed for cases if secret files are not the part of Debos
recipe directory and useful for CI.

This MR completely overrides https://github.com/go-debos/debos/pull/230

### DIFF
--- a/actions/debootstrap_action.go
+++ b/actions/debootstrap_action.go
@@ -57,6 +57,7 @@ import (
 	"strings"
 
 	"github.com/go-debos/debos"
+	"github.com/go-debos/fakemachine"
 )
 
 type DebootstrapAction struct {
@@ -80,11 +81,55 @@ func NewDebootstrapAction() *DebootstrapAction {
 	// Be secure by default
 	d.CheckGpg = true
 	// Use main as default component
-	d.Components = []string {"main"}
+	d.Components = []string{"main"}
 	// Set generic default mirror
 	d.Mirror = "http://deb.debian.org/debian"
 
 	return &d
+}
+
+func (d *DebootstrapAction) listOptionFiles(context *debos.DebosContext) []string {
+	files := []string{}
+	if d.Certificate != "" {
+		d.Certificate = debos.CleanPathAt(d.Certificate, context.RecipeDir)
+		files = append(files, d.Certificate)
+	}
+
+	if d.PrivateKey != "" {
+		d.PrivateKey = debos.CleanPathAt(d.PrivateKey, context.RecipeDir)
+		files = append(files, d.PrivateKey)
+	}
+
+	if d.KeyringFile != "" {
+		d.KeyringFile = debos.CleanPathAt(d.KeyringFile, context.RecipeDir)
+		files = append(files, d.KeyringFile)
+	}
+
+	return files
+}
+
+func (d *DebootstrapAction) Verify(context *debos.DebosContext) error {
+	files := d.listOptionFiles(context)
+
+	// Check if all needed files exists
+	for _, f := range files {
+		if _, err := os.Stat(f); os.IsNotExist(err) {
+			return err
+		}
+	}
+	return nil
+}
+
+func (d *DebootstrapAction) PreMachine(context *debos.DebosContext, m *fakemachine.Machine, args *[]string) error {
+
+	mounts := d.listOptionFiles(context)
+
+	// Mount configuration files outside of recipes directory
+	for _, mount := range mounts {
+		m.AddVolume(path.Dir(mount))
+	}
+
+	return nil
 }
 
 func (d *DebootstrapAction) RunSecondStage(context debos.DebosContext) error {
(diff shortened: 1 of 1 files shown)


actions/debootstrap_action.go
```diff
@@ -80,13 +81,63 @@ func NewDebootstrapAction() *DebootstrapAction {
...
+	mounts := d.listOptionFiles(context)
+
+	// Mount configuration files outside of recipes directory
+	for _, mount := range mounts {
+		mount = path.Dir(mount)
+
+		// Check with "/" suffix to avoid false results for the directory
+		// which includes the RecipeDir path in the name, i.e. "/some/dir/RecipeDirExtended"
+		if !strings.HasPrefix(mount+"/", context.RecipeDir+"/") {
+			m.AddVolume(mount)
```
sjoerdsimons, line 134: why this special-case for the recipe dir?

d4s: Hm... I thought I described in comment.
Corner case -- let's assume we have `/some/dir/Recipe` with recipes and `/some/dir/RecipeAdditions` with certificates -- in that case `strings.HasPrefix` will return `true` and mount point will be skipped.
I fall into that during the test.

sjoerdsimons: well yes obviously; but why are you not just adding a volume for everything rather then special casing in avoiding the recipedir?

d4s: wanted to avoid the additional mounts like:
```
virtfs-5       811220760 695403072  74540088  91% /home/d4s/go/src/github.com/go-debos/debos
virtfs-7       811220760 695403072  74540088  91% /home/d4s/go/src/github.com/go-debos/debos/keyring
```
but you're right -- looks like unneeded optimisation

actions/debootstrap_action.go
```diff
@@ -104,7 +149,7 @@ func (d *DebootstrapAction) RunSecondStage(context debos.DebosContext) error {
 
 	err := c.Run("Debootstrap (stage 2)", cmdline...)
 
-	if (err != nil) {
+	if err != nil {
```
obbardc, line 152: no change here ?

actions/debootstrap_action.go
```diff
@@ -125,8 +170,7 @@ func (d *DebootstrapAction) Run(context *debos.DebosContext) error {
 	if !d.CheckGpg {
 		cmdline = append(cmdline, fmt.Sprintf("--no-check-gpg"))
 	} else if d.KeyringFile != "" {
-		path := debos.CleanPathAt(d.KeyringFile, context.RecipeDir)
-		cmdline = append(cmdline, fmt.Sprintf("--keyring=%s", path))
+		cmdline = append(cmdline, fmt.Sprintf("--keyring=%s", d.KeyringFile))
```
obbardc, line 173: what if spaces are in `d.KeyringFile` ?

obbardc: this also applies to the other arguments

obbardc (changes requested):
added some comments

also the commit message/PR name is not consistent with the change. Something like "action/debootstrap: allow using files outside of RecipeDir for keyring"

actions/debootstrap_action.go
```diff
@@ -80,13 +81,63 @@ func NewDebootstrapAction() *DebootstrapAction {
 	// Be secure by default
 	d.CheckGpg = true
 	// Use main as default component
-	d.Components = []string {"main"}
+	d.Components = []string{"main"}
```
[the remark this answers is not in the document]
d4s: right, automatic `gofmt` work on save.

actions/debootstrap_action.go
d4s: automatic gofmt work on save.

actions/debootstrap_action.go
d4s: I answered in [comment for #230](https://github.com/go-debos/debos/pull/230#discussion_r484981271): since we are using `os/exec` -- arguments are passed to debootsrap as is. No issues with Verify part.

for different parameters we have an interesting issues related to blank spaces:
- for keyring -- in case if any **directory in the path** to keyring contains blank spaces -- `fakemachine` hangs after mounting such volume, no issues if the keyfile itself have blank spaces.
- for certificate and private-key -- the same story with the path to file -- `fakemachine` hangs. And with blank spaces in file name -- `debootsrap` it doesn't work since it do not expect blank spaces. For example: `PRIVATEKEY="--private-key=$2"` <- this is the argument passed to wget as is and this breaks the work of `debootstrap` itself.

d4s: TBH I don't think that's the issue for the current MR -- but there are a couple of issues which should be fixed in `fakemachine` and `debootstrap`.

d4s: Right -- `fakemachine` stuck in `m.RunInMachineWithArgs()`

sjoerdsimons: Generally i wouldn't expect debos to be safe against paths with spaces (or with weird characters) especially since we indeed callout to debootstrap and other tools which can break as well; So for this MR i don't see it as a blocker; It might be nice at some point to have tests with spaces in various paths for various actions just to improve debos' correctness here

sjoerdsimons (approved):

PR #237, opened 2020-10-27
User friendly cmd and doc

Scott31393:
Hi,
I have added two new tools script to build and run Docker images from the root project directory. I think in this way user can jump and run the project on the flight like me :)
I have updated also the Readme with new documentation about these new features.
Let me know what do you think about that. :)

Regards.
Tommaso

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,18 +52,31 @@ https://godoc.org/github.com/go-debos/debos/actions
 
 Official debos container is available:
 ```
-docker pull godebos/debos
+$ docker pull godebos/debos
 ```
-
 See [docker/README.md](https://github.com/go-debos/debos/blob/master/docker/README.md) for usage.
 
 ## Installation (under Debian)
 
-    sudo apt install golang git libglib2.0-dev libostree-dev qemu-system-x86 \
+    $ sudo apt install golang git libglib2.0-dev libostree-dev qemu-system-x86 \
          qemu-user-static debootstrap systemd-container
-    export GOPATH=/opt/src/gocode # or whatever suites your needs
-    go get -u github.com/go-debos/debos/cmd/debos
+    $ export GOPATH=/opt/src/gocode # or whatever suites your needs
+    $ go get -u github.com/go-debos/debos/cmd/debos
     /opt/src/gocode/bin/debos --help
+
+## Build Docker images from source (Using Dockerfile)
+
+```
+$ git clone https://github.com/go-debos/debos.git
+cd debos
+$ docker build -f docker/Dockerfile -t godebos/debos
+```
+or just run:
+
+```
+$ ./build-docker.sh
+```
+
 
 ## Simple example
 
@@ -97,13 +110,38 @@ make a tarball.
 
 To run it, create a file named `example.yaml` and run:
 
-    debos example.yaml
+    $ debos example.yaml
 
 The final tarball will be named "debian.tgz" if you would like to modify
 this name, you can provided a different name for the variable image like
 this:
 
-    debos -t image:"debian-arm64.tgz" example.yaml
+    $ debos -t image:"debian-arm64.tgz" example.yaml
+
+## Run example (using Docker)
+
+```$ sudo chmod 777 /dev/kvm```
+
+```$ sudo docker run --rm --interactive --tty --device /dev/kvm --user $(id -u) --workdir /recipes --mount "type=bind,source=$(pwd),destination=/recipes" --security-opt label=disable <debos-docker-image-id> example.yaml ```
+
+or run:
+
+```
+$ ./run-docker.sh <docker-image-id> <file>.yaml
+```
+
+
+Eg:
+
+```
+$ sudo docker images
+
+REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
+godebos/debos       <none>              52ed84763b21        6 weeks ago         822MB
+
+
+$ ./run-docker.sh 52ed84763b21 docker/example.yaml
+```
 
 ## Other examples
 
--- a/build-docker.sh
+++ b/build-docker.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+docker build -f docker/Dockerfile -t godebos/debos .
--- a/docker/example.yaml
+++ b/docker/example.yaml
@@ -1,0 +1,23 @@
+{{- $image := or .image "debian.tgz" -}}
+
+architecture: arm64
+
+actions:
+  - action: debootstrap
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -52,19 +52,32 @@ https://godoc.org/github.com/go-debos/debos/actions
...
+
+```
+$ git clone https://github.com/go-debos/debos.git
+cd debos
+$ docker build -f docker/Dockerfile -t godebos/debos
+```
+or just run:
+
+```
+$ ./build-docker.sh
```
punitagrawal, line 77: TBH, I am not sure about the benefit of adding the container build script. The build arguments are simple enough.

Also, as structured the change gives the impression that the build will also fetch the source code. I would suggest to split the instructions into fetching the source code and building the container.

README.md
```diff
@@ -97,13 +110,38 @@ make a tarball.
...
 The final tarball will be named "debian.tgz" if you would like to modify
 this name, you can provided a different name for the variable image like
 this:
 
-    debos -t image:"debian-arm64.tgz" example.yaml
+    $ debos -t image:"debian-arm64.tgz" example.yaml
+
+## Run example (using Docker)
+
+```$ sudo chmod 777 /dev/kvm```
```
obbardc, line 123: for debian at least; you should add user into kvm group and reboot

README.md
```diff
@@ -97,13 +110,38 @@ make a tarball.
...
+
+```$ sudo docker run --rm --interactive --tty --device /dev/kvm --user $(id -u) --workdir /recipes --mount "type=bind,source=$(pwd),destination=/recipes" --security-opt label=disable <debos-docker-image-id> example.yaml ```
+
+or run:
```
obbardc, line 127: i don't think these scripts add much value tbh

obbardc (changes requested):

README.md
punitagrawal: Strictly speaking a reboot is not required - using `newgrp` is usually sufficient. Or if you have the luxury, logout / login should do the trick.

README.md
obbardc: (but adding the docker build command to the readme is still important)

README.md
Scott31393: Ok. I test debos fs on imx8 :)

[seco-c61]( https://www.seco.com/it/products/sbc-c61)

Except tty configurations all seems work :) I'm in.

The log:

```
U-Boot 2019.04-gb13fa9c-dirty (Oct 26 2020 - 17:51:30 +0100)

CPU:   Freescale i.MX8MMQ rev1.0 1600 MHz (running at 1200 MHz)
CPU:   Industrial temperature grade (-40C to 105C) at 48C
Reset cause: POR
Model: SECO i.MX8MM C61 board
DRAM:  4 GiB
DEBUG! board_init, 377 BOOT FROM ----> 7 
MMC:   KERN_ERR, DEBUG FUNC = fsl_esdhc_init, LINE = 1250, 
KERN_ERR, DEBUG FUNC = fsl_esdhc_init, LINE = 1250, 
FSL_SDHC: 1, FSL_SDHC: 2
Loading Environment from MMC... Run CMD11 1.8V switch
OK
System data from MMC1... OK.
In:    serial
Out:   serial
Err:   serial

 BuildInfo:
  - ATF a-20190
  - U-Boot 2019.04-gb13fa9c-dirty

flash target is MMC:1
Run CMD11 1.8V switch
Net:   
Warning: ethernet@30be0000 (eth0) using random MAC address - ee:6a:e7:4d:9e:67
eth0: ethernet@30be0000
Fastboot: Normal
Normal Boot
Hit any key to stop autoboot:  0 
u-boot seco c61=> 
u-boot seco c61=> boot
Run CMD11 1.8V switch
switch to partit Run CMD11 1.8V switch
==> Running in Normal Mode...
Running in Normal Mode...
Run CMD11 1.8V switch
switch to partit 24891904 bytes read in 317 ms (74.9 MiB/s)
Run CMD11 1.8V switch
switch to part 57682 bytes read in 13 ms (4.2 MiB/s)
Running in Normal Mode...
## Flattened Device Tree blob at 43000000
   Booting using the fdt blob at 0x43000000
   Loading Device Tree to 00000000fe90f000, end 00000000fe920151 ... OK

Starting kernel ...

[    0.000000] Booting Linux on physical CPU 0x0000000000 [0x410fd034]
[    0.000000] Linux version 4.19.35-g8edaeca-dirty (tom@tom-desktop) (gcc version 8.3.0 (GCC)) #2 SMP PREEMPT Tue Oct 27 11:35:17 CET 2020
[    0.000000] Machine model: FSL i.MX8MM C61 board


# apt
apt         apt-cdrom   apt-get     apt-mark    
apt-cache   apt-config  apt-key

# ls
bin   dev  home  lost+found  mnt  proc  run   srv  tmp  var
boot  etc  lib   media       opt  root  sbin  sys  usr

```

README.md
sjoerdsimons: for debian kvm is tagged with uaccess so it will be accessible to the logged in user so the kvm group isn't necessarily needed.. Generally documentation should not suggest chowning device files, it may make sense to have a seperate note on how to check/get access to /dev/kvm but it is distro/situation specific.

build-docker.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+docker build -f docker/Dockerfile -t godebos/debos .
```
sjoerdsimons, line 3: this seems to add little value

run-docker.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+docker run --rm --interactive --tty --device /dev/kvm --user $(id -u) --workdir /recipes --mount "type=bind,source=$(pwd),destination=/recipes" --security-opt label=disable $1 $2
```
sjoerdsimons, line 3: Please wrap that in multiple lines; and probably just hardcode usage of go-debos/debos as an image name.

 Though as @obbardc mentioned it's a bit of a questionmark who this is for as to get it you'd either need to download it directly or checkout the git repo.. But if you're doing that you may as well just copy&paste the command and tune it. 

I'm not really against it though; for us the only drawback would be the duplication and it potentially going out of sync

docker/example.yaml
```diff
@@ -0,0 +1,23 @@
+{{- $image := or .image "debian.tgz" -}}
+
```
sjoerdsimons, line 2: Wht's the rationale for moving this example here?

sjoerdsimons (changes requested):

PR #241, opened 2020-11-26
cmd: Add backend argument to choose fakemachine backend

obbardc:
Fakemachine now allows the virtualisation backend to be chosen when
creating the machine, so allow the backend to be chosen by the user.


cmd/debos/debos.go
```diff
@@ -144,13 +144,13 @@ func main() {
 		return
 	}
 
-	/* If fakemachine is supported the outer fake machine will never use the
+	/* If fakemachine is enabled the outer fake machine will never use the
```
sjoerdsimons, line 147: Unsure how to comment on a whole commit rather then change lines. 

But this commit doesn't seem to belong in this PR anyhow. And the commit message doesn't explain the rationale for the change either. So please seperate it and explain why the behaviour change.

obbardc: agree, i will move that commit to a different PR

cmd/debos/debos.go
```diff
@@ -94,8 +95,10 @@ func main() {
 	}()
 
 	parser := flags.NewParser(&options, flags.Default)
-	args, err := parser.Parse()
+	opt := parser.FindOptionByLongName("backend")
+	opt.Choices = fakemachine.BackendNames()
```
sjoerdsimons, line 99: I wonder if  fakemachine vs. non-fakemachine should be folded into the backend choice hwere:
auto means a support fakemachine setup *or* the non-fakemachine setup; backend kvm should fail if there is no fakemachine support at all. while there could be a non-fakemachine backend which is our old --disable-fakemachine essentially?

alternatively this option should be "fakemachine backend" specifically so it's more clear

cmd/debos/debos.go
```diff
@@ -214,7 +223,13 @@ func main() {
 	}
 
 	if !options.DisableFakeMachine && !fakemachine.InMachine() && fakemachine.Supported() {
```
sjoerdsimons, line 225: This means that if the user explicitly decided to use a fakemachine backend (e.g. kvm) debos will still fall back on the non-fakemachine case if it's not supported.

obbardc: thanks, i've sorted out that case, now debos will not fallback to the non-fakemachine case if a backend is explicitly set.

sjoerdsimons: Yes but if fakemachine is not supported at all it will *still* fallback :) as fakemachine.Supported() returns false

And tbh these conditions are getting out of hand :). You probably want to determine what to launch earier so the remainder of the flow is easier.. Fwiw the code is a bit tricky as it needs to match the following cases:

* outer debos
  - If backend is "auto", use fakemachine if supported otherwise fallback to non-fakemachine
 - if backend is set to something else force fakemachine usage and fail if it doesn't work

* inner debos:
 - don't ever start fakemachine and start the actions; (this is determed by the InMachine() check)

Fwiw as test case you probably should have:
- normal setup:
  * automatic usage of (a) fakemachine (backend)
  * automatic fallback to non-fakemachine 
- disabled fakemachine usage
 * automatic usage of non fakemachine (no backends option)
 * failure when specifying a backend

obbardc: OK i agree, it's caused me a lot of headscratching each time i read the code :-)

hopefully now i have cleaned that up a little bit...

PR #242, opened 2020-11-26
Wip/obbardc/examples fix

obbardc:
fix some documentation and the examples


docker/README.md
```diff
@@ -17,7 +16,14 @@ Check that `kvm` is enabled and writable by the user running the docker containe
 To build `recipe.yaml`:
 ```
 cd <PATH_TO_RECIPE_DIR>
-docker run --rm --interactive --tty --device /dev/kvm --user $(id -u) --workdir /recipes --mount "type=bind,source=$(pwd),destination=/recipes" --security-opt label=disable godebos/debos <RECIPE.yaml>
+docker run --rm --interactive --tty \
+  --device /dev/kvm \
+  --user $(id -u):$(id -g) \
```
sjoerdsimons, line 21: that's typically not needed

docker/README.md
```diff
@@ -17,7 +16,14 @@ Check that `kvm` is enabled and writable by the user running the docker containe
...
+  --group-add $(getent group kvm | cut -d: -f3) \
```
sjoerdsimons, line 22: kvm has uaccess tagged on modern systems so this is also not needed

README.md
```diff
@@ -103,8 +103,11 @@ this:
 
 ## Other examples
 
-This example builds a customized image for a Raspberry Pi 3.
-https://github.com/go-debos/debos-recipes
+The example in this repository under [doc/examples/example.yaml](doc/examples/example.yaml)
+builds a basic Debian tarball.
+
+A collection of user-submitted recipes are available under the [debos-recipes](https://github.com/go-debos/debos-recipes)
+repository.
```
sjoerdsimons, line 110: I'd just removed that as debos-recipes isn't actually actively maintained at the moment

sjoerdsimons (changes requested):
see notes :)

docker/README.md
obbardc: right, just having a look at this again since i don't fully understand the permissions here.
kvm has uaccess tagged on my system, and my user is part of the kvm group:

```
$ ls -la /dev/kvm
crw-rw----+ 1 root kvm 10, 232 Jan 28 09:36 /dev/kvm

$ getfacl /dev/kvm
getfacl: Removing leading '/' from absolute path names
# file: dev/kvm
# owner: root
# group: kvm
user::rw-
user:obbardc:rw-
group::rw-
mask::rw-
other::---
```

debos with kvm backend runs when i run it locally outside of docker, but not inside a container it breaks without this hunk of the patch:
```
$ docker run --rm --interactive --tty --device /dev/kvm --user $(id -u) --workdir /recipes --mount "type=bind,source=$(pwd),destination=/recipes" --security-opt label=disable godebos/debos:bullseye doc/examples/example.yaml
Could not access KVM kernel module: Permission denied
qemu-system-x86_64: failed to initialize kvm: Permission denied
error starting kvm backend: <nil>
```

i am probably doing something wrong here, or my system is configured incorrectly!

obbardc: hm seems like if i remove the `--user` arg from docker then it works fine, but then i loose the file permissions on the recipies bind mount

README.md
```diff
@@ -57,14 +57,10 @@ https://godoc.org/github.com/go-debos/debos/actions
 
 ## Installation (Docker container)
 
-Official debos container is available:
-```
-docker pull godebos/debos
-```
+A Docker container containing the latest version of Debos is available from [Docker Hub](https://hub.docker.com/r/godebos/debos).
```
obbardc, line 60: we should point to ghcr first

README.md
```diff
@@ -57,14 +57,10 @@ https://godoc.org/github.com/go-debos/debos/actions
...
+See [docker/README.md](https://github.com/go-debos/debos/blob/master/docker/README.md) for more information and usage.
```
obbardc, line 61: also the links in this file need to be modified

README.md
```diff
@@ -57,14 +57,10 @@ https://godoc.org/github.com/go-debos/debos/actions
...
 
-See [docker/README.md](https://github.com/go-debos/debos/blob/master/docker/README.md) for usage.
-
-## Installation (under Debian)
+## Installation from source (under Debian)
```
obbardc, line 63: add a section about installing from packages in debian

docker/README.md
```diff
@@ -17,7 +16,14 @@ Check that `kvm` is enabled and writable by the user running the docker containe
...
+  --workdir /recipes \
+  --mount "type=bind,source=$(pwd),destination=/recipes" \
+  --security-opt label=disable \
+  godebos/debos recipe.yaml
```
obbardc, line 26: use ghcr

obbardc:
Closing in favour of #468

PR #244, opened 2020-12-07
actions/pack_action: Add several compression options

tintou:
Match the same compression options as unpack and allow to have an uncompressed tar when removing the `compression` parameter.

Signed-off-by: Corentin Noël <corentin.noel@collabora.com>

### DIFF
--- a/actions/pack_action.go
+++ b/actions/pack_action.go
@@ -12,17 +12,30 @@ Mandatory properties:
 
 - file -- name of the output tarball, relative to the artifact directory.
 
-- compression -- compression type to use. Only 'gz' is supported at the moment.
+Optional properties:
+
+- compression -- compression type to use. Currently only 'gz', 'bzip2' and 'xz'
+compression types are supported. Use 'none' for uncompressed tarball. The 'gz'
+compression type will be used by default.
 
 */
 package actions
 
 import (
+	"fmt"
 	"log"
 	"path"
+	"strings"
 
 	"github.com/go-debos/debos"
 )
+
+var tarOpts = map[string]string{
+	"gz":    "z",
+	"bzip2": "j",
+	"xz":    "J",
+	"none":  "",
+}
 
 type PackAction struct {
 	debos.BaseAction `yaml:",inline"`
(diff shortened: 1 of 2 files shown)


actions/pack_action.go
```diff
@@ -34,8 +37,19 @@ func (pf *PackAction) Run(context *debos.DebosContext) error {
 	pf.LogStart()
 	outfile := path.Join(context.Artifactdir, pf.File)
 
+	var tarOpt = "cf"
```
obbardc, line 40: this could break recipes since you are changing the default parameter of `Compression` from `gz`=>`none`.
Maybe you should set the default value for `Compression` to `gz` and handle `none` under `tarOpts`

actions/pack_action.go
```diff
@@ -34,8 +37,19 @@ func (pf *PackAction) Run(context *debos.DebosContext) error {
...
+	tarOpts := map[string]string{
+		"gz":    "z",
+		"bzip2": "j",
+		"xz":    "J",
+	}
+
+	if tarOptCompression := tarOpts[pf.Compression]; len(tarOptCompression) > 0 {
```
obbardc, line 47: you are silently allowing user to choose parameters which may not exist. It makes sense to check the value in the `Verify` function of the action. that way the recipe will fail before its even ran.

obbardc (changes requested):

actions/pack_action.go
tintou: The `Compression` parameter was under the mandatory section so it's not a break from a documentation perspective :stuck_out_tongue:

obbardc: nevertheless, there are some recipes i've worked on which don't include the `compression` argument for this action and assume `tar.gz`

tintou: Okay, now defaulting to gz and added `none` parameter.

actions/pack_action.go
tintou: I've added a check in Verify for early failure.

actions/pack_action.go
```diff
@@ -30,12 +34,43 @@ type PackAction struct {
...
+
+	return nil
+}
+
 func (pf *PackAction) Run(context *debos.DebosContext) error {
 	pf.LogStart()
 	outfile := path.Join(context.Artifactdir, pf.File)
 
+	var tarOpt = "cf"
+	if len(pf.Compression) > 0 {
```
obbardc, line 59: makes it more readable if you set the default compression type rather than the tar options; also this probably belongs in a constructor function like https://github.com/go-debos/debos/blob/master/actions/debootstrap_action.go#L77

obbardc (changes requested):

obbardc:
also tests fail (this reminds me to finish the CI commenting on PRs :p):

```
[91m# github.com/go-debos/debos/actions
../../actions/pack_action.go:47:10: undefined: fmt
../../actions/pack_action.go:48:20: undefined: strings
```

tintou:
Oops, third time’s a charm

obbardc (approved):
LGTM, thanks!

actions/pack_action.go
```diff
@@ -30,12 +36,42 @@ type PackAction struct {
...
 func (pf *PackAction) Run(context *debos.DebosContext) error {
 	pf.LogStart()
 	outfile := path.Join(context.Artifactdir, pf.File)
 
+	var tarOpt = "cf"
+	tarOpts := map[string]string{
+		"gz":    "z",
+		"bzip2": "j",
+		"xz":    "J",
+		"none":  "",
```
sjoerdsimons, line 68: Can you make it so this is used inboth verify and Run. Rather then Verify having it's own array of valid compression options that needs to be manually kept in sync? Otherwise looks great thanks!

tintou: Okay, done :slightly_smiling_face:

PR #247, opened 2021-02-12
Add fakemachine uml backend support to Debos docker image

obbardc:


docker/Dockerfile
```diff
@@ -82,6 +84,7 @@ RUN apt-get update && \
         systemd-container \
         u-boot-tools \
         unzip \
+        user-mode-linux \
```
obbardc, line 87: including this package means the docker tests attempt to run in UML

obbardc (commented):
docker tests now fail since uml has higher priority than running on the host:

```
Successfully built 47522abb9b36
Successfully tagged bb9556mxkbpndvl9t52x78h_sut:latest
Creating network "bb9556mxkbpndvl9t52x78h_default" with the default driver
Creating bb9556mxkbpndvl9t52x78h_sut_1 ...
[1A[2K Creating bb9556mxkbpndvl9t52x78h_sut_1 ... [32mdone[0m [1B
/recipes/simple /recipes
ptrace: Operation not permitted
error starting uml backend: <nil>
check_ptrace : expected SIGSTOP, got status = 9
Removing bb9556mxkbpndvl9t52x78h_sut_1 ...
[1A[2K Removing bb9556mxkbpndvl9t52x78h_sut_1 ... [32mdone[0m [1BGoing to remove bb9556mxkbpndvl9t52x78h_sut_1
executing recipes.test.yml (255)
```

obbardc:
Fixed the docker hub tests failing.

docker/Dockerfile
```diff
@@ -38,7 +38,8 @@ ARG DEBIAN_FRONTEND
 ARG GOPATH
 
 # Set HOME to a writable directory in case something wants to cache things
-ENV HOME=/tmp
+ENV HOME=/scratch
+ENV TMP=/scratch
```
sjoerdsimons, line 42: /scratch is very specific to how the Collabora gitlab runners are setup; So they shouldn't be set as defaults. iotw please drop this chunk :)

obbardc: done; uml complains if $TMP isn't set (or the default) so I've set that to `/tmp`

PR #248, opened 2021-02-28
Add support for running on other architectures than amd64

mimi89999:
A better approach than in https://github.com/go-debos/debos/pull/91.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"runtime"
 )
 
 type ChrootEnterMethod int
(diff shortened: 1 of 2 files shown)


obbardc (changes requested):
Generally I think a lookup table would be nicer than the case statements.

mimi89999:
@obbardc I think that a lookup table would make the code much less readable. I prefer to do it with switches like it's done currently.

mimi89999:
@obbardc Do I need to use a lookup table for you to accept this PR or is it OK like that?

commands.go
```diff
@@ -209,9 +210,39 @@ func (cmd *Command) restoreResolvConf(sum *[sha256.Size]byte) error {
...
-	q.Setup()
-	defer q.Cleanup()
+	var emulated bool
+	switch cmd.Architecture {
+	case "":
+		emulated = false
+	case "amd64":
+		emulated = runtime.GOARCH != "amd64"
+	case "armhf", "armel", "arm":
+		emulated = runtime.GOARCH != "arm" || runtime.GOARCH != "arm64"
```
sjoerdsimons, line 220: Not all armv8 systems can run in 32 bit mode (aarch32);  So assuming you don't have to run emulated isn't generally true here

sjoerdsimons:
Having one central table to map from debian architectures to go architectures, qemu emulates  would make a lot of sense here to avoid having to change a few different places to add new achitectures

commands.go
mimi89999: Do you know what systems can't? How can I detect that?

sjoerdsimons: Some of the serverclass ARM chips can't; Cavium ThunderX is the main example i know of.

For detection there are some notes on: https://www.spinics.net/lists/arm-kernel/msg746790.html

sjoerdsimons:
superseded by newer commits

PR #264, opened 2021-06-15
actions: run: Expose scratch dir to scripts

obbardc:


sjoerdsimons:
Why is scratch dir being exposed? Please at least put the rationale in the commit message :)

sjoerdsimons:
@obbardc ping

obbardc:
was thinking about exposing this to be used as a temporary area, a la `/tmp`, for cases where `/tmp` isn't big enough

This can probably be closed as I don't have an actual use-case other than removing some hardcoding of `/scratch` in some recipes :-)

Issue #238, opened 2020-10-28
actions: debootstrap: require suite argument


obbardc:
With the following action, debos fails. We should check `suite` is present and contains a string. Having a default value here doesn't make a lot of sense.
```
architecture: arm64

actions:
  - action: debootstrap
```

```
2020/10/28 10:28:05 ==== debootstrap ====
2020/10/28 10:28:06 Debootstrap | W: Unable to read /etc/apt/apt.conf.d/ - DirectoryExists (2: No such file or directory)
2020/10/28 10:28:06 Debootstrap | I: usage: [OPTION]... <suite> <target> [<mirror> [<script>]]
2020/10/28 10:28:06 Debootstrap | I: Try `debootstrap --help' for more information.
2020/10/28 10:28:06 Debootstrap | E: You must specify a suite and a target.
2020/10/28 10:28:06 debootstrap.log | cat: /scratch/root/debootstrap/debootstrap.log: No such file or directory
2020/10/28 10:28:06 Action `debootstrap` failed at stage Run, error: exit status 1
```

PR #267, opened 2021-07-22
action/debootstrap: make suite property mandatory for debootstrap action

vigneshraman:
Check if suite property is present for debootstarp action and contains
a string. If suite is not present, exit with an error.

Fixes: #238

### DIFF
--- a/actions/debootstrap_action.go
+++ b/actions/debootstrap_action.go
@@ -110,6 +110,10 @@ func (d *DebootstrapAction) listOptionFiles(context *debos.DebosContext) []strin
 }
 
 func (d *DebootstrapAction) Verify(context *debos.DebosContext) error {
+	if len(d.Suite) == 0 {
+		return fmt.Errorf("suite property not specified")
+	}
+
 	files := d.listOptionFiles(context)
 
 	// Check if all needed files exists


obbardc:
hmm, may be better to set a default suite (`main` for debian) and hope debootstrap complains for other distros ;-)

vigneshraman:
> hmm, may be better to set a default suite (`main` for debian) and hope debootstrap complains for other distros ;-)

default mirror is http://deb.debian.org/debian
default component is main

So for suite can we use 'stable' as default ?

actions/debootstrap_action.go
```diff
@@ -161,6 +161,10 @@ func (d *DebootstrapAction) Run(context *debos.DebosContext) error {
 	d.LogStart()
 	cmdline := []string{"debootstrap"}
 
+	if len(d.Suite) == 0 {
```
obbardc, line 164: move to `Verify` function so that we can error out early

actions/debootstrap_action.go
```diff
@@ -161,6 +161,10 @@ func (d *DebootstrapAction) Run(context *debos.DebosContext) error {
...
+		return fmt.Errorf("'suite' is a mandatory property for Debootstrap Action")
```
obbardc, line 165:
```suggestion
		return fmt.Errorf("suite property not specified")
```

obbardc (changes requested):
two small nits

vigneshraman:
@obbardc, have updated the commit.

obbardc (approved):

Issue #234, opened 2020-10-17
actions: image-partition: should only attempt to mount a partition with a filesystem


obbardc:
```
architecture: arm64
actions:
  - action: image-partition
    imagename: test.img
    imagesize: 500MB
    partitiontype: gpt
    mountpoints:
      - mountpoint: /
        partition: rootfs
    partitions:
      - name: rootfs
        fs: none
        start: 10M
        end: 100%
```

results in:
```
2020/10/17 10:53:06 ==== image-partition ====
2020/10/17 10:53:06 Action `image-partition` failed at stage Run, error: rootfs mount failed: no such device
2020/10/17 10:53:06 Warning: Failed to get unmount /: invalid argument
2020/10/17 10:53:06 Unmount failure can cause images being incomplete!
Powering off.
```

At the verify stage we should check if the filesystem is actually mountable (IOW; if `fs` is not `none`) and print a useful error message if not.

PR #268, opened 2021-07-22
actions/image-partition: check if the filesystem is mountable

vigneshraman:
If the filesystem type is 'none', it results in the below error.
==== image-partition ====
Action `image-partition` failed at stage Run, error: root mount failed: no such device
Warning: Failed to get unmount /boot/efi: no such file or directory
Unmount failure can cause images being incomplete!
Powering off.

Report an error message if filesystem type is 'none' and exit.

Fixes: #234

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -784,6 +784,10 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 		if m.part == nil {
 			return fmt.Errorf("Couldn't find partition for %s", m.Mountpoint)
 		}
+
+		if strings.ToLower(m.part.FS) == "none" {
+			return fmt.Errorf("Cannot mount %s: filesystem not present", m.Mountpoint)
+		}
 	}
 
 	// Calculate the size based on the unit (binary or decimal)


actions/image_partition_action.go
```diff
@@ -651,6 +651,8 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 		switch p.FS {
 		case "fat32":
 			p.FS = "vfat"
+		case "none", "None", "NONE":
```
obbardc, line 654: none is a valid filesystem type, this is wrong.

what you should probably meant to do is check for mounting something with a none mountpoint. So you should move this to the next `i.Mountpoints` for loop and change this to be something like:

```
if m.FS.ToLower() == "none" {
    return fmt.Errorf("Cannot mount filesystem with type '%s'", p.FS)
}
```

obbardc (changes requested):

actions/image_partition_action.go
vigneshraman: @obbardc , Have updated the commit. Thanks.

actions/image_partition_action.go
```diff
@@ -668,6 +668,11 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 
 		for pidx, _ := range i.Partitions {
 			p := &i.Partitions[pidx]
+
+			if strings.ToLower(p.FS) == "none" {
```
obbardc, line 672: you didnt test this did you?

consider this action for a test:
```
 - action: image-partition
   imagename: "test.img"
   imagesize: 1GB
   partitiontype: gpt
   mountpoints:
     - mountpoint: /
       partition: root
     - mountpoint: /boot
       partition: boot
   partitions:
     - name: boot
       fs: fat32
       start: 0%
       end: 64MB
     - name: test
       fs: none
       start: 64MB
       end: 128MB
     - name: root
       fs: ext4
       start: 128MB
       end: 100%
```

from a quick look it will fail for every partition which has the `none` type - even those which aren't being mounted. so you need to check the partition type AFTER you have found the correct partition for this mountpoint.

so i think the partition type check should probably go after `if m.part == nil {`

obbardc (changes requested):

actions/image_partition_action.go
vigneshraman: > you didnt test this did you?

I did test this with yaml file in https://github.com/eballetbo/debos-recipes/blob/master/images/weston/debian-qemu-uefi.yaml and it exited with error when fs was none.

Let me check with the one you provided.

vigneshraman: Updated the error message to print partition name also. I checked with the above file and it exited if fs is none for each partition.

obbardc: Right, the recipe i posted about should not fail because `test` isn't ever mounted :-)

hence why we need to do the check for mountpoints and not partitions.

vigneshraman: Sorry, I missed the point before :)
Will make the changes and test it.

actions/image_partition_action.go
```diff
@@ -676,6 +676,10 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 		if m.part == nil {
 			return fmt.Errorf("Couldn't find partition for %s", m.Mountpoint)
 		}
+
+		if strings.ToLower(m.part.FS) == "none" {
+			return fmt.Errorf("Cannot mount filesystem with type '%s' for partition name '%s'", m.part.FS, m.part.Name)
```
obbardc, line 681:
```suggestion
			return fmt.Errorf("Cannot mount %s: filesystem not present", m.Mountpoint)
```

obbardc (changes requested):

actions/image_partition_action.go
vigneshraman: Updated the commit. Thanks.

obbardc (approved):

PR #269, opened 2021-07-23
actions/image-partition: check label length in verify stage

vigneshraman:
The partition name (label) can be up to 11 characters long for vfat
and 16 characters long for ext3/4.

When partition name length is longer than 11 characters for vfat,
it fails with below message,
==== image-partition ====
cmdline: [mkfs.vfat -F32 -n efi456789012 /dev/vda1]
Formatting partition 1 | mkfs.vfat: Label can be no longer than 11 characters
Formatting partition 1 | mkfs.fat 4.2 (2021-01-31)
Action `image-partition` failed at stage Run, error: exit status 1

When partition name length is longer than 16 characters for ext2/3/4,
it shows the below warning message,
cmdline: [mkfs.ext4 -L root567890123456789 /dev/vda2]
Formatting partition 2 | Warning: label too long; will be truncated to 'root567890123456'

In case the partition name is longer than 11 for vfat, name is truncated
to 11 characters and a warning message is reported. In case of ext3/4,
it is truncated to 16 characters.

Originally-by: Christopher Obbard <chris.obbard@collabora.com>

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -54,7 +54,8 @@ Mandatory properties:
 
 - name -- is used for referencing named partition for mount points
 configuration (below) and label the filesystem located on this partition. Must be
-unique.
+unique. The partition name can be up to 11 characters long for vfat and 16 characters
+long for ext2/3/4. Name should be same as partition name in mount points.
 
 - fs -- filesystem type used for formatting.
 
@@ -613,6 +614,27 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 			}
 		}
 
+		if i.PartitionType == "gpt" {
+			var maxLength int = 0
+			if p.PartLabel == "" {
+				p.PartLabel = p.Name
+			}
+
+			switch p.FS {
+				case "vfat":
+					maxLength = 11
+				case "ext2", "ext3", "ext4":
+					maxLength = 16
+			}
+
+			if len(p.Name) > maxLength {
+				truncated := p.Name[0:maxLength]
+				log.Printf("Warning: partition label for %s '%s' is too long; truncated to '%s'", p.Name, p.PartLabel, truncated)
+				p.PartLabel = truncated
+				p.Name = truncated
+			}
+		}
+
 		if len(p.FSUUID) > 0 {
 			if p.FS == "btrfs" || p.FS == "ext2" || p.FS == "ext3" || p.FS == "ext4" || p.FS == "xfs" {
 				_, err := uuid.Parse(p.FSUUID)
@@ -668,6 +690,7 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 
 		for pidx, _ := range i.Partitions {
 			p := &i.Partitions[pidx]
+
 			if m.Partition == p.Name {
 				m.part = p
 				break


obbardc:
rather than exiting, would it be better to trim the label and report a warning?
also, the docs will probably need to be changed.

vigneshraman:
> rather than exiting, would it be better to trim the label and report a warning?
@obbardc, yes it would be better to trim and report warning in case of vfat.  For ext3/4, the label will be truncated to 16 characters by mkfs.ext4 and a warning is reported. So we can ignore ext3/4 case ?

> also, the docs will probably need to be changed.
Sure, will check this. Thanks.

actions/image_partition_action.go
```diff
@@ -669,6 +670,13 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 		for pidx, _ := range i.Partitions {
 			p := &i.Partitions[pidx]
 			if m.Partition == p.Name {
+				if p.FS == "vfat" {
```
obbardc, line 673: you shouldn't check that here, you should check it after: `for idx, _ := range i.Partitions {`

you should probably also determine the label there, something like (untested):

```
for idx, _ := range i.Partitions {
		if p.Name == "" {
			return fmt.Errorf("Partition without a name")
		}

		// moved this code from the Run() function...
		if p.PartitionType == "gpt" {
			if p.PartLabel == "" {
				p.PartLabel = p.Name
			}

			var maxLength int = 0
			switch p.FS {
				case "vfat":
					maxLength = 11
				case "ext2":
					maxLength = 16
				case "ext3":
					maxLength = 16
			}
			if maxLength > 0 {
					truncated := p.PartLabel[0:maxLength]
					log.Printf("Warning: partition label for %s '%s' is too long; truncated to '%s'", p.Name, p.PartLabel, truncated)
					p.PartLabel = truncated
			}
		}

		// check for duplicate partition names
		for j := idx + 1; j < len(i.Partitions); j++ {
			if i.Partitions[j].Name == p.Name {
				return fmt.Errorf("Partition %s already exists", p.Name)
			}
		}
}
```

obbardc (changes requested):

actions/image_partition_action.go
vigneshraman: @obbardc, this was added here since we need to check both m.Partition and p.Name. Can we do the above check before `if m.Partition == p.Name` check ?

actions/image_partition_action.go
```diff
@@ -668,6 +690,7 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 
 		for pidx, _ := range i.Partitions {
 			p := &i.Partitions[pidx]
+
```
obbardc, line 693: spurious change

actions/image_partition_action.go
```diff
@@ -613,6 +614,27 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 			}
 		}
 
+		if i.PartitionType == "gpt" {
+			var maxLength int = 0
+			if p.PartLabel == "" {
+				p.PartLabel = p.Name
+			}
+
+			switch p.FS {
```
obbardc, line 623: also, can we have a look at the maximum lengths for the other filesystems ?

actions/image_partition_action.go
```diff
@@ -54,7 +54,8 @@ Mandatory properties:
 
 - name -- is used for referencing named partition for mount points
 configuration (below) and label the filesystem located on this partition. Must be
-unique.
+unique. The partition name can be up to 11 characters long for vfat and 16 characters
+long for ext2/3/4. Name should be same as partition name in mount points.
```
obbardc, line 58:
```suggestion
long for ext2/3/4. Any longer labels will be automatically truncated.
```

also; the length should probably be moved to the `partlabel` documentation

actions/image_partition_action.go
```diff
@@ -54,7 +54,8 @@ Mandatory properties:
 
 - name -- is used for referencing named partition for mount points
 configuration (below) and label the filesystem located on this partition. Must be
-unique.
+unique. The partition name can be up to 11 characters long for vfat and 16 characters
```
obbardc, line 57:
```suggestion
unique. The partition label can be up to 11 characters long for vfat and 16 characters
```

obbardc (changes requested):

obbardc (changes requested):
you can also remove duplicate code in `Run()`:

```
		if p.PartLabel == "" {
			p.PartLabel = p.Name
		}
```

actions/image_partition_action.go
vigneshraman: Sure. Will check for btrfs, xfs, reiserfs, jfs.

vigneshraman:
Closing this PR. New proposed changes are in https://github.com/go-debos/debos/pull/271

Issue #219, opened 2020-06-15
actions: debootstrap: debootstrap gives a warning about no such file or directory /etc/apt/apt.conf.d


obbardc:
The debootstrap action shows a misleading warning:

```
2020/06/15 07:24:05 Debootstrap | W: Unable to read /etc/apt/apt.conf.d/ - DirectoryExists (2: No such file or directory)
```

andhe:
FWIW I just verified a hack like this silences the warning by debootstrap action:

```diff --git a/actions/debootstrap_action.go b/actions/debootstrap_action.go
index 506270c..fb976db 100644
--- a/actions/debootstrap_action.go
+++ b/actions/debootstrap_action.go
@@ -159,6 +159,11 @@ func (d *DebootstrapAction) RunSecondStage(context debos.DebosContext) error {
 
 func (d *DebootstrapAction) Run(context *debos.DebosContext) error {
        d.LogStart()
+
+       /* HACK: apt warns if directory is missing, so create it. */
+       mkaptdircmd := []string{"mkdir", "-p", "/etc/apt/apt.conf.d"}
+       debos.Command{}.Run("Debootstrap (prep)", mkaptdircmd...);
+
        cmdline := []string{"debootstrap"}
 
        if d.MergedUsr {
```

I guess the main question is if maybe fakemachine should provide a sane apt configuration out of the box or if debos should work around it like this?

obbardc:
yeah, a hack like that was exactly what i was thinking.

perhaps it should remove the directory after running if its empty.

andhe:
Since this is inside fakemachine, I don't really see much need for cleaning up. However, the hack should possibly also be added to the apt action which I guess also generate this warning (but haven't looked), for those with recipes that doesn't first use the debootstrap action.

PR #270, opened 2021-07-23
debootstrap: Make sure /etc/apt/apt.conf.d exists

obbardc:
Currently running the debootstrap action inside a fakemachine results
in the following warning:

    Debootstrap | W: Unable to read /etc/apt/apt.conf.d/ - DirectoryExists (2: No such file or directory

Make sure the directory exists in the fakemachine host before running
the debootstrap command.

Fixes: #219

### DIFF
--- a/actions/debootstrap_action.go
+++ b/actions/debootstrap_action.go
@@ -208,6 +208,14 @@ func (d *DebootstrapAction) Run(context *debos.DebosContext) error {
 	cmdline = append(cmdline, d.Mirror)
 	cmdline = append(cmdline, "/usr/share/debootstrap/scripts/unstable")
 
+	/* Make sure /etc/apt/apt.conf.d exists inside the fakemachine otherwise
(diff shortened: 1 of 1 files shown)


actions/debootstrap_action.go
```diff
@@ -208,6 +208,11 @@ func (d *DebootstrapAction) Run(context *debos.DebosContext) error {
 	cmdline = append(cmdline, d.Mirror)
 	cmdline = append(cmdline, "/usr/share/debootstrap/scripts/unstable")
 
+	/* Make sure /etc/apt/apt.conf.d exists before running debootstrap */
+	if err := os.MkdirAll(path.Join("/etc/apt/apt.conf.d"), os.ModePerm); err != nil {
```
sjoerdsimons, line 212: If you're running outside of fakemachine this then potentially creates a directory on the host system which probably should be avoided :)

Issue #251, opened 2021-03-23
actions: image-partition: Should check label length in Verify stage


obbardc:
`mkfs.vfat` now fails if the partition label is >11 characters. Older versions of `mkfs.vfat` just silently accepted the label length. We should probably check the label for all partition types length in the action's `Verify()` function.

```
2021/03/23 10:54:41 ==== image-partition ====
2021/03/23 10:54:42 Formatting partition 1 | mkfs.vfat: Label can be no longer than 11 characters
2021/03/23 10:54:42 Formatting partition 1 | mkfs.fat 4.2 (2021-01-31)
2021/03/23 10:54:42 Action `image-partition` failed at stage Run, error: exit status 1
```

obbardc:
mkfs.fat -n truncates a name longer than 11 bytes, and
mkfs.ext[34] -L truncate a label to 16 bytes.

PR #271, opened 2021-07-27
actions/image-partition: truncate filesystem label to maximum supported length

vigneshraman:
Added optional fslabel property to partition to allow truncation of filesystem
label and allow user to modify filesystem label without modifying the name.
Filesystem label defaults to the name property of the partition.

The filesystem label can be up to 11 characters long for vfat, 16 characters
long for ext2/3/4, 255 characters long for btrfs, 512 characters long for
hfs/hfsplus and 12 characters long for xfs. Any longer labels will be
automatically truncated.

Fixes: #251

Suggested-by: Christopher Obbard <chris.obbard@collabora.com>

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -43,6 +43,7 @@ Yaml syntax for partitions:
      - name: partition name
 	   partlabel: partition label
 	   fs: filesystem
+	   fslabel: filesystem label
 	   start: offset
 	   end: offset
 	   features: list of filesystem features
@@ -71,6 +72,12 @@ Optional properties:
 
 - partlabel -- label for the partition in the GPT partition table. Defaults
 to the `name` property of the partition. May only be used for GPT partitions.
+
+- fslabel -- (optional) Sets the volume name (label) of the filesystem. Defaults
+to the `name` property of the partition. The filesystem label can be up to 11
+characters long for vfat, 16 characters long for ext2/3/4, 255 characters long
+for btrfs, 512 characters long for hfs/hfsplus and 12 characters long for xfs.
+Any longer labels will be automatically truncated.
 
 - parttype -- set the partition type in the partition table. The string should
 be in a hexadecimal format (2-characters) for msdos partition tables and GUID format
@@ -168,6 +175,7 @@ type Partition struct {
 	number    int
 	Name      string
 	PartLabel string
+	FSLabel   string
 	PartType  string
 	Start     string
 	End       string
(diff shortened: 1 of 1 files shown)


actions/image_partition_action.go
```diff
@@ -613,6 +620,12 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 			}
 		}
 
+		if i.PartitionType == "gpt" {
```
obbardc, line 624: remove outer if statement

obbardc (changes requested):
looking at it properly; we may as well squash these two comments

also add rationale to adding seperate `fslabel` property in commit message. Something like "Added optional `fslabel` property to `partition` to allow truncation of filesystem label and allow user to modify filesystem label without modifying the name."

actions/image_partition_action.go
vigneshraman: Squashed the commits and updated the commit. Thanks.

actions/image_partition_action.go
```diff
@@ -72,6 +73,12 @@ Optional properties:
 - partlabel -- label for the partition in the GPT partition table. Defaults
 to the `name` property of the partition. May only be used for GPT partitions.
 
+- fslabel -- Sets the volume name (label) of the filesystem. Defaults to the
```
obbardc, line 76:
```suggestion
- fslabel -- (optional) Sets the volume name (label) of the filesystem. Defaults to the
```

obbardc (changes requested):
also change commit and PR title to something more descriptive; like `actions/image-partition: truncate filesystem label to max supported length`

actions/image_partition_action.go
vigneshraman: @obbardc, updated the commit

actions/image_partition_action.go
```diff
@@ -72,6 +73,12 @@ Optional properties:
 - partlabel -- label for the partition in the GPT partition table. Defaults
 to the `name` property of the partition. May only be used for GPT partitions.
 
+- fslabel -- (optional) Sets the volume name (label) of the filesystem. Defaults
```
obbardc, line 76:
```suggestion
- fslabel -- label for the filesystem. Defaults
```

obbardc (changes requested):
Couple of nits.

actions/image_partition_action.go
```diff
@@ -654,6 +663,31 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
...
+			case "hfs", "hfsplus":
+				maxLength = 255
+			case "xfs":
+				maxLength = 12
+		}
+
+		if maxLength > 0 && len(p.FSLabel) > maxLength {
+			truncated := p.FSLabel[0:maxLength]
+			log.Printf("Warning: fs label for %s '%s' is too long; truncated to '%s'", p.Name, p.FSLabel, truncated)
+			p.FSLabel = truncated
```
sjoerdsimons, line 689: Why truncate rather then erroring out?

sjoerdsimons:
fwiw this should really just check the name length in verify to error out early if the name is too long; the seperate fslabel doesn't really have a point, if you want to differentiate the partitation label and filesystem labet there is already partlabel for that..  And there really is no point in setting the name to different values other then using it in one of the labels

sjoerdsimons (changes requested):
fail don't truncate as noted in my lsat comment ;)

sjoerdsimons:
fixed by #531

PR #272, opened 2021-07-28
tests: add tests for "image-partition" action

vigneshraman:
Add recipe file to test "image-partition" action.


obbardc:
An alternative implementation has been merged into main with https://github.com/go-debos/debos/blob/main/tests/partitioning/test.yaml

PR #275, opened 2021-08-06
New Github workflow which publishes images on push to master

eds-collabora:
This replaces the old, simpler pipeline with a three phase process:

- First, build the image and cache it using docker buildx
- Second, run all the tests in parallel, restoring the image from the cache.
- Third, if this is a push to master, and the tests pass, push to DockerHub and GitHub containers.

This uses Buildkit caching aggressively, and will make use of the entire 5GiB allocation of cache space that GitHub provides over time.

It requires the following additional repository secrets:

- DOCKERHUB_USERNAME: the username to login as on DockerHub (e.g. go-debos)
- DOCKERHUB_PASSWORD: an access token for the DockerHub repository.


.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,198 @@
...
+        env:
+          IMAGE: ${{ env.TEST_TAG }}
+
+  publish:
+    name: Publish
+    needs:
+      - unit-test
+      - recipe-test
+      - uml-test
+    if: github.event_name != 'pull_request'
```
obbardc, line 152: what's the event which gets triggered when a pull request is merged ooi ?

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,198 @@
...
+
+      - name: Run test recipes using UML backend
+        run: |
+          docker-compose -f docker/recipes-test-uml.yml \
+            up --exit-code-from=sut
+        env:
+          IMAGE: ${{ env.TEST_TAG }}
+
+  publish:
+    name: Publish
```
obbardc, line 147: is there a way to make sure this doesn't get triggered in downstream forks ?

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,198 @@
...
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
```
obbardc, line 178: we will need to add `DOCKERHUB_USERNAME` and `DOCKERHUB_PASSWORD` as secrets and make sure they cannot be accessed by other pipeline triggerers :-)

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,198 @@
...
+        with:
+          path: /tmp/.build-cache
+          key: ${{ runner.os }}-docker-${{ github.sha }}
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
```
obbardc, line 195: since it says tags; i guess it pushes to an array of tags ?

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,198 @@
...
+    tags: [ 'v*.*.*' ]
+  pull_request:
+    branches: [ master ]
+  workflow_dispatch:
+    ~
+
+env:
+  DOCKERHUB_TAG: godebos/debos
+  GITHUB_TAG: ghcr.io/${{ github.repository_owner }}/debos
+  TEST_TAG: ${{ github.repository }}
```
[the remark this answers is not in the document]
obbardc: hmm, looks like you want to make this unique per pipeline run

obbardc (changes requested):

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,198 @@
+name: Build, test and deploy
```
obbardc, line 1: have you thought about having these as separate jobs rather than one single job ?

.github/workflows/docker-publish.yml
eds-collabora: It's "push" in that case.

.github/workflows/docker-publish.yml
eds-collabora: The meta step builds up the tags here; we pass it two bases `ghcr.io/go-debos/debos` and `godebos/debos`. It then appends something useful - ":master" or e.g. ":v1.0.1" to both. Those then become the two targets for the image.

obbardc: it should append `:latest`, not `:master` ?

.github/workflows/docker-publish.yml
eds-collabora: I'm not sure I understand here. In this repo, it will only trigger on a push to master (or manually from the UI - debatable whether that should be turned off or not). In a forked repo, they won't possess any of the authentication tokens for this repo or the dockerhub repo, so this step will fail if they run it without updating their own credentials [more precisely, as written it will perfectly happily push to their own container repo, but they will need to disable the DockerHub login step until they provide it with the correct credentials for their own repo].

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,198 @@
...
+    branches: [ master ]
+    tags: [ 'v*.*.*' ]
+  pull_request:
+    branches: [ master ]
+  workflow_dispatch:
+    ~
+
+env:
+  DOCKERHUB_TAG: godebos/debos
+  GITHUB_TAG: ghcr.io/${{ github.repository_owner }}/debos
```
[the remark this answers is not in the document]
eds-collabora: I like this form because a downstream fork can use it to push their own images to their own packages on GitHub. We can't make that automatically possible for DockerHub, but I don't see why we want to prevent forks from getting this benefit within Github? What's the downside for us here?

.github/workflows/docker-publish.yml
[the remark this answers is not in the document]
eds-collabora: This was just taste I think; I don't ever push to this tag, it's just a sane local default that's easy to read in the logs. It's only used in the following workflow:

- Reconstruct image from local buildx cache
- Install image on local docker instance under this name
- Trigger tests on that installed name, using docker-compose

It circumvents the really awkward relationship between docker-compose and docker build buildx.

.github/workflows/docker-publish.yml
eds-collabora: I can look into that. It's the default for the Github setup (which I think is master here). Will investigate.

.github/workflows/docker-publish.yml
obbardc: well, we don't want the step to fail in downstream forks ? :-)

.github/workflows/docker-publish.yml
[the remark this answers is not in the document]
obbardc: does it automatically create the container registry for the fork, or are manual steps required to do that for a fresh fork ?

.github/workflows/docker-publish.yml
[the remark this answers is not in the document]
obbardc: right; i am worried about passing the `IMAGE` to the unit tests etc since docker hub hardcodes that if you run tests there

.github/workflows/docker-publish.yml
eds-collabora: The idea with a single workflow was to test and then deploy a fixed target. The graph view of the jobs that the workflow contains that Github produces is quite nice and descriptive too, and captures the parallel parts well. The build process isn't completely safe here - there are clear interactions with docker caching because of syncing latest packages, and building from vcs - so keeping the parts together seemed advantageous.

.github/workflows/docker-publish.yml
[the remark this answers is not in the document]
eds-collabora: As I understand it, the registry always exists. I didn't do any setup for my fork.

.github/workflows/docker-publish.yml
eds-collabora: Ah I see. I'll investigate.

.github/workflows/docker-publish.yml
eds-collabora: The default is to tag pushes to master just as :master, and to only add :latest when a version is tagged:

> ### Latest tag
> latest tag is handled through the flavor input. It will be generated by default (auto mode) for:
>  -  type=ref,event=tag
>  - type=semver,pattern=...
>  - type=match,pattern=...

The default table of examples (which we're using here) looks like it has semver setup, giving:

> Event | Ref | Docker Tags
> -- | -- | --
> push | refs/heads/master | master
> push tag | refs/tags/v1.2.3 | v1.2.3, latest
> push tag | refs/tags/v2.0.8-beta.67 | v2.0.8-beta.67, latest

I could change it to tag master as latest as well, if that's desired. On the face of it, only tagging releases feels like quite a sensible default, but happy to change this @obbardc?

.github/workflows/docker-publish.yml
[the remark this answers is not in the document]
eds-collabora: I think you're right; I didn't understand how the DockerHub testing worked, but it had consequences just for local testing too. I've switched to using "debos" and "debos-builder" as obviously local tags, and removed the environment variable entirely.

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,198 @@
+name: Build, test and deploy
+
+on:
+  push:
+    branches: [ master ]
```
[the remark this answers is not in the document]
eds-collabora: > can we use a variable for main branch; in case it changes ?

Do you have a strategy in mind for accomplishing this? I tried a few things and I can't make this work:

- environment variables seem to be silently ignored at the top level (at least in this context) - my builds never ran
- YAML anchors and aliases aren't supported (ref [here](https://github.community/t/support-for-yaml-anchors/16128))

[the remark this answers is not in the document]
obbardc: how about we just simplify this to:

```
on: [ push, pull_request ]
```

.github/workflows/docker-publish.yml
obbardc: I think we should rename the file `build-test.yaml` as that is what the pipeline should do, the docker publishing is just a nice-to-have ;-)

.github/workflows/docker-publish.yml
obbardc: thinking should we not just push a container in all cases?

.github/workflows/docker-publish.yml
obbardc: Hm, we don't actually do many releases. Currently in Dockerhub the `latest` tag points to the last build `HEAD`.

.github/workflows/docker-publish.yml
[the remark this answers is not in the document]
obbardc: nice! In that case, can we replace `debos` with `${{ github.repository }}` in case forks don't use the same name ?

.github/workflows/docker-publish.yml
eds-collabora: That's probably a better idea than what I ended up doing. I'll make that change.

.github/workflows/docker-publish.yml
eds-collabora: Perhaps we should do a nightly scheduled build of master as a separate workflow? I'd be happy to do that if that sounds useful? It avoids some of the intricacies of keeping `:latest` up to date when it builds from different triggers.

obbardc: I think we just need to make sure that `latest` tag points to the last container built from `master` ;-)
Wasting actions time as a scheduled run is... meh

eds-collabora: I've done this, but the only way to avoid hard-coding master (given the lack of variable support, and the general state of conditionals) was to simply distinguish any push. I think this is safe in this repo, because there's only one branch. It isn't particularly good in forks though, that might have more branches, but I'm not sure how to resolve this. I could filter every step with the test against the environment, but that will lead to all four publish actions running, which is harder to follow (presently, two of the four run for any given workflow run, with the two that aren't appropriate marked as disabled).

.github/workflows/docker-publish.yml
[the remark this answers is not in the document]
eds-collabora: I've left the option of manual triggering in, but this is otherwise resolved. The justification for manual triggering is in the event of an unexpected build failure on master, which is possible given the update and checkout actions contained in the workflow.

PR #276, opened 2021-08-09
Add a pacstrap action

eds-collabora:


eds-collabora:
Checklist of comments on this code to date:
- [ ] [Don't provide the root dir](https://github.com/santoshmahto7/debos/commit/f57f91c3f9936308506f3c4aa66ec92141459b06#r53535960)
- [ ] [Remove hard-coded values](https://github.com/santoshmahto7/debos/commit/f57f91c3f9936308506f3c4aa66ec92141459b06#r53535903)
- [ ] [Don't set up the pacman key before pacstrap](https://github.com/santoshmahto7/debos/commit/f57f91c3f9936308506f3c4aa66ec92141459b06#r53536249)
- [ ] [Do not invoke pacstrap with -G](https://github.com/santoshmahto7/debos/commit/f57f91c3f9936308506f3c4aa66ec92141459b06#r53543231)
- [ ] [Drop mirror layout patch (Manjaro support](https://github.com/santoshmahto7/debos/commit/fe2c4bfe0b78713baba85ef373e5c6d0774a155b#commitcomment-53537458)
- [ ] [Drop user defined repositories patch](https://github.com/santoshmahto7/debos/commit/ba40558e45f406302b08f663dbd956cb8a12bac1#commitcomment-53537706)
- [ ] [Keep siglevel option patch (BlackArch support)](https://github.com/santoshmahto7/debos/commit/81052fb5011417d31ee0426f1d6b63d9316ee8ea#commitcomment-53538999)
- [ ] [Drop timeout patch](https://github.com/santoshmahto7/debos/commit/2e004a2cd0417c9f18c8af0cc1e5ef9321e466a2#commitcomment-53539052)

actions/pacstrap_action.go
```diff
@@ -0,0 +1,126 @@
...
+[core]
+Server = %[2]s/$repo/os/$arch
+
+[extra]
+Server = %[2]s/$repo/os/$arch
+
+[community]
+Server = %[2]s/$repo/os/$arch
+`
+
```
eds-collabora, line 49: Don't provide the root dir (original comment [here](https://github.com/santoshmahto7/debos/commit/f57f91c3f9936308506f3c4aa66ec92141459b06#r53535960))

actions/pacstrap_action.go
```diff
@@ -0,0 +1,126 @@
...
+type PacstrapAction struct {
+	debos.BaseAction `yaml:",inline"`
+	Mirror string
+}
+
+func NewPacstrapAction() *PacstrapAction {
+	d := PacstrapAction{}
+	// Set generic default mirror
+	d.Mirror = "http://mirrors.kernel.org/archlinux"
+
```
eds-collabora: No hardcoded values (original comment [here](https://github.com/santoshmahto7/debos/commit/f57f91c3f9936308506f3c4aa66ec92141459b06#r53535903))

actions/pacstrap_action.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	// Run pacman-key
+	cmdline := []string{"pacman-key", "--nocolor", "--config", configPath, "--init"}
+	err = debos.Command{}.Run("Pacman-key", cmdline...)
+	if err != nil {
+		return fmt.Errorf("Couldn't init pacman keyring: %v", err)
+	}
+
+	cmdline = []string{"pacman-key", "--nocolor", "--config", configPath, "--populate", "archlinux"}
```
eds-collabora: Don't need to call pacman-key before pacstrap here (original comment [here](https://github.com/santoshmahto7/debos/commit/f57f91c3f9936308506f3c4aa66ec92141459b06#r53536249))

actions/pacstrap_action.go
```diff
@@ -0,0 +1,126 @@
...
+	err = debos.Command{}.Run("Pacman-key", cmdline...)
+	if err != nil {
+		return fmt.Errorf("Couldn't populate pacman keyring: %v", err)
+	}
+
+	// Run pacstrap
+	cmdline = []string{"pacstrap", "-GM", "-C", configPath, context.Rootdir}
```
eds-collabora, line 109: Drop -G from the call to pacstrap (original comment [here](https://github.com/santoshmahto7/debos/commit/f57f91c3f9936308506f3c4aa66ec92141459b06#r53543231))

actions/pacstrap_action.go
```diff
@@ -81,29 +81,13 @@ func (d *PacstrapAction) Run(context *debos.DebosContext) error {
...
-	if err != nil {
-		return fmt.Errorf("Couldn't create var/lib/pacman in image: %v", err)
-	}
-	err = os.MkdirAll(path.Join(context.Rootdir, "etc", "pacman.d", "gnupg"), 0755)
-	if err != nil {
-		return fmt.Errorf("Couldn't create etc/pacman.d/gnupg in image: %v", err)
-	}
-
 	// Run pacman-key
 	cmdline := []string{"pacman-key", "--nocolor", "--config", configPath, "--init"}
```
eds-collabora, line 127: There's a conflict here between re-using the host's pacman configuration and making it work in fakemachine (without root privileges). If you don't have root privileges, you can't read some of the GPG configuration for the host, that means you need to make an unprivileged GPG configuration. It's probably possible (but really messy I think?) to detect whether you have the necessary privileges, but I got around that in a subsequent commit by making the public keyring available even when creating a new pacman configuration.

actions/pacstrap_action.go
eds-collabora: There's a conflict here between re-using the host's pacman configuration and making it work in fakemachine (without root privileges). If you don't have root privileges, you can't read some of the GPG configuration for the host, that means you need to make an unprivileged GPG configuration. It's probably possible (but really messy I think?) to detect whether you have the necessary privileges, but I got around that in a subsequent commit by making the public keyring available even when creating a new pacman configuration.

evelikov: Right forgot that the pacman/gnupg secrets are root-only. Let's add an inline comment vaguely alike:

```
The original pacman/gnupg secrets are root-only and we want to avoid running fakemachine/debos as root.
As such, explicitly run pacman-key --init so that new set is generated. 
```
Fwiw there's was some discussion upstream to make those another dedicated user, although that has stalled.

actions/pacstrap_action.go
```diff
@@ -0,0 +1,155 @@
...
+
+Mandatory properties:
+
+ - mirror -- the full url for the repository, with placeholders for
+   $arch and $repo as needed, as would be found in mirrorlist
+
+Optional properties:
+ - repositories -- list of repositories to use for packages selection.
+   Properties for repositories are described below.
+
```
eds-collabora, line 19: I'm hoping this new config data works better in terms of handling:
- [Avoiding the mirror layout patch](https://github.com/santoshmahto7/debos/commit/fe2c4bfe0b78713baba85ef373e5c6d0774a155b#commitcomment-53537458)
- [Keeping the siglevel patch (or at least functionality)](https://github.com/santoshmahto7/debos/commit/81052fb5011417d31ee0426f1d6b63d9316ee8ea#commitcomment-53538999)
- [Dropping the user defined repositories patch](https://github.com/santoshmahto7/debos/commit/ba40558e45f406302b08f663dbd956cb8a12bac1#commitcomment-53537706); although you can still select your repositories, the default is correct, and there's no possibility of mixing and matching between mirrors. I left the option to select them in because even arch has more repositories than this at need, and I couldn't see an obvious cost to doing so.

actions/pacstrap_action.go
eds-collabora: We do have to provide the GPG dir though, because it needs to be writable on some code paths.

actions/pacstrap_action.go
evelikov: Fully agree - the elaborate mirror-layout is gone. Having default repositories with override sounds reasonable - say you want a normal x86-64, or you want to add multilib on top. Or even switch to the testing version of said repos.

evelikov (approved):

actions/pacstrap_action.go
```diff
@@ -71,6 +72,15 @@ func NewPacstrapAction() *PacstrapAction {
 	return &d
 }
 
+func (d *PacstrapAction) PreMachine(context *debos.DebosContext, m *fakemachine.Machine, args *[]string) error {
```
evelikov, line 72: Nit: add a inline comment - "Use host keyring to speed up resolving public keys" or alike

actions/pacstrap_action.go
```diff
@@ -109,6 +119,17 @@ func (d *PacstrapAction) Run(context *debos.DebosContext) error {
 		return fmt.Errorf("Couldn't init pacman keyring: %v", err)
 	}
 
+	// Kickstart the public keyring if we can, to avoid expensive
+	// key retrieval
+	if info, err := os.Stat("/etc/pacman.d/gnupg/pubring.gpg"); err == nil && !info.IsDir() {
```
evelikov, line 134: nit `something := /etc/pacman.d/gnupg/pubring.gpg" then then use it throughout

actions/pacstrap_action.go
```diff
@@ -81,41 +81,54 @@ func (d *PacstrapAction) PreMachine(context *debos.DebosContext, m *fakemachine.
...
 	if err != nil {
 		return fmt.Errorf("Couldn't open pacman config: %v", err)
 	}
+	defer func() {
+		f.Close()
+		if err != nil {
+			os.Remove(configPath)
+		}
+	}()
+
```
evelikov, line 92: Nit: let's make this cleanup (nicely spotted btw) a separate patch. It's a bugfix, while the rest is cosmetics/style.

evelikov (changes requested):

actions/pacstrap_action.go
evelikov: Agreed.

actions/pacstrap_action.go
```diff
@@ -42,24 +46,31 @@ SigLevel = Required DatabaseOptional TrustAll
 
 const configRepoSection = `
 
-[%s]
-Server = %s
+[%[1]s]
+Server = %[2]s
```
evelikov, line 48: Not sure if I mentioned this before - with a single server/mirror we can drop the explicit "Server =" for each repository and use `Include = /etc/pacman.d/mirrorlist` instead. That means that we'll need to (re)create it and write the server URL in there.

actions/pacstrap_action.go
```diff
@@ -0,0 +1,155 @@
...
+ - repositories -- list of repositories to use for packages selection.
+   Properties for repositories are described below.
+
+Yaml syntax for repositories:
+
+ repositories:
+   - name: repository name
+     siglevel: signature checking settings (optional)
+*/
+package actions
```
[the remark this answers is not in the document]
evelikov: Makes sense, I'm in favour.

Pacman has CI building against Debian and Fedora amongst others. So theoretically it's possible for people to a) install pacman package on Fedora and b) kick off debos.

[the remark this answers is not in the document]
obbardc: yes, this is the next question, how we integrate that.
I was thinking we would have another Dockerfile which created an Arch image, but open to other suggestions (let's open another issue so we don't bikeshed here ;-) )

actions/pacstrap_action.go
evelikov: On Arch, the keychain will be checked early on (don't recall before or after the package downloads) and keys will be updated/revoked and discarded.

If Debian does not that, then it makes perfect sense to pull it independently, otherwise I'm not sure what's the benefit.

actions/pacstrap_action.go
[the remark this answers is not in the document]
evelikov: Fwiw I'm inclined that we can leave the Verify as follow-up PR. Two reasons come to mind:
 - the core changes needed - aka "Add the ability to check a command for existence"
 - no other action checks for it's respective commands in the Verify hook ... it seems

actions/pacstrap_action.go
```diff
@@ -78,6 +78,16 @@ func NewPacstrapAction() *PacstrapAction {
 	return &d
 }
 
+func (d *PacstrapAction) Verify(context *debos.DebosContext) error {
+	if err := (debos.Command{}.CheckExists("Pacstrap", "pacstrap", "-h")); err != nil {
+		return err
+	}
+	if err := (debos.Command{}.CheckExists("Pacman-key", "pacman-key", "-V")); err != nil {
```
evelikov, line 85: Having a helper which takes struct { label string; argv[] string } array should be cleaner IMHO. Not a big deal either way

actions/pacstrap_action.go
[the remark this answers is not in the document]
obbardc: yes i agree with folow-up; we should probably check the presence of all commands an action uses before running the recipe ;-).

@eds can you open an issue?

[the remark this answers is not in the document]
eds-collabora: Opened #280 to track this.

actions/pacstrap_action.go
eds-collabora: I think this is the main outstanding issue on this PR. Is there anything I can do to help resolve this?

My understanding is:
- We use TrustAll instead of TrustedOnly in pacman.conf, so that the keys for individual packages only have to be present, not trusted specifically.
- Copying the host's public keyring over avoids pacman having to obtain and verify a large number of keys during the action's run. We are re-using the work the host did in obtaining these keys originally.
- The patch set originally contained a (very slow) key population step, which has been removed. This downloaded and installed the keys afresh for all arch packagers (but specifically, just arch). This could allow TrustedOnly to be set for arch based rootfses.

In my mind the questions are:
- Do we think TrustAll is adequate here, or do we need to move to establishing specifically trusted keys?
- Is it acceptable to re-use the host's keying, when present?

obbardc: Personally I think we should use at little any configuration/keys from the host. But in arch-specific case (which i know very little about) this opinion may be the wrong opinion to have ;-)

evelikov: Why we should be (re)using as little * from the host - there's an implicit (and now explicit) statement yet there's no clear details why. Is there an assumption that "the host cannot be trusted", "they might have it wrong" or perhaps something else?

obbardc: differences between environments - using the fakemachine is supposed to create a somewhat known abstraction layer over whatever horrors the host machine has configured ;-)

evelikov: I think we can (should really) be using the Arch defaults for SigLevel. Which means dropping the TrustAll in the above hunk - thanks Ed, I didn't spot it previously. No reason comes to mind why we should diverge - but if there's a compelling one, we can. Obviously there should be a good inline comment explaining the decision.

As you know already - there is per-repo SigLevel override can be provided for anyone has third party repos. Say they have mandatory database signing, etc.

eds-collabora: I thought about this quite a bit. When I started testing with pacman as well, I realised the limitations of the pacstrap action as written. Very little of what it does actually persists into the rootfs (with the except of the mirror at least, following the latest patch set). I'm also fairly well persuaded by the idea of isolation from the host, where possible. At the very least, it makes things a lot more predictable, which can be important in a toolchain with many users.

The reworked patch set shows what it takes to achieve isolation. I could probably split the biggest commit into pieces with some thought. So in this new commit, the TrustedOnly default is restored, and trust levels (and siglevels) propagate into the rootfs. It also gives tight control over which keys and trust anchors are included in the rootfs.

The big disadvantage of this patch is I ended up needing a new tool to make it work. Specifically: I needed a bind mount to make pacman-key in any way isolated from the host, and the existing Command framework doesn't permit that without a chroot; a chroot I can't run in because this is a prerequisite for making it. So I made a "chmnt" command, source [here](https://gitlab.collabora.com/eds/chmnt/-/tree/master/cmd/chmnt), which lets you run a process with an altered filesystem and environment without chroot.

I'm not sure this is the change set we want, but I wanted to offer it up for discussion. I'll try to separate out the host isolation patch from the pacman sanity patch to make the scope clearer.

evelikov: Fwiw I like the idea behind `chmnt` and I think it makes sense regardless of the pacstrap/pacman actions. The extra patches push the complexity a bit and I won't have time to properly look until 30th Aug.

evelikov: Thinking about this a bit more and experimenting some, I'm concerned about growing debos codebase and maintainability. Especially since Arch moves fast - much unlike debos.

The idea behind `chmnt` is quite nice, yet I'm contemplating on the likelihood (and time-frame) of it being merged into debos.

So here are the individual steps I'd suggest:
 - let's bite the bullet and keep the `pacman-key --init/populate` within the `pacstrap` action
   - document akin to: "we use pacman-key locally for a) isolation and b) using the host data requires root access and running debos as root is no-go"
   - keep the original `mkdir(/var/lib/pacman)` and `mkdir(/etc/pacman.d/gnupg)`
   - nit: please drop the `--nocolor` argument
- drop the `/scratch` prefix from `pacman.conf` dirs
   - with the `mkdir(/etc/pacman.d/gnupg)` above, `pacman-key` will pull the keys within the debos instance itself
   - the removed `-G` (`pacstrap -GM...`) will ensure the files are copied properly into `/scratch`
- use `/etc/pacman.conf` instead of `/scratch/etc/pacman.conf`
   - former should be the bootstrap while the latter the actual user conf (part of the pacman package)
   - we can drop the custom `--config` and `-C` from `pacman-key` and `pacstrap` invocations, respectively
- stop generating `pacman.conf`, have a user-specified one - location/filename can be set in the yaml
   - figuring the optimal layout, might be an uphill battle
   - upstream Arch can add/change things, so a simple cp/paste is much clearer and scale-able than fixing/redeploying debos
- drop the `-M` from `pacstrap -M` - it prevents the mirrorlist (now generated) from being copied into `/scratch`

In brief/overview the code will look like:
 - read_yaml(pacman.conf location)
 - mount (add volume?) the file into /etc/pacman.conf
 - read_yaml(mirror url or filelocation)
 - create (or mount) mirrorlist
 - mkdir as needed for pacman-key
 - pacman-key --init; pacman-key --populate archlinux << full invocation
 - pacstrap /scratch << full invocation

evelikov: Note: The outline above should be the least amount of code, for debos maintainers to go through. Although if they like the work so-far (in part or in full) feel free to discard it.

eds-collabora: Although this MR as it stands is more complex than some of the alternatives,  it also attempts to address all the points raised in the (fairly extended) discussion to date. 

I'm happy to make any or all of the changes you suggest, but I think ideally at this stage, so that things don't become circular, it'd be great to get some feedback on the maintainers on where they see this patch set needing to go.

obbardc: emils advice looks good to me

eds-collabora: I had a go at this. I made the debos stanzas be simply `ConfigFile` plus optional `MirrorFile`.  There are a few problems:

- We can't mount `/etc/pacman.conf`; we can't directly change `/etc` in debos, we can mount subdirectories of `/etc`, but not the directory itself. This isn't crippling, but the `-C` arguments are here to stay.
-  I believe `pacman-key` runs on `GPGDir` from the provided pacman.conf. It can't be the expected thing (i.e. a normal GPGDir setting), because we do not have write privileges to the host GPG directory. I could mount over the top of the standard location, and require them not to change it, but I don't think that works in the non-fakemachine case. So I think we have to both parse it out of the pacman.conf and force the user to encode a fixed temporary directory into their pacman.conf (or flip it around and post-process pacman.conf).
- The mirror list needs to be present as an absolute path (as far as I can tell) within the user's pacman.conf. Generating pacman.conf let us hide this detail, but now, again, we're forcing the user's input to contain fixed absolute paths (or we have to post-process the provided pacman.conf).

I'm not keen on parsing pacman.conf in general, or post-processing it. Maybe there's a better option I'm missing here?

evelikov: > * We can't mount `/etc/pacman.conf`; we can't directly change `/etc` in debos, we can mount subdirectories of `/etc`, but not the directory itself. This isn't crippling, but the `-C` arguments are here to stay.

I've naively assumed that we use overlayfs, by which we can indirectly mount individual files. Seems that I was off - sorry.

There's nothing stopping us from creating the `/etc/pacman.conf` and copying the contents of the provided `ConfigFile` right?

> * I believe `pacman-key` runs on `GPGDir` from the provided pacman.conf. It can't be the expected thing (i.e. a normal GPGDir setting), because we do not have write privileges to the host GPG directory.

That's correct - which is why we `mkdir (/etc/pacman.d/gnupg)` before calling `pacman-key`.

> I could mount over the top of the standard location, and require them not to change it, but I don't think that works in the non-fakemachine case.

What is "the non-fakemachine case" - UML? Myself and others have been extensively using fakemachine, so UML is a non-goal in the foreseeable future.

> * The mirror list needs to be present as an absolute path (as far as I can tell) within the user's pacman.conf. Generating pacman.conf let us hide this detail, but now, again, we're forcing the user's input to contain fixed absolute paths (or we have to post-process the provided pacman.conf).
> 
> I'm not keen on parsing pacman.conf in general, or post-processing it. Maybe there's a better option I'm missing here?

You are correct - full/absolute path is required. Considering it's the default in the pacman.conf file, people will run into trouble only if they explicitly change it (afaict nobody does that) - at which point they're on their own.

I appreciate how you're analysing all the possibilities, yet considering the overall silence (reluctance?) by maintainers I think it's best to start (very) small and build from there.

eds-collabora: I'll do my best to give my understanding of the details, but hopefully @obbardc will correct my mistakes... 

> There's nothing stopping us from creating the `/etc/pacman.conf` and copying the contents of the provided `ConfigFile` right?

debos has three modes.

1. No fakemachine. Here, we're running on the host filesystem, as root. `context.Scratchdir` is some temp directory in the current directory. We are limited to chroot/nspawn for compartmentalisation. Since `PacstrapAction` runs before we have anything to chroot into, we can map no directories, we can only specify locations for things [one reason for chmnt is nspawn will not run chrooted onto `/`]
2. fakemachine=kvm. Here, we boot a fakemachine containing sections of the host filesystem. We can choose directories, but not files, to share (via volumes in fakemachine construction). And we can copy files into the temporary filesystem the fakemachine has. `context.Scratchdir` is `/scratch` in the fakemachine.
3. fakemachine=uml. This basically has all the same behaviours as KVM for the purposes of this part.
 
So `PacstrapAction` runs under different constraints to most debos actions, because we can't chroot yet (it's going to make us the viable chroot). That means our base functionality is whatever you could do from the command line as root (without forking - go doesn't really work very well for `fork()`, so you're better off making a dedicated command if you need to).

> That's correct - which is why we `mkdir (/etc/pacman.d/gnupg)` before calling `pacman-key`.

`pacman-key` will make the directory itself, which is fine. But if there is no fakemachine, there's no obvious way within debos to make that work.

> What is "the non-fakemachine case" - UML? Myself and others have been extensively using fakemachine, so UML is a non-goal in the foreseeable future.

I wonder if (@obbardc?) we are willing to commit functionality to the codebase that only works if fakemachine is present? It's not a matter of UML vs KVM, it's a matter of either vs nothing. It would be the first thing in the codebase that errored out if fakemachine wasn't present. I assumed that was unacceptable, but perhaps that was a mistake. It's easy enough technically to make `PreNoMachine` abort the build.
 
> You are correct - full/absolute path is required. Considering it's the default in the pacman.conf file, people will run into trouble only if they explicitly change it (afaict nobody does that) - at which point they're on their own.

That is quite an ugly restriction. A lot depends on where the maintainers want this to go I think. It starts to feel less and less like a cohesive whole, and I expected that to encourage a great deal of resistance.
 
> I appreciate how you're analysing all the possibilities, yet considering the overall silence (reluctance?) by maintainers I think it's best to start (very) small and build from there.

I agree in principle, but I'm not clear what sacrifices the maintainers are willing to make in that respect. Some stronger direction would go a long way here. I'm not clear that there exists any patch in this space that is mergeable, so perhaps we should start the discussion around that point:

What are the hard requirements for this to merge, in terms of complexity, uniformity, and isolation?

obbardc:
closing in favour of !377

PR #277, opened 2021-08-09
Add a pacman action

eds-collabora:


eds-collabora:
Summary of comments on this code to date:

- [ ] [Drop "--color never" from the pacman invocation](https://github.com/santoshmahto7/debos/commit/994a093e6e56ecbd875efb570e44ecb5fefd5e39#r53536394)
- [ ] [Drop timeout patch](https://github.com/santoshmahto7/debos/commit/25634f477acf727ab3375086c866bcc499fb0743#commitcomment-53536450)

actions/pacman_action.go
```diff
@@ -0,0 +1,40 @@
...
+type PacmanAction struct {
+	debos.BaseAction `yaml:",inline"`
+	Packages         []string
+}
+
+func (p *PacmanAction) Run(context *debos.DebosContext) error {
+	p.LogStart()
+
+	pacmanOptions := []string{"pacman", "--color", "never", "--noprogressbar", "--noconfirm", "-Syu"}
+	pacmanOptions = append(pacmanOptions, p.Packages...)
```
eds-collabora, line 31: Drop --colour never (original comment [here](https://github.com/santoshmahto7/debos/commit/994a093e6e56ecbd875efb570e44ecb5fefd5e39#r53536394))

actions/pacman_action.go
```diff
@@ -0,0 +1,40 @@
...
+ - action: pacman
+   packages:
+     - package1
+     - package2
+
+Mandatory properties:
+
+- packages -- list of packages to install
+*/
+package actions
```
obbardc, line 16: wondering if we should have an implementation for `Verify` which checks if pacman (and other used) commands are present on the system; if not we should bail out early?

evelikov (changes requested):

actions/pacman_action.go
eds-collabora: I think (from the pacstrap PR) we wanted to make that a separate follow-up. I have the code, but perhaps doesn't belong in this PR?

actions/pacman_action.go
```diff
@@ -0,0 +1,40 @@
...
+
+func (p *PacmanAction) Run(context *debos.DebosContext) error {
+	p.LogStart()
+
+	pacmanOptions := []string{"pacman", "--noprogressbar", "--noconfirm", "-Syu"}
+	pacmanOptions = append(pacmanOptions, p.Packages...)
+
+	c := debos.NewChrootCommandForContext(*context)
+	err := c.Run("Pacman", pacmanOptions...)
+	if err != nil {
```
obbardc, line 35: you could merge this line with the line above to improve readability

evelikov:
The original comment suggests dropping the `color` and `progressbar` cmdline options, plus some cosmetics. Feel free to ignore the cosmetics, but please drop the `progressbar` option ... or add an inline comment why we need it. With that, the MR is:
```
Reviewed-by: Emil Velikov <emil.l.velikov@gmail.com>
```

eds-collabora:
> The original comment suggests dropping the `color` and `progressbar` cmdline options, plus some cosmetics. Feel free to ignore the cosmetics, but please drop the `progressbar` option ... or add an inline comment why we need it.

I just didn't read the original comment carefully enough, sorry. Fixed.

obbardc:
closing in favour of !377

PR #281, opened 2021-08-13
actions: apt: Fix apt output not shown in gitlab CI

obbardc:


commands.go
```diff
@@ -58,6 +59,14 @@ func (w commandWrapper) out(atEOF bool) {
 	}
 }
 
+func (w commandWrapper) log(s string) {
+	/* remove any special characters which the shell may not support */
+	s = strings.ReplaceAll(s, "\r", "")
+	s = strings.ReplaceAll(s, "\b", "")
+
+	log.Printf("%s | %v", w.label, s)
+}
```
evelikov, line 68: AFAICT this "messes-up" the output in different way. In reality we cannot have our cake and eat it, but we could make this a cmdline option to toggle as needed?

obbardc: ok, I have split that other commit into another PR & will create an issue to manage gitlab getting confused. We can close this PR.

obbardc: https://github.com/go-debos/debos/issues/416

Issue #280, opened 2021-08-12
actions: check for presence of required commands in Verify()


eds-collabora:
Many actions depend on some set of tools being installed on the machine. debos should check that the required commands exist and are runnable, before attempting to run the recipe. A per-action Verify() already exists, and is the natural place for this.

PR #283, opened 2021-08-14
Use Verify to check required commands are present

eds-collabora:
For each action, check that as many commands it needs are available as possible. This can't be meaningfully extended to commands that run in the chroot, or might run in the chroot.

This follows up from comments on #276 which were captured as issue #280.


commands.go
```diff
@@ -277,6 +277,21 @@ func (cmd Command) Run(label string, cmdline ...string) error {
...
 }
 
+func (cmd Command) Run(label string, cmdline ...string) error {
+	w := newCommandWrapper(label, true)
+	defer w.flush()
+	return cmd.run(label, w, cmdline)
+}
+
+func (cmd Command) CheckExists(label string, cmdline ...string) error {
+	w := newCommandWrapper(label, false)
```
obbardc, line 287: is it worth being simple and just using `exec.LookPath` to check the executable exists somewhere rather than running the help commadn and checking the output ?

cmd/debos/debos.go
```diff
@@ -243,7 +243,34 @@ func main() {
 		return
 	}
 
+	if !runInFakeMachine && !fakemachine.InMachine() {
```
obbardc, line 246: looks like a spurious change ? why the additional check of `!runInFakeMachine` ?

eds-collabora: This condition governs PreNoMachine(), which should only be called when we don't intend to run in the fakemachine and we aren't running in the fakemachine. If I removed !runInFakeMachine, PreNoMachine would run on the host before PreMachine when we intend to run in a fakemachine. The condition used to be implicit because the block was after the bailout point for the runInFakeMachine case (that block returns); I had to move it forward so that I could give a solid consistent point at which CheckEnv runs (i.e. _after_ PreMachine _and_ PreNoMachine).

commands.go
eds-collabora: If you'd prefer, I can make that change. Actually running the command seemed safest and simplest to me; it catches things like broken installs, missing libraries, and so forth. Just the presence of an executable doesn't mean you can use it.

commands.go
```diff
@@ -32,12 +32,13 @@ type Command struct {
 
 type commandWrapper struct {
 	label  string
+	writeBeforeFlush bool
```
sjoerdsimons, line 35: can we name this in a somewhat more meaningfull  way?  I had to go over this a few times before i understood what this name meant ;)

commands.go
sjoerdsimons: Fwiw i'm fine either way; I'm not sure trying to run things add much given well, so many things could potentially be broken. What if the tools are slightly different versions? what if a tool isn't bulid with the right options? etc etc. So i think going beyond, does it exist in the path has minimal extra value.. 

That said as things run in fakemachine that may potentially break programs, but that case seems both extremely rare and questionable if you'd detect that with just running `--help`. 

But again either way works for me as i don't think running help on a bunch of tools will have a negative impact either ;)

cmd/debos/debos.go
sjoerdsimons: Yeah this seems correct.. The logic on what runs at a give point has always been a bit... wobbly.. but i don't think you made it worse ;) something to refactor in the future though

sjoerdsimons (changes requested):

obbardc:
@daissi assigning to you to ideally take over & rework.

obbardc:
continued in #598 ; closing this stale PR :-)

PR #284, opened 2021-08-26
actions/image_partition: Add property 'mkfs_parameters'

huongnguyenthi:
Image partition provides some properties for filesystem such as 'features', 'fsuuid'
'mkfs_paramters' allows more flexibility for setting mkfs command's parameters
Setting values in 'mkfs_parameters' are overrided by fixed properties.

Signed-off-by: Nguyen Thi Huong <huong4.nguyenthi@toshiba.co.jp>

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -49,6 +49,7 @@ Yaml syntax for partitions:
 	   flags: list of flags
 	   fsck: bool
 	   fsuuid: string
+	   mkfs_parameters: list of mkfs command line parameters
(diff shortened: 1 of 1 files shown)


sjoerdsimons:
Fwiw we're explicitly not exposing raw passthrough of mkfs features as that would break if the mkfs used changes; debos generally is meant to be more declarative.. So it would be interesting to hear what options you're intending to set. For now i'll close this PR as it has stalled for a while, please feel free to reopen

PR #285, opened 2021-08-27
image_partition: add ability to skip "start" and specify relative offsets

thanhtunng:
Currently start and end value must be specified explicitly

We can ignore start value unless required by using the end of
previous partition.

The size of partition can also be easier to see if using
end value as relative offset (e.g: +10Mib, +5%). However, 
the type of end value must be the same as start's

Example:
```
- action: image-partition
  imagename: "debian-rpi3.img"
  imagesize: 1GB
  partitiontype: msdos
  mountpoints:
    - mountpoint: /
      partition: root
    - mountpoint: /boot/firmware
      partition: firmware
      options: [ x-systemd.automount ]
  partitions:
    - name: firmware
      fs: vfat
      end: +5%
    - name: root
      fs: ext4
      start: 64MB
      end: +256MB
      flags: [ boot ]
```

firmware partition - start: 0% (as default), end: 5%
root partition - start: 64MB. end: 320MB

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -630,6 +630,17 @@ func CalculateOffset(start, end string) (string, error) {
 	return end, nil
 }
 
+func ValidPercentage(offset string) error {
+	val, typ, err := ParseOffset(offset)
+	if err != nil {
+		return err
+	}
+	if typ == "%" && val > 100 {
+		return fmt.Errorf("Size can not exceed 100%%")
+	}
+	return nil
+}
+
 func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 	if len(i.GptGap) > 0 {
 		log.Println("WARNING: special version of parted is needed for 'gpt_gap' option")
@@ -692,11 +703,22 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 		if p.Start == "" {
 			p.Start = prevEnd
 		}
+
+		err = ValidPercentage(p.Start)
+		if err != nil {
+			return err
+		}
+
 		if p.End == "" {
 			return fmt.Errorf("Partition %s missing end", p.Name)
 		}
 
 		p.End, err = CalculateOffset(p.Start, p.End)
+		if err != nil {
+			return err
+		}
+
+		err = ValidPercentage(p.End)
 		if err != nil {
 			return err
 		}


sangorrin:
Hi,

Just to make sure, this is not an **essential** feature for us.
I just thought the interface would be more user-friendly if we used relative offsets.
Perhaps there are other users interested.

Thanks,
Daniel

evelikov:
Tested the MR and it seems kind of broken. So it needs extra work until it's in good state.

Example:

```
architecture: amd64

actions:
  - action: image-partition
    imagename: disk.img
    imagesize: 2G
    partitiontype: gpt
    partitions:
      - name: rootfs
        fs: ext4
        end: +512Mi
      - name: home
        fs: ext4
        end: 100%
    mountpoints:
      - mountpoint: /
        partition: rootfs
      - mountpoint: /home
        partition: home
```

Issues:
 - rootfs start/end unit mismatch - trivial fix, ignore unit when startval == 0
 - creating the home partition outright fails

```
$ parted -a none -s -- /dev/disk/by-fakemachine-label/fakedisk-0 mkpart home ext4 512Mi 100%
Error: You requested a partition from 537MB to 2000MB (sectors 1048576..3906249)
The closest location we can manage is 537MB to 2000MB (sectors 1048577..3906216)
```

actions/image_partition_action.go
```diff
@@ -584,6 +586,61 @@ func (i ImagePartitionAction) PostMachineCleanup(context *debos.DebosContext) er
...
+	/* Extract value of offset (int) and type from string*/
+	var v, t []rune
+	for _, l := range offset {
+		switch {
+		case l >= '0' && l <= '9':
+			v = append(v, l)
+		case l >= 'A' && l <= 'Z', l >= 'a' && l <= 'z', l == '%':
+			t = append(t, l)
+		}
+	}
```
evelikov, line 599: This loop here allows for arbitrary alpha-numeric word soup. For example "A01Z99%" is considered sane by the above.
Let's use a regex akin to the one in https://github.com/go-debos/debos/pull/292.

actions/image_partition_action.go
```diff
@@ -584,6 +586,61 @@ func (i ImagePartitionAction) PostMachineCleanup(context *debos.DebosContext) er
...
+	if err != nil {
+		return 0, "", fmt.Errorf("Can not parse %s to integer", string(v))
+	}
+
+	return val, typ, nil
+}
+
+func CalculateOffset(start, end string) (string, error) {
+	/* Do addition if end value uses a '+' prefix */
+	if strings.HasPrefix(end, "+") {
```
evelikov, line 613: Nit: I would move the "HasPrefix" check in the caller.

thanhtunng:
```
rootfs start/end unit mismatch - trivial fix, ignore unit when startval == 0
```
@evelikov  Yeah the mismatch issue I already mentioned in commit message, but the case startval ==0 I didnt think about, thanks~
```
$ parted -a none -s -- /dev/disk/by-fakemachine-label/fakedisk-0 mkpart home ext4 512Mi 100%
Error: You requested a partition from 537MB to 2000MB (sectors 1048576..3906249)
The closest location we can manage is 537MB to 2000MB (sectors 1048577..3906216)
```
`MiB`? Furthermore not sure able to use "MiB" unit unless https://github.com/go-debos/debos/pull/292 is merged

evelikov:
You might be conflating things:
 Mi/MiB handling with partitions just works in the master branch, since the raw data from the yaml is passed to directly into parted as seen in the example above. This MR breaks that functionality and I think you should look into why.

The code in https://github.com/go-debos/debos/pull/292 covers the imagesize, which is a completely orthogonal thing.

obbardc:
@thanhtunng Do you have an update about this?

evelikov:
@obbardc I'm keen on pushing this forward. Is there any particular changes you'd like to see (aka anything more than what I've raised already)? I'm thinking about adding some tests to the CI - the snippet above + another one with explicit start and other units.

Any pointers how to properly fix the parted error mentioned before - https://github.com/go-debos/debos/pull/285#issuecomment-1018480147. It seems like we're either hitting a parted bug or an off-by-one error in debos.

obbardc:
@thanhtunng @evelikov Any update here ? It'd be great to get this reworked.

evelikov:
Nothing from me since debos maintainership has been MIA over the last 6 months or so. Will be quite busy in the upcoming 2-3 months so it's unlikely I'll have the time, sorry.

actions/image_partition_action.go
```diff
@@ -80,6 +75,11 @@ GPT sets the partition type to Linux Swap.
 For msdos partition types hex codes see: https://en.wikipedia.org/wiki/Partition_type
 For gpt partition type GUIDs see: https://systemd.io/DISCOVERABLE_PARTITIONS/
 
+- start -- offset from beginning of the disk there the partition starts.
```
obbardc, line 79: It'd be good to note why this is optional and what would happen if it isn't actually provided.

actions/image_partition_action.go
```diff
@@ -80,6 +75,11 @@ GPT sets the partition type to Linux Swap.
...
+
+For 'start' and 'end' properties offset can be written in human readable
```
obbardc, line 81: It'd be good to duplicate this inline so that both `start` and `end` properties have this.

actions/image_partition_action.go
```diff
@@ -642,12 +643,14 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 		}
 
 		if p.Start == "" {
-			return fmt.Errorf("Partition %s missing start", p.Name)
+			p.Start = prevEnd
```
obbardc, line 704: It'd be good to move this logic into the `Run` stage rather than `Verify` stage potentially with some comment  about what is going on.

actions/image_partition_action.go
```diff
@@ -642,12 +643,14 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
...
 		}
 		if p.End == "" {
```
obbardc, line 712: This really should have a test case as well.

actions/image_partition_action.go
obbardc: @thanhtunng Any update?

actions/image_partition_action.go
obbardc: I agree.

actions/image_partition_action.go
```diff
@@ -600,6 +646,7 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 	num := 1
 	prevEnd := "0%"
 	for idx, _ := range i.Partitions {
+		var err error
```
obbardc, line 660: you can remove this by having later on `p.End, err := blah`

actions/image_partition_action.go
```diff
@@ -649,6 +696,11 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 			return fmt.Errorf("Partition %s missing end", p.Name)
 		}
 
+		p.End, err = CalculateOffset(p.Start, p.End)
```
obbardc, line 716: as above, check here if prefix is `+`

actions/image_partition_action.go
```diff
@@ -630,6 +630,17 @@ func CalculateOffset(start, end string) (string, error) {
 	return end, nil
 }
 
+func ValidPercentage(offset string) error {
+	val, typ, err := ParseOffset(offset)
+	if err != nil {
+		return err
+	}
+	if typ == "%" && val > 100 {
+		return fmt.Errorf("Size can not exceed 100%%")
```
obbardc, line 639: nit: two %%

obbardc (changes requested):
We are missing some tests & fixing some nits. I'd be happy to merge once the above is complete.

actions/image_partition_action.go
```diff
@@ -60,12 +60,8 @@ unique.
...
 'none' fs type should be used for partition without filesystem.
 
-- start -- offset from beginning of the disk there the partition starts.
-
-- end -- offset from beginning of the disk there the partition ends.
-
-For 'start' and 'end' properties offset can be written in human readable
-form -- '32MB', '1GB' or as disk percentage -- '100%'.
+- end -- offset from beginning of the disk there the partition ends or
+offset from beginning of where the partition starts specified with prefix "+"
```
sjoerdsimons, line 64: Rather then having end be either end point on the disk *or* the size of the disk it would make more sense to have a seperate size property and require either end or size to be provided.

size should be in similar to the units parted accepts (https://www.gnu.org/software/parted/manual/html_node/unit.html) ideally; The image size parsing is *mostly* there, only `%` and `s` really sould be handled. I don't think anyone still cares about cylinders :).

same parsing should probably be use for start and end rather then the current state were we just pass on whatever the user gives to parted and hope :).. 

That way it's both more clear for end-user, cleanup one of the debos oddities and more flexible then this implementation where there seems to be very specific  requirements between the start and end value "types"

sjoerdsimons (changes requested):

obbardc:
@daissi assigning to you to ideally take over & rework.

obbardc:
continued in #599 ; closing this stale PR :-)

PR #291, opened 2021-09-11
docker: Don't use qemu from experimental 

obbardc:
qemu is no longer in experimental so the container build
will now fail. The required patches for qemu have been
backported into bullseye so let's revert using qemu from
experimental.

Link: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=988174

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -56,8 +56,7 @@ LABEL org.label-schema.docker.cmd 'docker run \
 
 # debos runtime dependencies
 # ca-certificates is required to validate HTTPS certificates when getting debootstrap release file
-RUN echo "deb http://deb.debian.org/debian experimental main" > /etc/apt/sources.list.d/experimental.list && \
-    apt-get update && \
+RUN apt-get update && \
     apt-get install -y --no-install-recommends \
         apt-transport-https \
         binfmt-support \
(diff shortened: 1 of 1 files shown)


sjoerdsimons:
I don't think we should be using testing as the base for the container; Apart from qemu, for which the fix will likely be done as a bullseey fix is there anything else you're considering here?

obbardc:
> I don't think we should be using testing as the base for the container; Apart from qemu, for which the fix will likely be done as a bullseey fix is there anything else you're considering here?

just to follow this up; since qemu hasn't been backported to bullsye; i am currently waiting for the qemu maintainer to backport the patches which should hopefully be at the next bullseye point release.

sjoerdsimons:
> > I don't think we should be using testing as the base for the container; Apart from qemu, for which the fix will likely be done as a bullseey fix is there anything else you're considering here?
> 
> just to follow this up; since qemu hasn't been backported to bullsye; i am currently waiting for the qemu maintainer to backport the patches which should hopefully be at the next bullseye point release.

Sounds good; For my own notes this is Debian bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=988174

obbardc:
with qemu version `1:5.2+dfsg-11+deb11u1` released into `bullseye-security` this can now be merged

PR #292, opened 2021-09-29
actions/image-partition: Allow binary unit measurements in the imagesize

venkata-pyla:
Image partitions should be allowed to create with two different
 unit measurements binary and decimal.

 binary units(KiB, MiB, GiB, TiB, PiB) are calculated in multiples of 1024,
 and decimal units(KB, MB, GB, TB, PB) are caculated in multiples of 1000.

 e.g:
  imagesize: 1GiB = 1073741824
  imagesize: 1GB  = 1000000000

Signed-off-by: venkata pyla <venkata.pyla@toshiba-tsip.com>

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -160,6 +160,7 @@ import (
 	"strings"
 	"syscall"
 	"time"
+	"regexp"
 
 	"github.com/go-debos/debos"
 )
@@ -678,7 +679,17 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 		}
 	}
 
-	size, err := units.FromHumanSize(i.ImageSize)
+	// Calculate the size based on the unit (binary or decimal)
+	// binary units are multiples of 1024 - KiB, MiB, GiB, TiB, PiB
+	// decimal units are multiples of 1000 - KB, MB, GB, TB, PB
+	var getSizeValueFunc func(size string) (int64, error)
+	if regexp.MustCompile(`^[0-9]+[kmgtp]ib+$`).MatchString(strings.ToLower(i.ImageSize)) {
+		getSizeValueFunc = units.RAMInBytes
+	} else {
+		getSizeValueFunc = units.FromHumanSize
+	}
+
+	size, err := getSizeValueFunc(i.ImageSize)
 	if err != nil {
 		return fmt.Errorf("Failed to parse image size: %s", i.ImageSize)
 	}


actions/image_partition_action.go
```diff
@@ -678,7 +679,17 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 		}
 	}
 
-	size, err := units.FromHumanSize(i.ImageSize)
+	// Calculate the size based on the unit (binary or decimal)
+	// binary units are multiples of 1024 - KiB, MiB, GiB, TiB, PiB
+	// decimal units are multiples of 1000 - KB, MB, GB, TB, PB
+	var getSizeValueFunc func(size string) (int64, error)
+	if regexp.MustCompile(`^[0-9]+[kKmMgGtTpP][iI][bB]+$`).MatchString(i.ImageSize){
```
obbardc, line 686: space before `{`

actions/image_partition_action.go
```diff
@@ -678,7 +679,17 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
...
+		getSizeValueFunc = units.RAMInBytes
+	}else {
```
obbardc, line 688: space after `}`

actions/image_partition_action.go
```diff
@@ -678,7 +679,17 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 		}
 	}
 
-	size, err := units.FromHumanSize(i.ImageSize)
+	// Calculate the size based on the unit (binary or decimal)
+	// binary units are multiples of 1024 - KiB, MiB, GiB, TiB, PiB
+	// decimal units are multiples of 1000 - KB, MB, GB, TB, PB
+	var getSizeValueFunc func(size string) (int64, error)
+	if regexp.MustCompile(`^[0-9]+[kKmMgGtTpP][iI][bB]+$`).MatchString(i.ImageSize){
+		getSizeValueFunc = units.RAMInBytes
```
obbardc, line 687: would it be worth renaming `RAMInBytes` to something more meaningful ?

actions/image_partition_action.go
```diff
@@ -678,7 +679,17 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 		}
 	}
 
-	size, err := units.FromHumanSize(i.ImageSize)
+	// Calculate the size based on the unit (binary or decimal)
```
obbardc, line 682: does this change the current behaviour ?

actions/image_partition_action.go
obbardc: also, rather than a regex you could do something a bit more readable like convert to lowercase, then check if contains mib, gib etc ?

obbardc (changes requested):

actions/image_partition_action.go
venkata-pyla: we can't do this as the the method name is defined in another repository https://github.com/docker/go-units/blob/master/size.go#L86

actions/image_partition_action.go
venkata-pyla: Thanks for the suggestion, i have modified the code and amended.

actions/image_partition_action.go
venkata-pyla: i corrected this and amended, please review, thanks

sjoerdsimons (approved):

obbardc:
@venkata-pyla There's a floating merge commit, can you remove that and rebase on top of the latest main ? :-)

obbardc (approved):

Issue #293, opened 2021-10-01
Support creating extend and logical partitions in case of msdos


daichifukui:
Hi.

In case the partition type is msdos, debos always set the partition label to primary for parted.
However, `parted` as such supports not only primary but extended and logical as a partition label.
Someone may still want to use msdos due to their limitation of BIOS, firmware or hardware.

If it makes sense, can you consider adding the extended and logical support?

Based on the background mentioned above, I have created a patch for this. 

https://github.com/daichifukui/debos/commit/5a38a0bb78b50a99755224ce6596b5ea946e0cd7

The mechanism is that if the number of partitions is more than 4, 3 of them are automatically created as primary, 1 of them is extended, and the remaining partitions are logical.  
If the number is less than or equal to 4, all of the partitions are created as primary.  

This patch was tested with the following two partition layouts. Both of the patterns worked fine.  

**CASE1**: 4 partitions (all of them are supposed to be primary):  
<details>
<summary> Partition layout case1 (click me to expand) </summary>

```
  - action: image-partition
    description: "Creating RW partitions"
    imagename: {{ $imagename }}
    imagesize: {{ $imagesize }}
    partitiontype: msdos
    ptuuid: {{ $ptuuid }}
    partitions:
      - name: BIOS
        fs: fat32
        start: 1MiB
        end: 2MiB
        partuuid: {{ $bios_partuuid }}
      - name: EFI
        fs: fat32
        start: 2MiB
        end: 130MiB
        partuuid: {{ $efi_partuuid }}
        fsuuid: {{ $efi_fsuuid }}
      - name: ROOTFS
        fs: {{ $filesystem }}
        start: 130MiB
        end: 90%
        partuuid: {{ $rootfs_partuuid }}
        fsuuid: {{ $rootfs_fsuuid }}
      - name: ROOTFS2
        fs: {{ $filesystem }}
        start: 90%
        end: 95%
```

</details>

**CASE2**: 5 partitions (Partitions are supposed to be 3 primary ones, 1 extended one, 1 logical one):  
<details>
<summary> Partition layout case2 (click me to expand) </summary>

```
  - action: image-partition
    description: "Creating RW partitions"
    imagename: {{ $imagename }}
    imagesize: {{ $imagesize }}
    partitiontype: msdos
    ptuuid: {{ $ptuuid }}
    partitions:
      - name: BIOS
        fs: fat32
        start: 1MiB
        end: 2MiB
        partuuid: {{ $bios_partuuid }}
[...]

PR #296, opened 2021-11-05
actions/image-partition: allow extended and logical

daichifukui:
This commit allows us to create extended or logical
partitions on MBR partition table (msdos) in order to
address the following issues:

* Because part-type for parted is hardcoded with primary,
we cannot create more than 4 partitions using extended and
logical.

* Even if we manage to create extended and logical partitions,
the verificaton step fails because the Verify function assumes
that the partition number is always sequetially incremental.

This commit makes it possible that if the number of partitions
is more than 4, 3 of them are automatically created as primary,
1 of them is extended, and the remaining partitions are logical.

Signed-off-by: Daniel Sangorrin <daniel.sangorrin@toshiba.co.jp>
Signed-off-by: Daichi Fukui <daichi1.fukui@toshiba.co.jp>

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -479,7 +479,7 @@ func (i ImagePartitionAction) Run(context *debos.DebosContext) error {
 		if i.PartitionType == "gpt" {
 			name = p.PartLabel
 		} else {
-			if cap(i.Partitions) <= 4 {
+			if len(i.Partitions) <= 4 {
 				name = "primary"
 			} else {
 				if idx < 3 {
@@ -670,7 +670,7 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 	for idx, _ := range i.Partitions {
 		p := &i.Partitions[idx]
 
-		if idx == 3 && cap(i.Partitions) > 4 {
+		if idx == 3 && len(i.Partitions) > 4 {
 			var name string
 			var part Partition
 
@@ -678,7 +678,7 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 			part.number    = idx+1
 			part.Name      = name
 			part.Start     = p.Start
-			tmp_n          := cap(i.Partitions)-1
+			tmp_n          := len(i.Partitions)-1
 			tmp            := &i.Partitions[tmp_n]
 			part.End       = tmp.End
 			part.FS        = "none"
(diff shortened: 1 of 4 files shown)


daichifukui:
This is related to https://github.com/go-debos/debos/issues/293

daichifukui:
Hi.

It's been a while since we began this discussion.
Does anyone have an opinion or feedback on this idea?

actions/image_partition_action.go
```diff
@@ -479,7 +479,17 @@ func (i ImagePartitionAction) Run(context *debos.DebosContext) error {
 		if i.PartitionType == "gpt" {
 			name = p.PartLabel
 		} else {
-			name = "primary"
+			if cap(i.Partitions) <= 4 {
```
sjoerdsimons, line 482:
```suggestion
			if len(i.Partitions) <= 4 {
```
cap is the capacity of the  array; the actual length might be shorter

actions/image_partition_action.go
```diff
@@ -656,6 +666,35 @@ func (i ImagePartitionAction) PostMachineCleanup(context *debos.DebosContext) er
 }
 
 func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
+
+	for idx, _ := range i.Partitions {
+		p := &i.Partitions[idx]
+
+		if idx == 3 && cap(i.Partitions) > 4 {
```
sjoerdsimons, line 673:
```suggestion
		if idx == 3 && len(i.Partitions) > 4 {
```
same

actions/image_partition_action.go
```diff
@@ -656,6 +666,35 @@ func (i ImagePartitionAction) PostMachineCleanup(context *debos.DebosContext) er
...
+			var name string
+			var part Partition
+
+			name = "extended"
+			part.number    = idx+1
+			part.Name      = name
+			part.Start     = p.Start
+			tmp_n          := cap(i.Partitions)-1
```
sjoerdsimons, line 681:
```suggestion
			tmp_n          := len(i.Partitions)-1
```

sjoerdsimons (changes requested):

sjoerdsimons (approved):

robertkirkman:
Hi I don't know enough about the project to know how to create a minimal reproducible example, but when i use [this](https://github.com/foxlet/VaporOS/blob/master/vapor-steamdeck-main.yaml) file with `debos`, if I use commit 03ec3cbfab6a73087ebbb58317f65e117342c1ba of `debos`  it works but if I use commit da773ccbc899d2f5b236f80db303062693f6f2b6 or any of the recent commits after that, the `debos` command gets stuck inside the fakemachine because of a partition named "extended" getting incorrectly created in the wrong place even though the `.yaml` file specifies `gpt` and not `mbr`

obbardc:
@robertkirkman Can you please open a separate issue about this so it doesn't get lost ?

robertkirkman:
ok I opened that here #473

PR #297, opened 2021-11-15
Allow setting volume id, partition table uuid, partition uuid

huongnguyenthi:
These patches provide setting for volume id of vfat, partition table uuid (disk identifier) 
and partition uuid (only for gpt) instead of random values.

The purpose  is controlling random data and achieve reproducible builds.

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -57,6 +57,7 @@ Yaml syntax for partitions:
 	   flags: list of flags
 	   fsck: bool
 	   fsuuid: string
+	   partuuid: string
 
 Mandatory properties:
 
@@ -99,6 +100,8 @@ checks in boot time. By default is set to `true` allowing checks on boot.
 
 - fsuuid -- file system UUID string. This option is only supported for btrfs,
 ext2, ext3, ext4 and xfs.
+
+- partuuid -- GPT partition UUID string.
 
 Yaml syntax for mount points:
 
@@ -178,6 +181,7 @@ type Partition struct {
 	Name      string
 	PartLabel string
 	PartType  string
+	PartUUID  string
 	Start     string
 	End       string
 	FS        string
@@ -504,6 +508,13 @@ func (i ImagePartitionAction) Run(context *debos.DebosContext) error {
 			}
 		}
 
+		/* PartUUID will only be set for gpt partitions */
+		if len(p.PartUUID) > 0 {
+			err = debos.Command{}.Run("sfdisk", "sfdisk", "--part-uuid", context.Image, fmt.Sprintf("%d", p.number), p.PartUUID)
+			if err != nil {
+				return err
+			}
+		}
 
 		devicePath := i.getPartitionDevice(p.number, *context)
 
@@ -684,6 +695,18 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 
 		if i.PartitionType != "gpt" && p.PartLabel != "" {
 			return fmt.Errorf("Can only set partition partlabel on GPT filesystem")
+		}
+
+		if len(p.PartUUID) > 0 {
+			switch i.PartitionType {
+			case "gpt":
+				_, err := uuid.Parse(p.PartUUID)
+				if err != nil {
+					return fmt.Errorf("Incorrect partition UUID %s", p.PartUUID)
+				}
+			default:
+				return fmt.Errorf("Setting the partition UUID is not supported for %s", i.PartitionType)
+			}
 		}
 
 		if p.PartType != "" {
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -12,6 +12,7 @@ Yaml syntax:
    imagename: image_name
    imagesize: size
    partitiontype: gpt
+   diskid: string
    gpt_gap: offset
    partitions:
      <list of partitions>
@@ -36,6 +37,13 @@ Partition properties are described below.
 
 - mountpoints -- list of mount points for partitions.
 Properties for mount points are described below.
+
+Optional properties:
+
+- diskid -- disk unique identifier string. For 'gpt' partition table, 'diskid'
+should be in GUID format (e.g.: '00002222-4444-6666-AAAA-BBBBCCCCFFFF' where each
+character is an hexadecimal digit). For 'msdos' partition table, 'diskid' should be
+a 32 bits hexadecimal number (e.g. '1234ABCD' without any dash separator).
 
 Yaml syntax for partitions:
 
@@ -192,6 +200,7 @@ type ImagePartitionAction struct {
 	ImageName        string
 	ImageSize        string
 	PartitionType    string
+	DiskID           string
 	GptGap           string "gpt_gap"
 	Partitions       []Partition
 	Mountpoints      []Mountpoint
@@ -437,6 +446,15 @@ func (i ImagePartitionAction) Run(context *debos.DebosContext) error {
 	if err != nil {
 		return err
 	}
+
+	if len(i.DiskID) > 0 {
+		command := []string{"sfdisk", "--disk-id", context.Image, i.DiskID}
+		err = debos.Command{}.Run("sfdisk", command...)
+		if err != nil {
+			return err
+		}
+	}
+
 	for idx, _ := range i.Partitions {
 		p := &i.Partitions[idx]
 
@@ -611,6 +629,23 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 		_, err := units.FromHumanSize(i.GptGap)
 		if err != nil {
 			return fmt.Errorf("Failed to parse GPT offset: %s", i.GptGap)
+		}
+	}
+
+	if len(i.DiskID) > 0 {
+		switch i.PartitionType {
+		case "gpt":
+			_, err := uuid.Parse(i.DiskID)
+			if err != nil {
+				return fmt.Errorf("Incorrect disk GUID %s", i.DiskID)
+			}
+		case "msdos":
+			_, err := hex.DecodeString(i.DiskID)
+			if err != nil || len(i.DiskID) != 8 {
+				return fmt.Errorf("Incorrect disk ID %s, should be 32-bit hexadecimal number", i.DiskID)
+			}
+			// Add 0x prefix
+			i.DiskID = "0x" + i.DiskID
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


huongnguyenthi:
Hello,

Could you please check this PR and let me know your feedback?
The build failure might be caused by server side so could you please help to check or re-run build test?
Build process is still successful in my local environment.

Thanks,
Huong

actions/image_partition_action.go
```diff
@@ -12,6 +12,7 @@ Yaml syntax:
    imagename: image_name
    imagesize: size
    partitiontype: gpt
+   ptuuid: string
```
obbardc, line 15:
```suggestion
   partuuid: string
```

actions/image_partition_action.go
```diff
@@ -37,6 +38,13 @@ Partition properties are described below.
 - mountpoints -- list of mount points for partitions.
 Properties for mount points are described below.
 
+Optional properties:
+
+- ptuuid -- partition table UUID string. For 'gpt' partition tables, 'ptuuid' should
```
obbardc, line 43:
```suggestion
- partuuid -- partition table UUID string. For 'gpt' partition tables, 'partuuid' should
```

actions/image_partition_action.go
```diff
@@ -37,6 +38,13 @@ Partition properties are described below.
...
+be in GUID format (e.g.: '00002222-4444-6666-AAAA-BBBBCCCCFFFF' where each character
```
obbardc, line 44: does it work with lowercase uuids too ?

actions/image_partition_action.go
```diff
@@ -37,6 +38,13 @@ Partition properties are described below.
...
+is an hexadecimal digit). For 'msdos' partition tables, 'ptuuid' should be a 32 bits
+hexadecimal number (e.g. '1234ABCD' without any dash separator).
```
obbardc, line 46: does it work with lowercase uuids ?

actions/image_partition_action.go
```diff
@@ -191,6 +204,7 @@ type ImagePartitionAction struct {
 	ImageName        string
 	ImageSize        string
 	PartitionType    string
+	PTUUID           string
```
obbardc, line 207:
```suggestion
	PartUUID         string
```

actions/image_partition_action.go
```diff
@@ -420,6 +437,15 @@ func (i ImagePartitionAction) Run(context *debos.DebosContext) error {
 	if err != nil {
 		return err
 	}
+
+	if i.PTUUID != "" {
+		command := []string{"sfdisk", "--disk-id", context.Image, i.PTUUID}
```
obbardc, line 442: do we need to install sfdisk in Dockerfile ?

actions/image_partition_action.go
```diff
@@ -597,6 +629,24 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
...
 
+	if i.PTUUID != "" {
+		switch i.PartitionType {
+		case "gpt":
+			_, err := uuid.Parse(i.PTUUID)
+			if err != nil {
+				return fmt.Errorf("Incorrect PTUUID %s", i.PTUUID)
+			}
+		case "msdos":
+			ptUUIDLen := 8
```
obbardc, line 640:
```suggestion
			uuidLen := 8
```

obbardc (changes requested):
PS: you can re-run the check by force-pushing OR closing/reopening the PR

actions/image_partition_action.go
huongnguyenthi: Thanks for your review and suggestion.
`ptuuid` means uuid  value for **partition table**
 `partuuid`  has been used for **partition** uuid variable in https://github.com/go-debos/debos/pull/297/commits/24b4a83cd978534305b9c6c43e268d196a14f390

actions/image_partition_action.go
huongnguyenthi: Yes, it does. I also verified with lowercase uuids.

actions/image_partition_action.go
huongnguyenthi: Yes, it does.

actions/image_partition_action.go
huongnguyenthi: Yes, sure. I will revise it.

actions/image_partition_action.go
huongnguyenthi: No, we don't need to do this.
`fdisk ` package contains `sfdisk` binary.
Please refer to https://packages.debian.org/bullseye/amd64/fdisk/filelist

actions/image_partition_action.go
```diff
@@ -191,6 +204,7 @@ type ImagePartitionAction struct {
 	ImageName        string
 	ImageSize        string
 	PartitionType    string
+	UUID           string
```
obbardc, line 207: Need two extra spaces before string to keep formatting

actions/image_partition_action.go
```diff
@@ -469,6 +495,12 @@ func (i ImagePartitionAction) Run(context *debos.DebosContext) error {
 			}
 		}
 
+		if p.PartUUID != "" {
```
obbardc, line 498: May be better to check if len > 0

actions/image_partition_action.go
```diff
@@ -597,6 +629,24 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 		}
 	}
 
+	if i.UUID != "" {
```
obbardc, line 632: May be better to check if len > 0

actions/image_partition_action.go
```diff
@@ -628,6 +685,18 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
 			return fmt.Errorf("Can only set partition partlabel on GPT filesystem")
 		}
 
+		if p.PartUUID != "" {
```
obbardc, line 688: Check Len > 0

actions/image_partition_action.go
obbardc: Also check if it's gpt

obbardc (changes requested):

actions/image_partition_action.go
huongnguyenthi: I think  no need to check `gpt` here because  `gpt` is  checked when validate  `partUUID`  in `Verify()` function

actions/image_partition_action.go
huongnguyenthi: Thanks for your comment. I will revise it.

actions/image_partition_action.go
obbardc: yeah, it was more so we don't get confused in future.

Maybe a comment as a reminder would be OK ? Like: `/* PartUUID will only be set for gpt partitions */`

huongnguyenthi: Yes, I will do that.

actions/image_partition_action.go
```diff
@@ -12,6 +12,7 @@ Yaml syntax:
    imagename: image_name
    imagesize: size
    partitiontype: gpt
+   uuid: string
```
obbardc, line 15: After some thought, `partitionuuid` is probably best

actions/image_partition_action.go
```diff
@@ -597,6 +630,24 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
...
 
+	if len(i.UUID) > 0 {
+		switch i.PartitionType {
+		case "gpt":
+			_, err := uuid.Parse(i.UUID)
+			if err != nil {
+				return fmt.Errorf("Incorrect UUID %s", i.UUID)
+			}
+		case "msdos":
+			uuidLen := 8
```
obbardc, line 641: remove this

actions/image_partition_action.go
```diff
@@ -597,6 +630,24 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
...
+			_, err := hex.DecodeString(i.UUID)
+			if len(i.UUID) != uuidLen || err != nil {
```
obbardc, line 643:
```suggestion
			if err != nil || len(i.UUID) != 8 {
```

actions/image_partition_action.go
```diff
@@ -614,12 +665,19 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
...
-			if p.FS == "btrfs" || p.FS == "ext2" || p.FS == "ext3" || p.FS == "ext4" || p.FS == "xfs" {
+			switch p.FS {
+			case "btrfs", "ext2", "ext3", "ext4", "xfs":
 				_, err := uuid.Parse(p.FSUUID)
 				if err != nil {
 					return fmt.Errorf("Incorrect UUID %s", p.FSUUID)
 				}
-			} else {
+			case "vfat", "fat32":
+				fsUUIDLen := 8
```
obbardc, line 675: remove this

actions/image_partition_action.go
```diff
@@ -614,12 +665,19 @@ func (i *ImagePartitionAction) Verify(context *debos.DebosContext) error {
...
+				_, err := hex.DecodeString(p.FSUUID)
+				if len(p.FSUUID) != fsUUIDLen || err != nil {
```
obbardc, line 677:
```suggestion
				if err != nil || len(p.FSUUID) != 8 {
```

obbardc (changes requested):
almost there...!

actions/image_partition_action.go
huongnguyenthi: Thanks for your suggestion. It looks consistent with `partitiontype` variable.

obbardc (approved):

actions/image_partition_action.go
```diff
@@ -192,6 +200,7 @@ type ImagePartitionAction struct {
 	ImageName        string
 	ImageSize        string
 	PartitionType    string
+	PartitionUUID    string
```
sjoerdsimons, line 207: This should really be called DiskGUID or somesuch (as per the gpt naming) or Disk id/disk signature (as per msdos naming) not partition UUID. 

Partition UUID  typically refers to the per partition UUID  available in GPT tables. Thee naming/descripton actually made me think that was what this was implementing untill i saw the sfdisk callout.

sjoerdsimons (changes requested):
Seems mostly good to go;  just renaming partitionuuid to e.g. disk guid so it's less confusing

sjoerdsimons (approved):

PR #299, opened 2022-01-03
Added verbose parameter in the README file

zamansoum:
added verbose parameter ( -v) in the go get command to track progress during the install process on Debian


sjoerdsimons:
With the move to go modules and go get being deprecated for installing binaries now this should be converted to go install it seems; doesn't help that that seems unable to use current HEAD of he default branch so those installation instructions seem broken atm regardless...

PR #300, opened 2022-01-04
ostree-deploy: Unreference all libostree variables

obbardc:
Unreference the ostree origin variable which wasn't
yet unreferenced.

Defer the unreference of all of the variables which
could be created from libostree, so they are unrefrenced
in the case of an error path returning before the
Unref() method call.

Fixes: 9b648141392f4 ("ostree-deploy: fix umount error")

### DIFF
--- a/actions/ostree_deploy_action.go
+++ b/actions/ostree_deploy_action.go
@@ -121,6 +121,12 @@ func (ot *OstreeDeployAction) Run(context *debos.DebosContext) error {
 	repoPath := "file://" + path.Join(context.Artifactdir, ot.Repository)
 
 	sysroot := ostree.NewSysroot(context.Rootdir)
+
+	/* libostree keeps some information, like repo lock file descriptor, in
+	 * thread specific variables. As GC can be run from another thread, it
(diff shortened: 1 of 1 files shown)


actions/ostree_deploy_action.go
```diff
@@ -121,6 +121,12 @@ func (ot *OstreeDeployAction) Run(context *debos.DebosContext) error {
 	repoPath := "file://" + path.Join(context.Artifactdir, ot.Repository)
 
 	sysroot := ostree.NewSysroot(context.Rootdir)
+
+	/* libostree keeps some information, like repo lock file descriptor, in
+	 * thread specific variables. As GC can be run from another thread, it
+	 * may not been able to access this, preventing to free them correctly.
+	 * To prevent this, explicitly dereference libostree objects. */
+	defer sysroot.Unref()
```
adalessandro, line 129: Calling `Unref()` doesn't seem to be safe, as there's a finalizer set to call `Unref()` as well, see https://github.com/sjoerdsimons/ostree-go/blob/master/pkg/otbuiltin/sysroot.go#L36

PR #302, opened 2022-01-07
actions: run: Check script exists during recipe verification

obbardc:
If a script doesn't exist or the permissions are wrong, the
recipe exits during execution. Let's check early if the
script exists on the host filesystem and that the file has
executable permission bit set.


actions/run_action.go
```diff
@@ -68,6 +70,22 @@ func (run *RunAction) Verify(context *debos.DebosContext) error {
 	if run.Script == "" && run.Command == "" {
 		return errors.New("Script and Command both cannot be empty")
 	}
+
+	if run.Script != "" {
+		argv := strings.SplitN(run.Script, " ", 2)
+		script := debos.CleanPathAt(argv[0], context.RecipeDir)
```
sjoerdsimons, line 76: in PreMachine we roughly have the same code to determine the path of the script ; Probably worth havign a single utility function for it

actions/run_action.go
```diff
@@ -68,6 +70,22 @@ func (run *RunAction) Verify(context *debos.DebosContext) error {
...
+
+		stat, err := os.Stat(script)
+		if err != nil {
+			return err
+		}
+
+		/* check the script is executable */
+		if stat.IsDir() || stat.Mode()&0111 == 0 {
```
sjoerdsimons, line 84: what if the script is a symlink?

sjoerdsimons:
fwiw this would be a great change to start adding test for

actions/run_action.go
obbardc: in a separate commit i reworked that so we do that determination only once and earlier which improves readability to me

actions/run_action.go
obbardc: Split the executable/dir check into two

it also works; clarified that test in the commit message

actions/run_action.go
```diff
@@ -120,7 +120,7 @@ func (run *RunAction) doRun(context debos.DebosContext) error {
 		}
 		cmdline = []string{run.scriptPath}
 		cmdline = append(cmdline, run.scriptArgs...)
-		label = path.Base(run.Script)
+		label = path.Base(run.scriptPath)
```
sjoerdsimons, line 139: but then you don't see the arguments to the script anymore, which is a loss.. Probably this doesn't belong in this PR (it doesn't help with your stated target i think) but preventing "pollution" we could do either by:
* unify truncating the length like the Command handling does
* printing what gets run *first* and then either not or only providing a minimal label on each line after

actions/run_action.go
```diff
@@ -82,6 +83,21 @@ func (run *RunAction) Verify(context *debos.DebosContext) error {
...
 		run.scriptPath = debos.CleanPathAt(args[0], context.RecipeDir)
 		run.scriptArgs = args[1:]
+
+		/* Check the script exists on the filesystem (following symlinks) */
+		stat, err := os.Stat(run.scriptPath)
+		if err != nil {
+			return err
+		}
+
+		if stat.IsDir() {
```
sjoerdsimons, line 93: Rather then checking if it's a Dir this should proably check if it's a regular file

actions/run_action.go
```diff
@@ -82,6 +83,21 @@ func (run *RunAction) Verify(context *debos.DebosContext) error {
...
+			return fmt.Errorf("script %s is a directory", run.Script)
+		}
+
+		/* Check the script is executable */
+		if stat.Mode()&0111 == 0 {
```
sjoerdsimons, line 98: executable and readable ideally; though i don't think golang gives you nice helpers for that unfortunately :/

.github/workflows/ci.yaml
```diff
@@ -136,6 +136,8 @@ jobs:
             test: { name: "arch", case: "arch" }
           - backend: kvm
             test: { name: "apertis", case: "apertis" }
+          - backend: kvm
+            test: { name: "run", case: "run" }
```
sjoerdsimons, line 140: kvm actually has the highest latency atm because of the runner setup; and really this should probably run on all types.. That said could you merge it with an existing tests so we don't need a new CI job for it ? 

e.g. maybe merge it with the recipes test and the overlay test :)

sjoerdsimons (changes requested):

actions/run_action.go
obbardc: dropped the commit

.github/workflows/ci.yaml
obbardc: i think the tests need to be reworked to use the exitcode check to determine failures :-)

actions/run_action.go
obbardc: done, i hope

actions/run_action.go
obbardc: done, i hope. had to open file to see if it is readable but happy to do it another way.

.github/workflows/ci.yaml
obbardc: @sjoerdsimons had a quick go at adding these tests into exit_code test. I think I need to add two separate recipes for the two expected failures (Marked in the code as TODO for now; will rework once this CI run passes).

actions/run_action.go
```diff
@@ -61,31 +62,38 @@ type RunAction struct {
...
 
 	if run.Script == "" && run.Command == "" {
 		return errors.New("need to set 'script' or 'command'")
 	}
+
+	if run.Script != "" {
+		/* Extract the full script path from the arguments */
+		args := strings.Split(run.Script, " ")
+		run.scriptPath = debos.CleanPathAt(args[0], context.RecipeDir)
+		run.scriptArgs = args[1:]
```
sjoerdsimons: Rather then doing this in the verifying function and passing it along in the action struct, why not split it out in a function you called in both places?

tests/exit_test/run.yaml
```diff
@@ -0,0 +1,45 @@
...
+  - action: run
+    description: Run a script without arguments in chroot
+    chroot: true
+    script: run/scripts/test.sh
+
+  - action: run
+    description: Test running a command
+    command: echo test
+
+  # TODO: This should fail in the Verify stage, we cannot yet check that in CI
```
sjoerdsimons: split that out and run with expect_failure ?

obbardc: done !

actions/run_action.go
obbardc: done i think; commits need cleaning up if you;re happy with logic

obbardc: OK, cleaned up the commits a bit. Hope it;s all good now.

sjoerdsimons (dismissed):

obbardc:
@sjoerdsimons manually rebased on top of main; should be OK to go now.

obbardc:
@tobsan can you review this please? I've rebased & reworked onto your recent changes. Maybe I missed something obvious.

actions/run_action.go
```diff
@@ -104,18 +144,17 @@ func (run *RunAction) doRun(context debos.Context) error {
...
-		if len(script) > 1 {
-			label += " " + strings.Join(script[1:], " ")
+			scriptDir := path.Dir(scriptPath)
+			cmd.AddBindMount(scriptDir, "/tmp/script")
+			scriptPath = strings.Replace(scriptPath, scriptDir, "/tmp/script", 1)
 		}
+
+		cmdline = []string{scriptPath}
+		cmdline = append(cmdline, scriptArgs...)
+		label = path.Base(run.Script)
```
tobsan: To not undo #675 this should be `label = path.Base(scriptPath)` or `path.Base(cmdline[0])`.

actions/run_action.go
```diff
@@ -64,14 +66,54 @@ type RunAction struct {
...
 	if run.PostProcess && run.Chroot {
 		return errors.New("cannot run postprocessing in the chroot")
 	}
 
 	if run.Script == "" && run.Command == "" {
 		return errors.New("need to set 'script' or 'command'")
 	}
+
+	if run.Script != "" {
+		scriptPath, _ := run.extractScriptArgs(run.Script, context.RecipeDir)
```
tobsan, line 93: Would it make sense to put this whole block as a check-executable helper in `filesystem.go`? OTOH probably only the run action that needs it right now...

tobsan (changes requested):

actions/run_action.go
obbardc: good catch, thanks!

actions/run_action.go
obbardc: for now I'd rather leave it in this function, but good point.

actions/run_action.go
```diff
@@ -131,18 +168,17 @@ func (run *RunAction) doRun(context debos.Context) error {
 	}
 
 	if run.Script != "" {
-		script := strings.SplitN(run.Script, " ", 2)
-		script[0] = debos.CleanPathAt(script[0], context.RecipeDir)
+		scriptPath, scriptArgs := run.extractScriptArgs(run.Script, context.RecipeDir)
```
obbardc, line 171: need to make sure the latest rebase includes current logic in the run action file...

PR #301, opened 2022-01-05
Report action cleanup errors & various cleanup fixes for recipe action

obbardc:
recipe action cleanups.

PR #303, opened 2022-01-07
Refactor error handling

obbardc:
Currently the debos error handling sets an exitcode
during running, then exits with that code. Since we
only ever exit with a 1, unless an error occurs while
running fakemachine, let's switch around the error
handling to use the context.State and check the state
after running.

See-also the Fakemachine improvements: https://github.com/go-debos/fakemachine/pull/74

### DIFF
--- a/cmd/debos/debos.go
+++ b/cmd/debos/debos.go
@@ -163,7 +163,7 @@ func main() {
 		// attempt to create a fakemachine
 		m, err = fakemachine.NewMachineWithBackend(options.Backend)
 		if err != nil {
-			log.Printf("error creating fakemachine: %v", err)
+			log.Printf("Couldn't create fakemachine: %v", err)
 
 			/* fallback to running on the host unless the user has chosen
 			 * a specific backend */
@@ -253,7 +253,7 @@ func main() {
 		}
 		memsize, err := units.RAMInBytes(options.Memory)
 		if err != nil {
-			fmt.Printf("Couldn't parse memory size: %v\n", err)
+			log.Printf("Couldn't parse memory size: %v\n", err)
 			context.State = debos.Failed
 			return
 		}
@@ -268,7 +268,7 @@ func main() {
 		if options.ScratchSize != "" {
 			size, err := units.FromHumanSize(options.ScratchSize)
 			if err != nil {
-				fmt.Printf("Couldn't parse scratch size: %v\n", err)
+				log.Printf("Couldn't parse scratch size: %v\n", err)
 				context.State = debos.Failed
 				return
 			}
@@ -318,12 +318,13 @@ func main() {
 
 		exitcode, err := m.RunInMachineWithArgs(args)
 		if err != nil {
-			fmt.Println(err)
+			log.Printf("Couldn't start fakemachine: %v\n", err)
 			context.State = debos.Failed
 			return
 		}
 
 		if exitcode != 0 {
+			log.Printf("fakemachine failed with non-zero exitcode: %d\n", exitcode)
 			context.State = debos.Failed
 			return
 		}
@@ -355,6 +356,7 @@ func main() {
 	if _, err = os.Stat(context.Rootdir); os.IsNotExist(err) {
 		err = os.Mkdir(context.Rootdir, 0755)
 		if err != nil && os.IsNotExist(err) {
+			log.Printf("Couldn't create rootdir: %v\n", err)
 			context.State = debos.Failed
 			return
 		}
--- a/cmd/debos/debos.go
+++ b/cmd/debos/debos.go
@@ -50,37 +50,36 @@ func warnLocalhost(variable string, value string) {
 		    Consider using an address that is valid on your network.`
 
 	if strings.Contains(value, "localhost") ||
-	   strings.Contains(value, "127.0.0.1") ||
-	   strings.Contains(value, "::1") {
+		strings.Contains(value, "127.0.0.1") ||
+		strings.Contains(value, "::1") {
 		log.Printf(message, variable)
 	}
 }
 
-
 func main() {
-	context := debos.DebosContext { &debos.CommonContext{}, "", "" }
+	context := debos.DebosContext{&debos.CommonContext{}, "", ""}
 	var options struct {
-		Backend       string            `short:"b" long:"fakemachine-backend" description:"Fakemachine backend to use" default:"auto"`
-		ArtifactDir   string            `long:"artifactdir" description:"Directory for packed archives and ostree repositories (default: current directory)"`
-		InternalImage string            `long:"internal-image" hidden:"true"`
-		TemplateVars  map[string]string `short:"t" long:"template-var" description:"Template variables (use -t VARIABLE:VALUE syntax)"`
-		DebugShell    bool              `long:"debug-shell" description:"Fall into interactive shell on error"`
-		Shell         string            `short:"s" long:"shell" description:"Redefine interactive shell binary (default: bash)" optionsl:"" default:"/bin/bash"`
-		ScratchSize   string            `long:"scratchsize" description:"Size of disk backed scratch space"`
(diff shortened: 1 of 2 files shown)


obbardc:
There is also https://github.com/go-debos/fakemachine/pull/74 which goes alongside this PR

cmd/debos/debos.go
```diff
@@ -253,7 +253,7 @@ func main() {
 		}
 		memsize, err := units.RAMInBytes(options.Memory)
 		if err != nil {
-			fmt.Printf("Couldn't parse memory size: %v\n", err)
+			log.Printf("Couldn't parse memory size: %v\n", err)
```
evelikov, line 256: To be honest I'm not sure when I should be using one or the other (log vs fmt). Does golang or go-debos have some guideline when to use which one? 

Looking at fakemachine - it exclusively uses fmt. While debos has a 2:1 ratio for fmt vs log. Should any of the other fmt instanced be also swapped - either here or a separate issue/PR?

cmd/debos/debos.go
```diff
@@ -15,18 +15,18 @@ import (
 	"github.com/jessevdk/go-flags"
 )
 
-func checkError(context *debos.DebosContext, err error, a debos.Action, stage string) int {
+func checkError(context *debos.DebosContext, err error, a debos.Action, stage string) bool {
```
evelikov, line 18: Personally I would use with "error", since that gives us the option to bubble it up, plus it seems more consistent with the rest of the codebase.

cmd/debos/debos.go
```diff
@@ -15,18 +15,18 @@ import (
...
 	if err == nil {
-		return 0
+		return true
 	}
 
 	context.State = debos.Failed
 	log.Printf("Action `%s` failed at stage %s, error: %s", a, stage, err)
 	debos.DebugShell(*context)
```
evelikov, line 25: Side nit: is it only me that finds the random call to "DebugShell" in a function call "checkError" odd?

cmd/debos/debos.go
obbardc: In debos we should use log; which adds pretty bits e.g. datestamp. For fakemachine, we don't have a log so fmt is fine.

We should transititon to log in debos ;-)

cmd/debos/debos.go
obbardc: right, the idea here is that if there is an error, instead of quitting we show a debug shell if that's what the user wanted. I'd be happy to accept a PR to add a comment to do that if you like.

commands.go
```diff
@@ -306,14 +310,14 @@ func newQemuHelper(c Command) qemuHelper {
 	case "amd64", "i386":
 		/* Dummy, no qemu */
 	default:
-		log.Panicf("Don't know qemu for Architecture %s", c.Architecture)
+		return q, fmt.Errorf("Don't know qemu for architecture %s", c.Architecture)
```
evelikov, line 313: Should probably be `return nil, err...` aka `s/q/nil/`

cmd/debos/debos.go
obbardc: right that actually should be the case for the functions which consume `checkError` too; currnetly they return an int.....

cmd/debos/debos.go
obbardc: actually, I think I can handle that inside this PR.

cmd/debos/debos.go
evelikov: Having slept on it, I am leaning towards removing the helper all together. As mentioned previously - it masks the `DebugShell` call, but it also masks the `context.State = debos.Failed` handling. The existing codebase has a bunch of the latter open-coded, making it look fairly imbalanced.

On the other hand the `State` is used solely within the single `PostMachineCleanup`. Perhaps we should check/update all actions adding the callback and/or state-check. Mildly related - we're setting up `defer PostMachineCleanup` prior to the `Pre[No,]Machine` calls, which seems odd.

I'm not sure if these are something to care about (or are even genuine issues). Let me know it they make sense and I can open issues for each one - to be fixed at misc later point.

obbardc:
> Any updates on this? As mentioned [above](https://github.com/go-debos/debos/pull/303#issuecomment-1301087693) - the `newQemuHelper` suggestion is the only one of value, where the rest are a matter of taste so don't read too much into them.

It's on my TODO list for Jan.

commands.go
sjoerdsimons: it should indeed

cmd/debos/debos.go
sjoerdsimons: fwiw this helper should really stay; the naming could be better (handleError?) but otherwise it seems ok; 

setting up the defer PostMachineCleanup before any Pre actions is expected such that the cleanups also occur if the Machine setup failed

evelikov:
If you can add some tests, that would be great. The MR from Ed had a recent amount IMHO.

commands.go
obbardc: sorted that one out

obbardc:
Next steps are to merge the commits from #301 into this branch.

sjoerdsimons:
@obbardc please fix the CI errors

obbardc:
> @obbardc please fix the CI errors

Sure, this also needs further rework so I have marked it as a draft.

cmd/debos/debos.go
obbardc: reversed the boolean logic and called it `handleError` - it returns true once an error had been handled.

cmd/debos/debos.go
sjoerdsimons, line 1: You mention "this paradigm is quite fragile" but all this change seems to do is rather then having a local variable track the error state (in the form of error code) you moved it to marking it in the context. I'm not really sure why that's less fragile ;)

it does make the code a big more clear though so seems reasonable

cmd/debos/debos.go
```diff
@@ -50,37 +50,36 @@ func warnLocalhost(variable string, value string) {
 		    Consider using an address that is valid on your network.`
 
 	if strings.Contains(value, "localhost") ||
-	   strings.Contains(value, "127.0.0.1") ||
-	   strings.Contains(value, "::1") {
+		strings.Contains(value, "127.0.0.1") ||
+		strings.Contains(value, "::1") {
```
sjoerdsimons, line 54: these don't look like typo fixes

cmd/debos/debos.go
```diff
@@ -50,37 +50,36 @@ func warnLocalhost(variable string, value string) {
...
-		ScratchSize   string            `long:"scratchsize" description:"Size of disk backed scratch space"`
-		CPUs          int               `short:"c" long:"cpus" description:"Number of CPUs to use for build VM (default: 2)"`
-		Memory        string            `short:"m" long:"memory" description:"Amount of memory for build VM (default: 2048MB)"`
-		ShowBoot      bool              `long:"show-boot" description:"Show boot/console messages from the fake machine"`
-		EnvironVars   map[string]string `short:"e" long:"environ-var" description:"Environment variables (use -e VARIABLE:VALUE syntax)"`
-		Verbose       bool              `short:"v" long:"verbose" description:"Verbose output"`
-		PrintRecipe   bool              `long:"print-recipe" description:"Print final recipe"`
-		DryRun        bool              `long:"dry-run" description:"Compose final recipe to build but without any real work started"`
-		DisableFakeMachine bool         `long:"disable-fakemachine" description:"Do not use fakemachine."`
+		Backend            string            `short:"b" long:"fakemachine-backend" description:"Fakemachine backend to use" default:"auto"`
```
sjoerdsimons, line 62: more spurious changes that aren't typo fixes

sjoerdsimons (changes requested):

sjoerdsimons:
@obbardc any plans to move this forward ; last commetns were quite small (though would be nice to add some test for exist status ofcourse)

sjoerdsimons:
Closing as it's been superseded by #555

PR #304, opened 2022-01-12
actions/overlay: Check source exists during recipe verification

obbardc:
If an overlay coming from the recipe directory doesn't
exist, the recipe exits during execution. Let's check
early if the overlay exists on the host filesystem.

### DIFF
--- a/actions/overlay_action.go
+++ b/actions/overlay_action.go
@@ -26,6 +26,7 @@ package actions
 
 import (
 	"fmt"
+	"os"
 	"path"
 
 	"github.com/go-debos/debos"
@@ -42,6 +43,13 @@ func (overlay *OverlayAction) Verify(context *debos.DebosContext) error {
 	if _, err := debos.RestrictedPath(context.Rootdir, overlay.Destination); err != nil {
 		return err
 	}
+
+	if len(overlay.Origin) == 0 {
+		if _, err := os.Stat(overlay.Source); err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
 


sjoerdsimons:
bors try

actions/overlay_action.go
```diff
@@ -42,6 +43,13 @@ func (overlay *OverlayAction) Verify(context *debos.DebosContext) error {
 	if _, err := debos.RestrictedPath(context.Rootdir, overlay.Destination); err != nil {
 		return err
 	}
+
+	if len(overlay.Origin) == 0 {
+		if _, err := os.Stat(overlay.Source); err != nil {
```
sjoerdsimons, line 48: the default origin for overlays is the recipedir, you need to take that into account otherwise it will fail when the recipe is not in the cwd of debos. As the new tests nicely show ;)

sjoerdsimons (changes requested):

tests/overlay-non-existent-destination/overlay-non-existent-destination.yaml
```diff
@@ -1,6 +1,12 @@
 architecture: amd64
 
 actions:
+  # This overlay action is expected to error out here because the source
+  # doesn't exist in the recipe directory.
```
sjoerdsimons: so now the next part is no longer tested because this one already errored out right or am i missing something?

tests/overlay-non-existent-destination/overlay-non-existent-destination.yaml
sjoerdsimons, line 1: would it make sense to do the rename in the same commit where you changed it ? also this seems to only delete not add the new name?

sjoerdsimons (changes requested):

tests/overlay-non-existent-destination/overlay-non-existent-destination.yaml
obbardc: split it into two tests now, and the tests actually run now (they didn't before).....

tests/overlay-non-existent-destination/overlay-non-existent-destination.yaml
obbardc: done

obbardc:
@sjoerdsimons I've rebased this PR on main, fixed the merge issues & updated the commit messages to be a bit more sane.

PR #305, opened 2022-01-25
Warn on execution with root user

sefricke:
Debos shouldn't be executed as root user, throw a warning if the user
attempts to do it anyway.

Signed-off-by: Sebastian Fricke <sebastian.fricke@collabora.com>

### DIFF
--- a/cmd/debos/debos.go
+++ b/cmd/debos/debos.go
@@ -88,6 +88,10 @@ func main() {
 		"no_proxy",
 	}
 
+	if os.Getuid() == 0 {
+		log.Printf("Warning, you should not execute debos as root!")
+	}
+
 	var exitcode int = 0
 	// Allow to run all deferred calls prior to os.Exit()
 	defer func() {


cmd/debos/debos.go
```diff
@@ -88,6 +88,10 @@ func main() {
 		"no_proxy",
 	}
 
+	if os.Getuid() == 0 {
+		log.Printf("Warning, you should not execute debos as root!")
```
sjoerdsimons, line 92: Maybe; but then a lot of the debos runs in CI run as root as well (though obviously quite containerized).. So it's not that black and white

obbardc:
It's my understanding that debos may still be useful to be run run as root, in the case that it is running without fakemachine?

Perhaps it would be better to perform this check in go-debos/fakemachine instead?

cmd/debos/debos.go
sjoerdsimons: or maybe to put it better; in CI things often run as uid 0 even it not being the "systems" root

PR #307, opened 2022-02-09
actions/apt: allow downloading packages without installing them

a-wai:
It can sometimes be useful to only download packages (along with their
dependencies) and have them processed at a later point in time. This
patch allows making use of the `--download-only` apt option to that
effect.

Signed-off-by: Arnaud Ferraris <arnaud.ferraris@collabora.com>

### DIFF
--- a/actions/apt_action.go
+++ b/actions/apt_action.go
@@ -5,6 +5,7 @@ Install packages and their dependencies to the target rootfs with 'apt'.
 
 Yaml syntax:
  - action: apt
+   downloadonly: bool
    recommends: bool
    unauthenticated: bool
    update: bool
@@ -17,6 +18,8 @@ Mandatory properties:
 - packages -- list of packages to install
 
 Optional properties:
+
+- downloadonly -- boolean indicating if packages should be downloaded but not installed
 
 - recommends -- boolean indicating if suggested packages will be installed
 
@@ -32,6 +35,7 @@ import (
 
 type AptAction struct {
 	debos.BaseAction `yaml:",inline"`
+	Downloadonly     bool
 	Recommends       bool
 	Unauthenticated  bool
 	Update           bool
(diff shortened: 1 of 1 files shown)


actions/apt_action.go
```diff
@@ -72,9 +80,15 @@ func (apt *AptAction) Run(context *debos.DebosContext) error {
...
 	}
-	err = c.Run("apt", "apt-get", "clean")
-	if err != nil {
-		return err
+	/* 
+	 * Keep packages for further processing if they weren't
+	 * installed immediately
+	 */
+	if !apt.Downloadonly {
+		err = c.Run("apt", "apt-get", "clean")
```
obbardc, line 88: what if another apt action is run after with downloadonly set to false ? It will run apt-get clean ?

obbardc (changes requested):

actions/apt_action.go
a-wai: Indeed, the way I imagined things:
- run apt with `downloadonly`
- immediately copy packages elsewhere
- you're free to do whatever you want afterwards

This will probably cause multiple issues along the way and is bound to confuse users, so I'll rework this so the "copy packages elsewhere" bit is also handled in the same action.

Any suggestion about the options in this case? Would a single `downloadonly: <path>` option work, or would you prefer both `downloadonly: true` and `downloadto: <path>`?

borkd: `downloadonly: <path>` seems more straightforward

sjoerdsimons:
fwiw this seems rare enough that i'm not really sure to do it in a debos action.. and as i'ts quite stale i'll close this for now; Though feel free to open the discussion ofcourse if you'd still like to see this move forward :)

a-wai:
> fwiw this seems rare enough that i'm not really sure to do it in a debos action.. and as i'ts quite stale i'll close this for now

No objection, I don't have a clear use-case for that anymore, I should've closed it a while ago...

PR #315, opened 2022-04-13
Use go modules

obbardc:


sjoerdsimons:
@obbardc draft but you'd like a rview ? that's confusing :)

go.mod
```diff
@@ -0,0 +1,20 @@
+module github.com/go-debos/debos
+
+go 1.15
+
+require (
+	github.com/docker/go-units v0.4.0
+	github.com/go-debos/fakemachine v0.0.0-20220412175549-8719c1b0b5bb
```
sjoerdsimons, line 7: time to properly tag a fakemachine release?

obbardc:
> @obbardc draft but you'd like a rview ? that's confusing :)

Just to see early if there is anything glaringly wrong :-)

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
```
sjoerdsimons, line 1: Can you split out this commit in a seperate MR

go.mod
```diff
@@ -0,0 +1,20 @@
+module github.com/go-debos/debos
```
sjoerdsimons, line 1: This can be merge seperately from introducing better tsting so pleas split it out

go.mod
obbardc: not quite; but i think in april we will do so ;-)

PR #318, opened 2022-04-14
ci: Run tests in test containers

obbardc:
Like fakemachine, run the tests inside specific containers such that the environments are supported and actual testing may be performed.

see https://github.com/go-debos/fakemachine/commit/368366c63456ca416c6cd1a97842a0f75260b666


sjoerdsimons:
@obbardc with the various changs that got merged i guess this one can be closed now?

PR #319, opened 2022-04-14
Build docker container in GitHub actions

obbardc:
This is a continuation of work done by @eds-collabora in !275

This replaces the old, simpler pipeline with a three phase process:

- First, build the image and cache it using docker buildx.
- Second, run all the tests in parallel, restoring the image from the cache.
- Thirdly, if the tests pass:
  - if this is a push to the main branch, push to DockerHub.
  - push to GitHub Container registry (PS: will push to a user's own fork).

This uses Buildkit caching aggressively, and will make use of the entire 5GiB allocation of cache space that GitHub provides over time.

It requires the following additional repository secrets:

- DOCKERHUB_USERNAME: the username to login as on DockerHub (e.g. go-debos)
- DOCKERHUB_PASSWORD: an access token for the DockerHub repository.

Closes: #275


obbardc:
@sjoerdsimons I think this is the most important thing to get right next, since Dockerhub builds are disabled we currently are not be able to build any new images.

.github/workflows/build-test.yaml
```diff
@@ -1,29 +1,246 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Extract Docker metadata
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: ${{ env.GITHUB_TAG }}
+          tags: |
+            "type=raw,value=latest,enable=${{ github.ref == format('refs/heads/{0}', github.event.repository.default_branch) }}"
```
sjoerdsimons, line 159: you probably want to set:
```
flavor: |
  latest = auto
```
to avoid generating latest on tagging?

obbardc: from what i have seen `docker/metadata-action` is quite fickle and only creates the latest tag for releases which are tagged in git which doesn't work for our workflow; we want any push to the default branch to also be tagged as latest.

I'll do some more testing on a fork though :-)

sjoerdsimons: sorry i meant set `latest = false`

But we should really move to pushing :latest on tags/releases indeed rather then on every push

obbardc: right that's a great way forward then - I will remove this hack then ;-)

.github/workflows/build-test.yaml
```diff
@@ -1,29 +1,246 @@
...
+          context: .
+          push: false
+          tags: debos
+          file: docker/Dockerfile
+          cache-from: type=local,src=/tmp/.build-cache
+          load: true
 
       - name: Run test recipes using UML backend
         run: |
           docker-compose -f docker/recipes-test-uml.yml \
```
sjoerdsimons, line 134: now that no builds are done in docker-hub anymore can we get rid of the docker-compose wrapping? (can be later though)

obbardc: yeah - i also want to move Dockerfile to the root and combine the readme files - added a TODO :-)

sjoerdsimons:
@obbardc  can you rebas ethis; seems it's now impossible to merge

PR #321, opened 2022-04-15
For multiline commands, only use the first line and "..." as the label

andrewshadura:
When a multiline command is used with "run" action, it’s sometimes
difficult to figure out what’s going on, as it’s printed in its entirety
for the each line of its own output. For example:
```yaml
  - action: run
    chroot: false
    command: |
      echo test1
      echo test2
      echo test3
```
Prints this:

    2022/04/15 10:52:46 echo test1
    echo test2
    echo test3
     | test1
    2022/04/15 10:52:46 echo test1
    echo test2
    echo test3
     | test2
    2022/04/15 10:52:46 echo test1
    echo test2
    echo test3
     | test3

For clarity, it’s better to only use the first line:

    2022/04/15 11:03:10 echo test1... | test1
    2022/04/15 11:03:10 echo test1... | test2
    2022/04/15 11:03:10 echo test1... | test3

### DIFF
--- a/actions/run_action.go
+++ b/actions/run_action.go
@@ -51,6 +51,10 @@ import (
 	"github.com/go-debos/debos"
 )
 
+const (
+	maxLabelLength = 40
+)
+
 type RunAction struct {
 	debos.BaseAction `yaml:",inline"`
 	Chroot           bool
@@ -113,11 +117,21 @@ func (run *RunAction) doRun(context debos.DebosContext) error {
 		label = path.Base(run.Script)
 	} else {
 		cmdline = []string{run.Command}
-		commands := strings.Split(run.Command, "\n")
+
+		// Remove leading and trailing spaces and — importantly — newlines
+		// before splitting, so that single-line scripts split into an array
+		// of a single string only.
+		commands := strings.Split(strings.TrimSpace(run.Command), "\n")
 		label = commands[0]
 
-		// Make it clear a multi-line command is being run
-		if len(commands) > 1 {
+		// Make it clear a long or a multi-line command is being run
+		if len(label) > maxLabelLength {
+			label = label[:maxLabelLength]
+
+			label = strings.TrimSpace(label)
+
+			label += "..."
+		} else if len(commands) > 1 {
 			label += "..."
 		}
 	}
--- a/actions/run_action.go
+++ b/actions/run_action.go
@@ -113,7 +113,13 @@ func (run *RunAction) doRun(context debos.DebosContext) error {
 		label = path.Base(run.Script)
 	} else {
 		cmdline = []string{run.Command}
-		label = run.Command
+		commands := strings.Split(run.Command, "\n")
+		label = commands[0]
+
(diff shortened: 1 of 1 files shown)


actions/run_action.go
```diff
@@ -113,7 +113,13 @@ func (run *RunAction) doRun(context debos.DebosContext) error {
 		label = path.Base(run.Script)
 	} else {
 		cmdline = []string{run.Command}
-		label = run.Command
+		commands := strings.Split(run.Command, "\n")
+		label = commands[0]
+
+		// Make it clear a multi-line command is being run
+		if len(commands) > 1 {
+			label += "..."
```
obbardc, line 121: perhaps we should trim the label to something like 40 characters before applying `...` ?

obbardc (changes requested):

actions/run_action.go
andrewshadura: Is there an existing idiom for this elsewhere in the code I can reuse?

obbardc: how about something like (untested)

```
commands := strings.Split(run.Command, "\n")

// Make it clear a multi-line command is being run
let maxLabelLength := 40
if len(commands[0]) > maxLabelLength || len(commands) > 1 {
	label = commands[0][:maxLabelLength]
	label = strings.Trim(label)
	label += "..."
}
```

andrewshadura:
As I understand, slicing a string past its end is an error in Go.

actions/run_action.go
andrewshadura: As I understand, slicing a string past its end is an error in Go.

andrewshadura: also, `let` is a rustism 🙂

andrewshadura: Your suggestion seems okay, but maybe it'd be better to truncate long labels elsewhere, in the log printer?

actions/run_action.go
```diff
@@ -113,11 +117,17 @@ func (run *RunAction) doRun(context debos.DebosContext) error {
 		label = path.Base(run.Script)
 	} else {
 		cmdline = []string{run.Command}
-		commands := strings.Split(run.Command, "\n")
+		commands := strings.Split(strings.TrimSpace(run.Command), "\n")
```
obbardc, line 120: would suggest moving the `strings.TrimSpace` to the next line so the label really is trimmed then ?

andrewshadura: The purpose here was to make sure that LFs at the end of the line don’t increase the number of lines.

andrewshadura: `Split` splits a single LF-terminated line into an array of two strings, the first being the text of the line and the second an empty string.

obbardc: right, that wasn't obvious till you mentioned it, perhaps a comment would be good ?

andrewshadura: Done.

obbardc (approved):

PR #327, opened 2022-04-22
Only push to Dockerhub if run from upstream repository

obbardc:
Seems like the conditional for checking main branch doesn't
work on forks; let's ensure that we only attempt to push to
Dockerhub if we are not in a fork.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -194,7 +194,7 @@ jobs:
       - unit-tests
       - recipe-tests
       - uml-tests
-    if: github.event_name == 'push' && github.ref == format('refs/heads/{0}', github.event.repository.default_branch)
+    if: github.event_name == 'push' && github.repository == 'go-debos/debos'
     runs-on: ubuntu-latest
     permissions:
       contents: read


.github/workflows/ci.yaml
```diff
@@ -194,7 +194,7 @@ jobs:
       - unit-tests
       - recipe-tests
       - uml-tests
-    if: github.event_name == 'push' && github.ref == format('refs/heads/{0}', github.event.repository.default_branch)
+    if: github.event_name == 'push' && github.repository == 'go-debos/debos'
```
sjoerdsimons, line 197: This pushes from all branches in `go-debos/debos` to dockerhub which i don't think was the intention?  potentially this should be an additional check or you could move `DOCKERHUB_TAG` to the repository setup and only  run this job if that exist (so another organisation could plug in their own tag + creds to build their own cntainers )

sjoerdsimons (changes requested):

.github/workflows/ci.yaml
obbardc: done!

PR #334, opened 2022-05-03
image_partition_action: Add support for FS extended options

cazou:
Extended options can be used for specific file system tuning.

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -54,6 +54,7 @@ Yaml syntax for partitions:
 	   start: offset
 	   end: offset
 	   features: list of filesystem features
+	   extendedoptions: list of filesystem extended options
 	   flags: list of flags
 	   fsck: bool
 	   fsuuid: string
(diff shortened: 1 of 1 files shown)


actions/image_partition_action.go
```diff
@@ -178,18 +178,19 @@ import (
...
+	Name             string
+	PartLabel        string
+	PartType         string
+	PartUUID         string
+	Start            string
+	End              string
+	FS               string
+	Flags            []string
+	Features         []string
+	ExtendedOptions  []string
```
obbardc, line 191: missing documentation

actions/image_partition_action.go
```diff
@@ -178,18 +178,19 @@ import (
...
-	PartType  string
-	PartUUID  string
-	Start     string
-	End       string
-	FS        string
-	Flags     []string
-	Features  []string
-	Fsck      bool "fsck"
-	FSUUID    string
+	number           int
```
obbardc, line 181: nitpick as we haven't formatted the code with gofmt yet; but the formatting is a spurious change; we should only change the line which is relevent

obbardc (changes requested):

actions/image_partition_action.go
cazou: So you'd rather have unaligned types for now ? It won't match the rest of the code. Should I add an additional formatting commit ?

cazou:
@obbardc How is this latest version ?

PR #335, opened 2022-05-05
tests: action:raw "offset" field, both bytes and sector

mend0za:
Test functionality of action: raw, using addressing by bytes and by sector.

Test writes 1 sector (512 bytes) pattern to selected offsets, addressing them by byte and by sector,  and then check result image do the testing pattern is written down.


sjoerdsimons:
This need to be added to the workflow to actually get tested in CI

tests/action-raw-offset/test-action-raw-offset.yaml
```diff
@@ -0,0 +1,44 @@
...
+    description: Deploy test block
+    source: overlays/test
+
+  - action: image-partition
+    description: Create {{ $image_test }}
+    imagename: {{ $image_test }}
+    imagesize: {{ sector 128 }}
+    partitiontype: msdos
+
+# By Byte
```
obbardc, line 21: to clean these up, move the `description` to after `action` and remove the comments as they're just a duplicate of the description

mend0za: done

daissi:
Since we now have tests for raw action with f0b70e0bac686db6e415f458a5f792ab6ac39600 I wonder if this PR is still needed. If I am not wrong f0b70e0bac686db6e415f458a5f792ab6ac39600 covers all cases of this PR.

obbardc:
agree

PR #358, opened 2022-08-19
actions: debootstrap: Expose include and exclude arguments

obbardc:
Optionally expose the --include and --exclude debootstrap
arguments to allow packages to be included and excluded
at runtime.

### DIFF
--- a/actions/debootstrap_action.go
+++ b/actions/debootstrap_action.go
@@ -17,6 +17,8 @@ Yaml syntax:
    keyring-file:
    certificate:
    private-key:
+   include: <list of packages>
+   exclude: <list of packages>
 
 Mandatory properties:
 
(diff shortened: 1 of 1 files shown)


sjoerdsimons:
Please document a use-case for this; generally you shouldn't be including extra package  when doing a debootstrap run nor is it very advisable to exclude specific packages

obbardc:
> Please document a use-case for this; generally you shouldn't be including extra package when doing a debootstrap run nor is it very advisable to exclude specific packages

Right, this was really created to expose those properties for power-users.

For my case though, this is an early hack to exclude specific packages in cases where debootstrap force a package to be installed when it may not be available for a downstream (for instance, `usr-is-merged` is forced to be installed on any suite which is not in the standard list, so a Debian/Ubuntu downstream which uses a non-standard suite will break. Debootstrap developers are not overly interested in making it work with downstreams and instead have suggested that we exclude the package from the early debootstrap command).

obbardc:
With #362 this is not required anymore.

Issue #361, opened 2022-09-20
New debootstrap is unstable to bootstrap pre-bookworm debian derivatives


sjoerdsimons:
Debian unstable/testing as well as the latest debian stable release have a patch to the debian-common script which requires 
`usr-is-merged` on anything that's not an older known Debian release..

Which breaks using that debootstrap for any derivative distribution with a different codename, which is rather unfortunate as that's a reasonable common usage.

Without changing debootstrap we have the following options:

* #16 - Making the debootstrap script configurable.. While that would give an escape hatch, it will require using a different 
           script entirely as the offending code is in debian-common
* #127 - Support another bootstrapping tool; That won't fix current recipes unless it's done as a drop-in replacement. But switching
              the bootstrapping tool underneath the user opens a whole other can of worms. And really i'd like to avoid having to
              support multiple tools, even with all its issues debootstrap is still the official bootstrapping tools
* #358 - Expose debootstraps include and exclude options; Which we avoided thusfar as debootstrap really isn't the right location
              to add more packages (use apt actions for that) or to remove package as that's risky as well. When using a minbase 
              variant the latter really shouldn't be needed. Not unsurprising it was only proposed due to this issue. And ofcourse, this 
              still requires users to switch their recipes so it's quite annoying.

TL;DR we can't have nice things; and in particular all these options require users to chagne their recipes which is really not great

obbardc:
One attempted modification to debootstrap was to [only install usr-is-merged if the suite is testing or unstable](https://salsa.debian.org/installer-team/debootstrap/-/merge_requests/73); it wasn't accepted and the developer's suggestion was to exclude `usr-is-merge` by `debootstrap`'s `--exclude` argument.

sjoerdsimons:
Digging into this further with some pointers from @smcv; The current overview is:
* usr-is-merged is a dummy package to say:
 * This system was created in a merged usr form
 * This system expclitly opted out from merged-usr (flag file in etc: `unsupported-skip-usrmerge-conversion`); Which is
   mostly intended for QA reasons.

New debootstrap (stable and onwards) will on codenames that it doesn't recognize as an older debian version always require this package *and* if `--no-merged-usr` is given will create the flag file mentioned above. Ideally with the goal of avoiding the installation of `usrmerge`. Though when testing with debootstrapping sid for the new init-system-helpers (with updated dependencies) it seems that still pulls in both `usrmerge` and `usr-is-merged`.

[...]

PR #362, opened 2022-09-22
fix non-debian builds based on current debian releases

sjoerdsimons:
Work around new debootstrap behaviour by always excluding the
 usr-is-merged package as this allows derivatives based on older debian
 versions to build again. See #361 for more details


obbardc (approved):

obbardc:
looks good to me!

bors r+

Issue #344, opened 2022-05-27
image-partition action takes a long time with udev in debian bookworm & unstable


obbardc:
Upgrading udev from 250 to 251 causes image-partition action to hang on calls to parted.

obbardc:
Seems like the calls to parted just busy-wait for a few minutes and eventually do return.

obbardc:
This appears to just affect parted in Debian, which [is patched to call](https://salsa.debian.org/parted-team/parted/-/blob/master/debian/patches/udevadm-settle.patch) `udevadm-settle` around opening and closing the image. Since we hold a `LOCK_EX` on the image; for the whole time the image-partition `Run` function is called, the `udevadm settle` which the patch calls is always waiting and eventually times out.

We should fix parted eventually; but for now as a workaround we can unlock the image around the parted calls.

PR #368, opened 2022-10-12
image-partition: Fix parted timeouts with udev >251

obbardc:
systemd-udevd version 251 introduced a patch where if a device is acquired with a `flock(LOCK_EX)` then udev doesn't process any events until the device is closed. This is great; as it will stop some race conditions.

In the image-partition action we `flock(LOCK_EX)` the image then unlock the image once we have finished processing the image. So we follow systemd suggestion to lock the device while processing, then let udev process the events after all our processing has finished.

Unfortunately, Debian's parted packaging contains a patch which calls `udevadm settle`, which waits for the device to be unlocked (or times out after a while). Obviously, since we have locked the device, all of our parted calls end up with the extra overhead of the `udevadm settle` timeout.

The proper fix is to remove the `udevadm settle` call from Debian's parted packaging and instead call `flock(LOCK_EX)` instead. With this fix in place, we still need to unlock the device around calls to parted (and potentially other programs which modify block devices which are slowly being modified to call `flock(LOCK_EX)` instead).

See: https://github.com/go-debos/debos/issues/344
See: https://systemd.io/BLOCK_DEVICE_LOCKING/
See: https://salsa.debian.org/parted-team/parted/-/blob/master/debian/patches/udevadm-settle.patch
See: https://github.com/systemd/systemd/commit/5d354e525a56955ae7f68062e283dda85ab07794

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -421,6 +421,14 @@ func (i *ImagePartitionAction) PreNoMachine(context *debos.DebosContext) error {
 	return nil
 }
 
+func lockImageExclusively(imageFD *os.File) error {
+	return syscall.Flock(int(imageFD.Fd()), syscall.LOCK_EX)
+}
+
+func lockImageShared(imageFD *os.File) error {
+	return syscall.Flock(int(imageFD.Fd()), syscall.LOCK_SH)
+}
+
 func (i ImagePartitionAction) Run(context *debos.DebosContext) error {
 	i.LogStart()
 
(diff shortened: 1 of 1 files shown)


actions/image_partition_action.go
```diff
@@ -438,16 +446,18 @@ func (i ImagePartitionAction) Run(context *debos.DebosContext) error {
...
 	command := []string{"parted", "-s", context.Image, "mklabel", i.PartitionType}
 	if len(i.GptGap) > 0 {
 		command = append(command, i.GptGap)
 	}
+	/* Unlock image for parted call; then lock afterwards.
+	 * See https://github.com/go-debos/debos/issues/344#issuecomment-1275835432
+	 */
+	lockImageShared(imageFD)
 	err = debos.Command{}.Run("parted", command...)
+	lockImageExclusively(imageFD)
```
sjoerdsimons, line 460: 2 things; first this code is copied 3 times; almost like it should be a runParted function instead :)

But more importantly i'm not sure if this is the right thing, the whole reason for exclusive locking was to avoid udev trigger the kernel to rescan the partition causing block devices to temporarily disappear. So it's re-opening the window for the old race condition again unfortunately

obbardc: @sjoerdsimons do you have a suggestion on a way I can move this forward?

sjoerdsimons: I think the first thing to do here is to make sure we have a testcase again for the old error case; What is happening is quite well described on the commit and the original issue, but i've been failing to reproduce it last night with the flock introducing commit reverted which is quite annoying.

Once we can reproduce it a potential way to fix this without temporarily dropping flock's (which makes taking it somewhat  useless i think, it just potentialy makes the race window samller) could be to only take the lock *after* all partitioning is done and then have debos validate all partition devices have (re)appeared  before continuing with the formatting pass.  

As this would then block all further udev processing while waiting for the kernel to do all the reprobing and recreation of devices. The one thing to verify for that though would be to ensure the `BLKRRPART` ioctl *only* returns *after* dropping the previous set of partition devices which would mean it's safe to assume that when they reappear the operation has finished. 

Fwiw this area is nasty in that it's a _very_ small race window (which is why it took so long previous to reproduce and work out what was going on); So i want to be really sure we don't re-introduce races as that will come back and bite us for sure :)

PR #370, opened 2022-10-14
godoc: add blank line in front of yaml examples

jakob-tsd:
The yaml examples get somewhat mangled on godoc.org currently.

Example from https://pkg.go.dev/github.com/go-debos/debos/actions@main#hdr-Recipe_syntax :

  action: apt recommends: bool unauthenticated: bool update: bool packages:
  package1
  package2

Adding a blank line in front of the blocks makes the examples render as preformatted text.


obbardc:
Thanks for your patch!

bors r+

PR #371, opened 2022-10-14
docs: RawAction: list possible values for origin

jakob-tsd:
Getting the raw action to work with an actual file was somewhat confusing for me.

Looking at the code, the origin is searched for in context.Origins, which is only set in three places:

 $ git grep Origins
 download_action.go:     context.Origins[d.Name] = originPath
 filesystem_deploy_action.go:    context.Origins["filesystem"] = context.ImageMntDir
 ostree_deploy_action.go:                context.Origins["filesystem"] = context.ImageMntDir

In addition, "recipe" is special-cased in DebosContext.Origin.

List all four options to hopefully make it less confusing for the next person.

### DIFF
--- a/actions/actions_doc.go
+++ b/actions/actions_doc.go
@@ -2,5 +2,16 @@
 
 /*
 Package 'actions' implements 'debos' modules used for OS creation.
+
+The origin property
+
+Several actions have the 'origin' property. Possible values for the
+'origin' property are:
+
+  1) 'recipe' ....... directory the recipe is in
+  2) 'filesystem' ... target filesystem root directory from previous filesystem-deploy action or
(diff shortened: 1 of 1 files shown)


actions/raw_action.go
```diff
@@ -12,7 +12,12 @@ Yaml syntax:
 
 Mandatory properties:
 
-- origin -- reference to named file or directory.
+- origin -- reference to directory or previous action. Possible values:
+
+  1) "recipe" ....... directory the recipe is in
```
obbardc, line 17: this documentation should be moved somewhere generic, since `recipe` is generic we only want to update it once.

jakob-tsd:
Moved to package comment.

PR #372, opened 2022-10-15
image-partition: Minimize the image lock scope

sjoerdsimons:
Newer systemd/udev versions will delay processing events for block devices if another process holds an exclusive file lock, which is meant to improve coordination with e.g. partitioners.

This also means `udevadm settle` will block until it hits a timeout if being called while a lock is held. Which happens with the current debos locking scheme (holding the lock for all paritioning operations) due to Debians parted running `udevadm settle` after committing partition changes.

However the debos current locking will equivalently be problematic when partitioning tools will take over systemds device locking recommendations as they'll want to flock the device themselves.

To fix both current and future issues change the debos locking scheme to *only* hold an exclusive lock for operations using the partition devices (formatting and mounting) to avoid udev triggering partition table rescans and devices disappearing when debos actually wants to use them.

Fixes: #344
See: https://systemd.io/BLOCK_DEVICE_LOCKING/
Signed-off-by: Sjoerd Simons <sjoerd@collabora.com>

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -206,6 +206,23 @@ type Mountpoint struct {
 	part       *Partition
 }
 
+type imageLocker struct {
+	fd *os.File
+}
+
+func lockImage(context *debos.DebosContext) (*imageLocker, error) {
+	fd, err := os.Open(context.Image)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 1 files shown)


obbardc:
works well in all cases locally

bors merge

obbardc (approved):

sjoerdsimons:
bors retry

PR #376, opened 2022-10-30
Allow to set kernel and initrd on machine level

avoidik:
depends on https://github.com/go-debos/fakemachine/pull/119


avoidik:
this will allow us to set custom kernel and initrd on machine level like this

```terminal
$ debos --verbose --show-boot --debug-shell -k /boot/vmlinuz -r initramfs.cpio debos-recipe.yaml
```

cmd/debos/debos.go
```diff
@@ -69,6 +69,8 @@ func main() {
 		CPUs          int               `short:"c" long:"cpus" description:"Number of CPUs to use for build VM (default: 2)"`
 		Memory        string            `short:"m" long:"memory" description:"Amount of memory for build VM (default: 2048MB)"`
 		ShowBoot      bool              `long:"show-boot" description:"Show boot/console messages from the fake machine"`
+		KernelPath    string            `short:"k" long:"kernel" description:"kernel to use"`
+		InitrdPath    string            `short:"r" long:"ram-disk" description:"ramdisk to use"`
```
obbardc, line 73: The initramfs is generated by fakemachine, whats the reason behind setting your own initd?
Also, it may make sense to set the path to the kernel modules

avoidik: in this context I assume that everything is prebaked and fakemachine is using initramfs/initrd as is

avoidik: I think I get it, we still need the path to the kernel modules when only '--kernel' flag is set

avoidik: > whats the reason behind setting your own initd?

the reason is to decouple from the host's kernel and modules when init-modules are missing

avoidik: okay, lets remove it

obbardc (changes requested):
Looks like my review was pending and wasn't ever posted. Sorry about that.

PR #377, opened 2022-10-31
Arch support - aka pacstrap, pacman and tests

evelikov:
This PR adds support for creating pacman based images (be that Arch or alike). Similar to the Debian variant we have two important actions - pacstrap and pacman.
 - pacstrap, requires pacman.conf and mirrorlist files
 - pacman is a simple list of packages, just like apt
 
 This PR supersedes https://github.com/go-debos/debos/pull/276 and https://github.com/go-debos/debos/pull/277, where the most significant changes are within the pacstrap action. In particular, we no longer write the config/mirrorlist files.

Considering that pacman is changing faster than debos, it makes sense to have the files as drop-ins instead of trying to hard-code them within debos itself.

Note: the final commit adds a simple CI file, which is not wired. Tips on how to achieve that would be appreciated.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -125,6 +125,8 @@ jobs:
           - { name: "debian (arm64)", case: "debian", variables: "-t architecture:arm64" }
           - { name: "debian (armhf)", case: "debian", variables: "-t architecture:armhf" }
         include:
+          - backend: { name: "arch", backend: "--fakemachine-backend=qemu" }
+            test: { name: "arch", case: "arch" }
           - backend: { name: "qemu", backend: "--fakemachine-backend=qemu" }
             test: { name: "partitioning", case: "partitioning" }
           - backend: { name: "uml", backend: "--fakemachine-backend=uml" }
--- a/actions/pacman_action.go
+++ b/actions/pacman_action.go
@@ -1,0 +1,39 @@
+/*
+Pacman Action
+
+Install packages and their dependencies to the target rootfs with 'pacman'.
+
+	# Yaml syntax:
+	- action: pacman
+	  packages:
+	    - package1
+	    - package2
+
+Mandatory properties:
+
+- packages -- list of packages to install
+*/
+package actions
+
+import (
+	"github.com/go-debos/debos"
+)
+
+type PacmanAction struct {
+	debos.BaseAction `yaml:",inline"`
+	Packages         []string
+}
+
+func (p *PacmanAction) Run(context *debos.DebosContext) error {
+	p.LogStart()
+
+	pacmanOptions := []string{"pacman", "-Syu", "--noconfirm"}
+	pacmanOptions = append(pacmanOptions, p.Packages...)
+
+	c := debos.NewChrootCommandForContext(*context)
+	if err := c.Run("pacman", pacmanOptions...); err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/actions/pacstrap_action.go
+++ b/actions/pacstrap_action.go
@@ -1,0 +1,133 @@
+/*
+Pacstrap Action
+
+Construct the target rootfs with pacstrap tool.
+
+	# Yaml syntax:
+	- action: pacstrap
+	  config: <in-tree pacman.conf file>
+	  mirror: <in-tree mirrorlist file>
+
+Mandatory properties:
+
+  - config -- the pacman.conf file which will be used through the process
+  - mirror -- the mirrorlist file which will be used through the process
+*/
+package actions
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+
+	"github.com/go-debos/debos"
+	"github.com/go-debos/fakemachine"
+)
+
+type PacstrapAction struct {
+	debos.BaseAction `yaml:",inline"`
+	Config           string `yaml:"config"`
+	Mirror           string `yaml:"mirror"`
+}
+
+func (d *PacstrapAction) listOptionFiles(context *debos.DebosContext) ([]string, error) {
+	files := []string{}
+
+	if d.Config == "" {
+		return nil, fmt.Errorf("No config file set")
+	}
+	d.Config = debos.CleanPathAt(d.Config, context.RecipeDir)
+	files = append(files, d.Config)
+
+	if d.Mirror == "" {
+		return nil, fmt.Errorf("No mirror file set")
+	}
+	d.Mirror = debos.CleanPathAt(d.Mirror, context.RecipeDir)
+	files = append(files, d.Mirror)
+
+	return files, nil
+}
+
+func (d *PacstrapAction) Verify(context *debos.DebosContext) error {
+	files, err := d.listOptionFiles(context)
+	if err != nil {
+		return err
+	}
+
+	// Check if all needed files exists
+	for _, f := range files {
+		if _, err := os.Stat(f); os.IsNotExist(err) {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (d *PacstrapAction) PreNoMachine(context *debos.DebosContext) error {
+	return fmt.Errorf("action requires fakemachine")
+}
+
+func (d *PacstrapAction) PreMachine(context *debos.DebosContext, m *fakemachine.Machine, args *[]string) error {
+	mounts, err := d.listOptionFiles(context)
+	if err != nil {
+		return err
+	}
+
+	// Mount configuration files outside of recipes directory
+	for _, mount := range mounts {
+		m.AddVolume(path.Dir(mount))
+	}
+
+	return nil
+}
+
+func (d *PacstrapAction) Run(context *debos.DebosContext) error {
+	d.LogStart()
+
+	files := map[string]string{
+		"/etc/pacman.conf":         d.Config,
+		"/etc/pacman.d/mirrorlist": d.Mirror,
+	}
+
+	// Copy the config/mirrorlist files
+	for dest, src := range files {
+		if err := os.MkdirAll(path.Dir(dest), 0755); err != nil {
+			return err
+		}
+
+		read, err := ioutil.ReadFile(src)
+		if err != nil {
+			return err
+		}
+
+		if err = ioutil.WriteFile(dest, read, 0644); err != nil {
+			return err
+		}
+	}
+
+	// Setup the local keychain, within the fakemachine instance, since we
+	// don't have access to the host one.
+	// Even if we did, blindly copying it might not be a good idea.
+	cmdline := []string{"pacman-key", "--init"}
+	if err := (debos.Command{}.Run("pacman-key", cmdline...)); err != nil {
+		return fmt.Errorf("Couldn't init pacman keyring: %v", err)
+	}
+
+	// When there's no explicit keyring suite we populate all available
+	cmdline = []string{"pacman-key", "--populate"}
+	if err := (debos.Command{}.Run("pacman-key", cmdline...)); err != nil {
+		return fmt.Errorf("Couldn't populate pacman keyring: %v", err)
+	}
+
+	// Run pacstrap
+	cmdline = []string{"pacstrap", context.Rootdir}
+	if err := (debos.Command{}.Run("pacstrap", cmdline...)); err != nil {
+		log := path.Join(context.Rootdir, "var/log/pacman.log")
+		_ = debos.Command{}.Run("pacstrap.log", "cat", log)
+		return err
+	}
+
+	return nil
+}
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -26,6 +26,22 @@ COPY . $GOPATH/src/github.com/go-debos/debos
 WORKDIR $GOPATH/src/github.com/go-debos/debos/cmd/debos
 RUN go install ./...
 
+# Pull the latest archlinux-keyring, since the one in Debian is outdated
+# https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1026080
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        pkgconf \
+        python3-all \
+        sq \
+        systemd && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN git clone https://gitlab.archlinux.org/archlinux/archlinux-keyring && \
+    cd archlinux-keyring && \
+    git checkout -B latest-release 20221213 && \
+    make build && \
+    make PREFIX=/usr KEYRING_TARGET_DIR=/usr/share/keyrings/ DESTDIR=/arch-keyring install
+
 ### second stage - runner ###
 FROM debian:bullseye-slim as runner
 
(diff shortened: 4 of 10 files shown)


evelikov:
@obbardc feel free to close the superseded PRs - I don't have access.

actions/pacman_action.go
```diff
@@ -0,0 +1,39 @@
...
+
+type PacmanAction struct {
+	debos.BaseAction `yaml:",inline"`
+	Packages         []string
+}
+
+func (p *PacmanAction) Run(context *debos.DebosContext) error {
+	p.LogStart()
+
+	pacmanOptions := []string{"pacman", "-Syu", "--noconfirm"}
```
obbardc, line 30: i think it'd be great to check in Verify if pacman is installed

But this can be a follow-up as I we should do the same check with all actions. Can you open an issue about that ?

actions/pacman_action.go
```diff
@@ -0,0 +1,39 @@
...
+	pacmanOptions = append(pacmanOptions, p.Packages...)
+
+	c := debos.NewChrootCommandForContext(*context)
+	if err := c.Run("Pacman", pacmanOptions...); err != nil {
```
obbardc, line 34:
```suggestion
	if err := c.Run("pacman", pacmanOptions...); err != nil {
```
to follow other actions

actions/pacstrap_action.go
```diff
@@ -0,0 +1,129 @@
...
+	debos.BaseAction `yaml:",inline"`
+	Config           string `yaml:"config"`
+	Mirror           string `yaml:"mirror"`
+}
+
+func (d *PacstrapAction) listOptionFiles(context *debos.DebosContext) ([]string, error) {
+	files := []string{}
+
+	if d.Config == "" {
+		return nil, fmt.Errorf("No config file set, aborting.")
```
obbardc, line 38:
```suggestion
		return nil, fmt.Errorf("No config file set")
```

actions/pacstrap_action.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	d.Config = debos.CleanPathAt(d.Config, context.RecipeDir)
+	files = append(files, d.Config)
+
+	if d.Mirror == "" {
+		return nil, fmt.Errorf("No mirror file set, aborting.")
```
obbardc, line 44:
```suggestion
		return nil, fmt.Errorf("No mirror file set")
```

actions/pacstrap_action.go
```diff
@@ -0,0 +1,129 @@
...
+		if err != nil {
+			return err
+		}
+
+		if err = ioutil.WriteFile(dest, read, 0644); err != nil {
+			return err
+		}
+	}
+
+	// Setup the local keychain, within the fakemachine instance, since we
```
obbardc, line 106: what happens if this is run with `--disable-fakemachine` ? does it stamp on the host file ?

tests/arch/test.yaml
```diff
@@ -0,0 +1,13 @@
+---
```
obbardc, line 1: to hook this up; you can:

1) modify `docker/Dockerfile` to install `pacman-package-manager` and `arch-install-scripts` (better put it in its own commented section a.la `# Install Arch dependencies`)
2) modify [CI definition](https://github.com/go-debos/debos/blob/main/.github/workflows/ci.yaml#L122) to add an arch test e.g:
```
          - { name: "arch", case: "arch" }
```

tests/arch/test.yaml
```diff
@@ -0,0 +1,13 @@
...
+{{- $architecture := or .architecture "amd64"}}
+architecture: {{$architecture}}
```
obbardc, line 3:
```suggestion
architecture: amd64
```
just hardcode it as we only support one arch anyway ?

actions/pacstrap_action.go
```diff
@@ -0,0 +1,129 @@
...
+	if d.Mirror == "" {
+		return nil, fmt.Errorf("No mirror file set, aborting.")
+	}
+	d.Mirror = debos.CleanPathAt(d.Mirror, context.RecipeDir)
+	files = append(files, d.Mirror)
+
+	return files, nil
+}
+
+func (d *PacstrapAction) Verify(context *debos.DebosContext) error {
```
obbardc, line 52: Should we check `context.Architecture==amd64` too ?

actions/pacman_action.go
```diff
@@ -0,0 +1,39 @@
...
+
+- packages -- list of packages to install
+*/
+package actions
+
+import (
+	"github.com/go-debos/debos"
+)
+
+type PacmanAction struct {
```
obbardc, line 22: Should we add `Verify` function to check `context.Architecture==amd64` ?

actions/pacman_action.go
evelikov: Greed - a handful of actions make this assumption. Opened https://github.com/go-debos/debos/issues/379

tests/arch/test.yaml
evelikov: For all the arch specific questions: as it's written there is nothing preventing you from trying other architectures - be that arm, aarch or risc-v. Personally I haven't tested them and instead opted for being consistent with the Debian/debootstrap/apt codebase.

Can change that, but it seems very strange to single-out the Arch/pacstrap/pacman actions.

evelikov: If we are to hard-code things, should we do the same for Debian/debootstrap/apt?

actions/pacstrap_action.go
evelikov: No idea to be honest - `fakemachine` is the only option that has been tried.

Unless this is a blocker, I'm leaning that we can error out if run w/o `fakemachine`. Are you comfortable with that?

obbardc: > Unless this is a blocker, I'm leaning that we can error out if run w/o `fakemachine`. Are you comfortable with that?

You can do that as long as the CI tests pass ;-). We can add non-fakemachine support as a follow-up.

tests/arch/test.yaml
evelikov: https://archlinux32.org/ started since 32bit was dropped from the official distro - supports i486, i686 and pentium4
https://archlinuxarm.org/ has been a thing for 10+ years - supports armv7 and armv8 ... in the past it had armv6 support IIRC

evelikov: There is also WIP risc-v somewhere (sorry cannot find the link).

All of the above should just work with the PR, of course assuming debos/qemu runs on that platform.

actions/pacstrap_action.go
evelikov: At a glance the least invasive approach is to:
 - save the context type within `debos.DebosContext`
 - in the `Verify` hooks for `pacstrap/pacman` will print "Unsupported backend"
 - where the `Run` hooks will silently error out

It does feel a bit hacky, but the alternative is quite CI churn heavy. The current testing has no "known failing" concept. 

Does the above sound reasonable or do you have a particular approach in mind?

tests/arch/test.yaml
evelikov: This doesn't quite work - `pacman-package-manager` is available only in `bookworm` and `sid`. The other package is present across the board.

obbardc: bump the container baseline to bookworm for testing this PR, we can cherry pick that into another PR later on.

actions/pacstrap_action.go
obbardc: I think it'd be great to first answer the question of what could happen to the user's machine if pacman/pacstrap action is run with no fakemachine

but your approach sounds ok; but instead of having to add a separate field to context and running in `Verify` you could just error out in `PreNoMachine` instead ;-)

tests/arch/test.yaml
obbardc: right - thanks for the history. I thought arch was arm64 only.

So, the named architectures here should just build ootb we can just resolve this.

docker/Dockerfile
```diff
@@ -88,6 +88,13 @@ RUN apt-get update && \
         zip && \
     rm -rf /var/lib/apt/lists/*
 
+# Install Arch dependencies
```
obbardc, line 91: Just need to change the container baseline from `bullseye` to `bookworm` - since `pacman-package-manager` is not available in bullseye.

If it works, we can ask the author to backport the package to bullseye-backports and install from there, or just stick with bumping the container baseline. If you can do that in a commit on top, we can always decide that in another PR.

actions/pacstrap_action.go
evelikov: Yes, when run without fakemachine the current pacstrap stage will try and override the system files - pacman.conf and mirrorlist. Assuming debos has enough permissions (run as root or otherwise) it will succeed.

Added an error out in `PreNoMachine` although that is causing the existing CI to choke. Any pointers would be apprecaited.

docker/Dockerfile
evelikov: Bumped - that does cause debian and aperitis tests to fail. Let me know what you'd prefer - omit the bump, fix those tests, other. 

Once the Arch/pacstrap/pacman CI finishes that it - it'll take some time.

actions/pacstrap_action.go
evelikov: As per the CPU architecture discussion above, the check should not exist.

.github/workflows/ci.yaml
```diff
@@ -121,6 +121,7 @@ jobs:
           - { name: "uml", backend: "--fakemachine-backend=uml" }
         test:
           - { name: "recipes", case: "recipes" }
+          - { name: "arch", case: "arch" }
```
obbardc, line 124: if you just want arch tests to run in a fakemachine and not in the `--disable-fakemachine` case, you can do this instead:

```
include:
  - backend: { name: "arch", backend: "--fakemachine-backend=uml" }
```

docker/Dockerfile
evelikov: Seems like the bump to bookworm is breaking DNS/name resolution for all distros - Debian, Aperitis and Arch.

obbardc: sorry I forgot to mention - in that update you will also need to install the "systemd-resolved" package

actions/pacstrap_action.go
evelikov: Fixed with the "include" snippet that you're shared. Thank you :bowing_man:

docker/Dockerfile
evelikov: Fixed thanks.

evelikov:
Split the bookworm uprev to https://github.com/go-debos/debos/pull/382

The failures are due to json formatting variance. At a glance the proper solution is to call `jq .` on both files prior to feeding them into `diff`. Will open a separate PR in a second

obbardc:
@evelikov We're tracking pulling `pacman-package-manager` into `bullseye-backports` under https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1023386

evelikov:
Glancing at the Debian bug - seems like the backport is not so straight-forward.

To unblock us, we can build pacman locally in the short term. Then switch to the official package as it becomes available. @obbardc what do you think?

obbardc:
> Glancing at the Debian bug - seems like the backport is not so straight-forward.
> 
> To unblock us, we can build pacman locally in the short term. Then switch to the official package as it becomes available. @obbardc what do you think?

That sounds like a great plan!

evelikov:
Was quite sick last week and my TODO stack is quite high, so I don't think I'll get to it this week.
The upstream discussion seems nearly resolved so, it might be better to revisit this next week/other after.

Will flip this to draft, so it doesn't get merged accidentally - the mirrorlist URLs need fixing.

actions/pacstrap_action.go
```diff
@@ -0,0 +1,133 @@
...
+		if err = ioutil.WriteFile(dest, read, 0644); err != nil {
+			return err
+		}
+	}
+
+	// Setup the local keychain, within the fakemachine instance, since we
+	// don't have access to the host one.
+	// Even if we did, blindly copying it might not be a good idea.
+	cmdline := []string{"pacman-key", "--init"}
+	if err := (debos.Command{}.Run("Pacman-key", cmdline...)); err != nil {
```
obbardc, line 114: nit: pacman-key lowercase

actions/pacstrap_action.go
```diff
@@ -0,0 +1,133 @@
...
+		return fmt.Errorf("Couldn't init pacman keyring: %v", err)
+	}
+
+	// When there's no explicit keyring suite we populate all available
+	cmdline = []string{"pacman-key", "--populate"}
+	if err := (debos.Command{}.Run("Pacman-key", cmdline...)); err != nil {
```
obbardc, line 120: nit: pacman-key lowercase

actions/pacstrap_action.go
```diff
@@ -0,0 +1,133 @@
...
+		return fmt.Errorf("Couldn't populate pacman keyring: %v", err)
+	}
+
+	// Run pacstrap
+	cmdline = []string{"pacstrap", context.Rootdir}
+	if err := (debos.Command{}.Run("Pacstrap", cmdline...)); err != nil {
```
obbardc, line 126: nit: pacstrap lowercase

obbardc (changes requested):

actions/pacstrap_action.go
obbardc: this is the last nit remaining @evelikov :p

evelikov: Sorry about that. Think I've got them all covered now :crossed_fingers:

evelikov:
Quick look at the CI shows something quite odd:

We enable `bullseye-backports` before we do anything Arch related.

Yet `arch-install-scripts v23-3` from `bullseye` gets installed which [is missing pacstrap](https://packages.debian.org/bullseye/all/arch-install-scripts/filelist). Instead I would expect the one from backports to be installed (v28-1~bpo11+1) which [has the required tool](https://packages.debian.org/bullseye-backports/all/arch-install-scripts/filelist).

@obbardc any ideas how can we convince `apt` to use the backports variants for the whole stack?

obbardc:
> Quick look at the CI shows something quite odd:
> 
> We enable `bullseye-backports` before we do anything Arch related.
> 
> Yet `arch-install-scripts v23-3` from `bullseye` gets installed which [is missing pacstrap](https://packages.debian.org/bullseye/all/arch-install-scripts/filelist). Instead I would expect the one from backports to be installed (v28-1~bpo11+1) which [has the required tool](https://packages.debian.org/bullseye-backports/all/arch-install-scripts/filelist).
> 
> @obbardc any ideas how can we convince `apt` to use the backports variants for the whole stack?

Ah, that explains things :-)! Please try `apt-get install -t bullseye-backports <package_name>`

evelikov:
The issue we're seeing here is that one of the keys have expired and thus validation fails.

The Debian `archlinux-keyring` package is tad old - upstream has a new release to fix that.

obbardc:
> The issue we're seeing here is that one of the keys have expired and thus validation fails.
> 
> The Debian `archlinux-keyring` package is tad old - upstream has a new release to fix that.

I opened an upstream ticket (note that you could also do the same in future if other packages appear outdated)
It's tracked at: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1026080
For now to move this patch forward, since the backport may take weeks/months, can you somehow install the keys manually and put the bug link in the source code ?

evelikov:
Sure thing - pulled the keyring installation locally. I'm wondering if it makes sense to actually keep it all together, since there can be quite notable delays until an updated package lands in backports.

If I understand correctly, there is a 1-2 week delay between the new version migrating from unstable to testing. Only after it lands in testing it's allowed in backports. During which time the CI will be broken and as result bors won't be able to merge.

obbardc:
Nice work!

bors r+

evelikov:
Now that is a much appreciated Xmas/NYE present. Thanks o/

PR #380, opened 2022-11-02
action: move and inline LogStart() further up

evelikov:
Currently each action starts with LogStart(), adding some annoying duplication. In addition we have a LogStart(), whereas there's no end/stop equivalent.

Instead add the on-line log.Print() wihtin the debos action loop, just prior to calling Run(). The same file deals with all the other pretty logging.

NOTE: this conflicts trivially with https://github.com/go-debos/debos/pull/377

### DIFF
--- a/action.go
+++ b/action.go
@@ -3,7 +3,6 @@ package debos
 import (
 	"bytes"
 	"github.com/go-debos/fakemachine"
-	"log"
 )
 
 type DebosState int
@@ -73,10 +72,6 @@ type Action interface {
 type BaseAction struct {
 	Action      string
 	Description string
-}
-
-func (b *BaseAction) LogStart() {
-	log.Printf("==== %s ====\n", b)
 }
 
 func (b *BaseAction) Verify(context *DebosContext) error { return nil }
--- a/actions/apt_action.go
+++ b/actions/apt_action.go
@@ -44,8 +44,6 @@ func NewAptAction() *AptAction {
 }
 
 func (apt *AptAction) Run(context *debos.DebosContext) error {
-	apt.LogStart()
-
 	aptConfig := []string{}
 
 	/* Don't show progress update percentages */
--- a/cmd/debos/debos.go
+++ b/cmd/debos/debos.go
@@ -28,6 +28,7 @@ func checkError(context *debos.DebosContext, err error, a debos.Action, stage st
 
 func do_run(r actions.Recipe, context *debos.DebosContext) int {
 	for _, a := range r.Actions {
+		log.Printf("==== %s ====\n", a)
 		err := a.Run(context)
 
 		// This does not stop the call of stacked Cleanup methods for other Actions
--- a/actions/debootstrap_action.go
+++ b/actions/debootstrap_action.go
@@ -181,7 +181,6 @@ func (d *DebootstrapAction) isLikelyOldSuite() bool {
 }
 
 func (d *DebootstrapAction) Run(context *debos.DebosContext) error {
-	d.LogStart()
 	cmdline := []string{"debootstrap"}
 
 	if d.MergedUsr {
--- a/actions/download_action.go
+++ b/actions/download_action.go
@@ -125,7 +125,6 @@ func (d *DownloadAction) Verify(context *debos.DebosContext) error {
 
 func (d *DownloadAction) Run(context *debos.DebosContext) error {
 	var filename string
-	d.LogStart()
 
 	url, err := d.validateUrl()
 	if err != nil {
--- a/actions/filesystem_deploy_action.go
+++ b/actions/filesystem_deploy_action.go
@@ -117,7 +117,6 @@ func (fd *FilesystemDeployAction) setupKernelCmdline(context *debos.DebosContext
 }
 
 func (fd *FilesystemDeployAction) Run(context *debos.DebosContext) error {
-	fd.LogStart()
 	/* Copying files is actually silly hafd, one has to keep permissions, ACL's
 	 * extended attribute, misc, other. Leave it to cp...
 	 */
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -444,8 +444,6 @@ func (i *ImagePartitionAction) PreNoMachine(context *debos.DebosContext) error {
 }
 
 func (i ImagePartitionAction) Run(context *debos.DebosContext) error {
-	i.LogStart()
-
 	/* On certain disk device events udev will call the BLKRRPART ioctl to
 	 * re-read the partition table. This will cause the partition devices
 	 * (e.g. vda3) to temporarily disappear while the rescanning happens.
--- a/actions/ostree_commit_action.go
+++ b/actions/ostree_commit_action.go
@@ -78,7 +78,6 @@ func emptyDir(dir string) {
 }
 
 func (ot *OstreeCommitAction) Run(context *debos.DebosContext) error {
-	ot.LogStart()
 	repoPath := path.Join(context.Artifactdir, ot.Repository)
 
 	emptyDir(path.Join(context.Rootdir, "dev"))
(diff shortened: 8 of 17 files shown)


cmd/debos/debos.go
```diff
@@ -28,6 +28,7 @@ func checkError(context *debos.DebosContext, err error, a debos.Action, stage st
 
 func do_run(r actions.Recipe, context *debos.DebosContext) int {
 	for _, a := range r.Actions {
+		log.Printf("==== %s ====\n", a)
```
obbardc, line 31: does this also print when running nest recipes with the `recipe` action? :p

evelikov: Oops fixed, or at least `tests/recipes/test.yaml` seems to be happy.

action.go
```diff
@@ -75,10 +74,6 @@ type BaseAction struct {
 	Description string
 }
 
-func (b *BaseAction) LogStart() {
```
obbardc, line 78: to remove duplicates of Printf, I think it'd be best to keep this function and call it from the the main function in the two places ?

action.go
```diff
@@ -3,7 +3,6 @@ package debos
 import (
 	"bytes"
 	"github.com/go-debos/fakemachine"
-	"log"
```
obbardc, line 6: keep me plz

actions/apt_action.go
```diff
@@ -44,7 +44,6 @@ func NewAptAction() *AptAction {
 }
 
 func (apt *AptAction) Run(context *debos.DebosContext) error {
```
obbardc, line 46: missing for arch actions ;-)

action.go
evelikov: Not sure I follow: are you saying that using two "log.Printf(...)" lines is too much repetition?

In general it make sense to keep some form of abstraction assuming a) it's more than one-liner b) a specific format is used and/or c) there are symmetric LogEnd/LogFoo helpers. Without any of these the function seems like a unnecessary distraction IMHO.

action.go
evelikov: If I keep it (yet remove LogStart as per above) the build will fail, since it's unused. Am I missing something?

actions/apt_action.go
evelikov: Correct - there is a big NOTE that aims to highlight that

action.go
obbardc: Yeah I think two separate instances is too much repetition for something like this where the same behaviour is in two places - if the style of one printf call is modified, the other call (just for nested recipes) may get missed. I'd prefer to move this function to `cmd/main/debos.go`.

action.go
obbardc: This refers to https://github.com/go-debos/debos/pull/380#discussion_r1060542966

action.go
evelikov: Shouldn't there be a CI test to catch such changes/issues? Would it make sense to add "LogEnd" in that case?

action.go
evelikov: I suspect you meant the import should be kept alongside the helper, right?

The linked suggests moving the helper to another file, yet this comment says the import should be kept as-is, which will lead to build failure.

obbardc: yes, lets just resolve this one instead...

action.go
obbardc: My opinion is no to both of those questions.

actions/apt_action.go
obbardc: right I only looked at the commit description & contents.

action.go
evelikov: Your suggestion does not seem to be doable. For example, applying the following simple diff fails to build as below.

```diff
index 9f390db..0c2e426 100644
--- a/actions/pacman_action.go
+++ b/actions/pacman_action.go
@@ -25,8 +25,6 @@ type PacmanAction struct {
 }

 func (p *PacmanAction) Run(context *debos.DebosContext) error {
-       p.LogStart()
-
        pacmanOptions := []string{"pacman", "-Syu", "--noconfirm"}
        pacmanOptions = append(pacmanOptions, p.Packages...)

diff --git a/actions/pacstrap_action.go b/actions/pacstrap_action.go
index 25c1624..e0eb9d2 100644
--- a/actions/pacstrap_action.go
+++ b/actions/pacstrap_action.go
@@ -84,8 +84,6 @@ func (d *PacstrapAction) PreMachine(context *debos.DebosContext, m *fakemachine.
 }

 func (d *PacstrapAction) Run(context *debos.DebosContext) error {
-       d.LogStart()
-
        files := map[string]string{
                "/etc/pacman.conf":         d.Config,
                "/etc/pacman.d/mirrorlist": d.Mirror,
diff --git a/cmd/debos/debos.go b/cmd/debos/debos.go
index 951cd2e..513b3b7 100644
--- a/cmd/debos/debos.go
+++ b/cmd/debos/debos.go
@@ -15,6 +15,10 @@ import (
        "github.com/jessevdk/go-flags"
 )

+func (b *debos.BaseAction) LogStart() {
+       log.Printf("==== %s ====\n", b)
+}
+
 func checkError(context *debos.DebosContext, err error, a debos.Action, stage string) int {
        if err == nil {
                return 0
```

```shell
$ go build -v ./...
github.com/go-debos/debos/cmd/debos
# github.com/go-debos/debos/cmd/debos
cmd/debos/debos.go:18:7: cannot define new methods on non-local type debos.BaseAction
````

Comment (obbardc) [reply action.go:78]: > Your suggestion does not seem to be doable. For example, applying the following simple diff fails to build as below.

Maybe you missed my edit to the suggestion:
> I'd prefer to keep this function and call it from `cmd/main/debos.go` in the two places where it is used.

Comment (obbardc) [reply action.go:78]: @evelikov I implemented the above suggestion at https://github.com/go-debos/debos/pull/417 and will close this PR. Can you review it?

**Comment by evelikov** [comment]
> @evelikov Can you please rebase onto latest main? I tried myself but cannot push to your fork.

Do you have maintainer access? The `Allow edits and access to secrets by maintainers` is ticked.

Either way - rebase pushed. Tests are failing badly here (even against main), so it'll leave it to Github to check if I haven't broken anything.

**Comment by obbardc** [comment]
> > @evelikov Can you please rebase onto latest main? I tried myself but cannot push to your fork.
> Do you have maintainer access? The `Allow edits and access to secrets by maintainers` is ticked.

Yes - as spoken about many times I am a maintainer of this project ;-)

> Either way - rebase pushed. Tests are failing badly here (even against main), so it'll leave it to Github to check if I haven't broken anything.
Thanks - curious to see which tests pass as GitHub says `All checks have passed - 23 successful and 2 skipped checks`

**Review by obbardc — APPROVED** [review]

**Comment by obbardc** [comment]
> In case you're curious here's the go test log

Right, seems like you're just missing `dpkg` locally and one test is failing. I was confused at this originally, it seems like the tests fail as there is error output but that's just misleading.
PS: I'd be willing to skip that test if that file doesn't exist in $path ;-)

**Comment by obbardc** [comment]
@evelikov This needs rebasing.

**Comment by obbardc** [comment]
@evelikov I have rebased your branch and solved the conflicts. I cannot access your fork. Here's the git log:

```
$ git push evelikov HEAD --force
Enumerating objects: 352, done.
Counting objects: 100% (309/309), done.
Delta compression using up to 16 threads
Compressing objects: 100% (200/200), done.
Writing objects: 100% (252/252), 47.69 KiB | 3.97 MiB/s, done.
Total 252 (delta 134), reused 71 (delta 30), pack-reused 0
remote: Resolving deltas: 100% (134/134), completed with 28 local objects.
To github.com:evelikov/debos.git
 ! [remote rejected] HEAD -> evelikov/remove-logStart (permission denied)
error: failed to push some refs to 'github.com:evelikov/debos.git'
```

obbardc:
FTR, `git push -u evelikov HEAD:remove-logStart --force` worked.

PR #381, opened 2022-11-02
image-partition: add unlock() in the error paths

evelikov:
Earlier commit added image locking around the partition handling. Yet we forgot to unlock in the error path - fix that.

Fixes: 1002e0a ("image-partition: Minimize the image lock scope")

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -531,6 +531,7 @@ func (i ImagePartitionAction) Run(context *debos.DebosContext) error {
 		if err != nil {
 			return err
 		}
+		defer lock.unlock()
 
 		err = i.formatPartition(p, *context)
 		if err != nil {
@@ -566,6 +567,8 @@ func (i ImagePartitionAction) Run(context *debos.DebosContext) error {
 	if err != nil {
 		return err
 	}
+	defer lock.unlock()
+
 	for _, m := range i.Mountpoints {
 		dev := i.getPartitionDevice(m.part.number, *context)
 		mntpath := path.Join(context.ImageMntDir, m.Mountpoint)


actions/image_partition_action.go
```diff
@@ -572,6 +573,7 @@ func (i ImagePartitionAction) Run(context *debos.DebosContext) error {
 		os.MkdirAll(mntpath, 0755)
 		err = syscall.Mount(dev, mntpath, m.part.FS, 0, "")
 		if err != nil {
+			lock.unlock()
```
obbardc, line 576: wouldn't it be better to add a deferred call to unlock when lockImage is called ?

evelikov: Unless I've missing something, we have a few options when using the deferred call:
 - leave the existing `unlock` - thus, the in deferred call we'll end-up calling `man 3 close` with an fd that's already closed - quite bad but not critical
 - remove the existing `unlock` - increasing the contention, aka making the process slower than it needs to be
 - remove the deferred call (somehow?) so we don't call `man 3 close` twice (as outlined above)

Since none of the above seems particularly appealing, I went with the simple 2-liner solution. Can change that thought - let me know your preference.

obbardc: I'd prefer the first option - but rather than calling `close` twice, we could track the state of if the file is opened/closed in `imageLocker` ? Maybe in go calling `.Close()` on an already closed fd already has some magic to prevent a use-after-free type situation ?

The current solution here works well - but I am worried about future code additions which could return before calling `lock.unlock()`.

evelikov: A brief diffing around shows:

The `File.Close()` will properly invalidate the internal file-descriptor reference. Thus a second call to `File.Close()` will _not_ call `man 3 close`, instead if will return an error `file already closed`. Hence using `defer` is the proper way to go.

PR #382, opened 2022-11-02
docker: Bump container baseline to bookworm

evelikov:
Some actions (e.g. pacman, pacstrap) require software which is not
present in the current stable Debian release; bump the container
baseline to the bookworm testing release to ensure we can access these
dependencies.

Signed-off-by: Emil Velikov <emil.velikov@collabora.com>

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,7 +3,7 @@ ARG DEBIAN_FRONTEND=noninteractive
 ARG GOPATH=/usr/local/go
 
 ### first stage - builder ###
-FROM debian:bullseye-slim as builder
+FROM debian:bookworm-slim as builder
 
 ARG DEBIAN_FRONTEND
 ARG GOPATH
@@ -27,7 +27,7 @@ WORKDIR $GOPATH/src/github.com/go-debos/debos/cmd/debos
 RUN go install ./...
 
 ### second stage - runner ###
-FROM debian:bullseye-slim as runner
+FROM debian:bookworm-slim as runner
 
 ARG DEBIAN_FRONTEND
 ARG GOPATH
@@ -80,6 +80,7 @@ RUN apt-get update && \
         rsync \
         systemd \
         systemd-container \
+        systemd-resolved \
         u-boot-tools \
         unzip \
         user-mode-linux \


docker/Dockerfile
```diff
@@ -27,7 +27,7 @@ WORKDIR $GOPATH/src/github.com/go-debos/debos/cmd/debos
 RUN go install ./...
```
obbardc, line 27: need to make the first stage bookworm too - otherwise we're building on one distro and running on another ;-)

docker/Dockerfile
```diff
@@ -80,6 +80,7 @@ RUN apt-get update && \
         rsync \
         systemd \
         systemd-container \
+        systemd-resolved \
         u-boot-tools \
         unzip \
```
obbardc, line 85: commit message can be something like

```
Bump container baseline to bookworm

Some actions (e.g. pacman, pacstrap) require software which is not present in the current stable Debian release; bump the container baseline to the bookworm testing release to ensure we can access these dependencies.
```

evelikov:
Rebased on top of main to include CI/json fix (aka https://github.com/go-debos/debos/pull/383)

sjoerdsimons:
I don't think we should make our default container be based on Debian testing 

From a search i assume the dependencies you're looking for is the `arch-install-scripts` package for pacstrap (which is available in backports); In which case that could be installed in the docker container as well without bumping debian version.

pacman is `pacman-package-manager` in Debian, which at the moment only available in testing indeed, will that be needed on the root filesystem as well?  For debian the bootstrapping doesn't need apt, that's only ever ran inside the bootstrapped filesystem, but i really don't know how pacstrap/pacman work together in that regard.

If pacman-package-manager is indeed needed in the container; I'd suggest creating a seperate build (which you can probably do with the same Dockerfile) for a  `debos-testing` or `debos-bookworm` or whatever container, which would be beneficial for our overall testing anyway. And in that container we can then install the bits needed for the arch tooling as well.

fwiw i actually had expected we would get an arch based docker container to use the arch tools, but i guess from this that's not how things are currently used? (I'm fine either way fwiw)

obbardc:
> I don't think we should make our default container be based on Debian testing

OK, the container should track Debian stable then.

> From a search i assume the dependencies you're looking for is the `arch-install-scripts` package for pacstrap (which is available in backports); In which case that could be installed in the docker container as well without bumping debian version.
> 
> pacman is `pacman-package-manager` in Debian, which at the moment only available in testing indeed, will that be needed on the root filesystem as well? For debian the bootstrapping doesn't need apt, that's only ever ran inside the bootstrapped filesystem, but i really don't know how pacstrap/pacman work together in that regard.

pacstrap is just a wrapper around pacman: all of the bootstrapping logic is inside pacman.
I wonder if we get `pacman-package-manager` in bullseye-backports that could move that forward?

> If pacman-package-manager is indeed needed in the container; I'd suggest creating a seperate build (which you can probably do with the same Dockerfile) for a `debos-testing` or `debos-bookworm` or whatever container, which would be beneficial for our overall testing anyway. And in that container we can then install the bits needed for the arch tooling as well.
> 
> fwiw i actually had expected we would get an arch based docker container to use the arch tools, but i guess from this that's not how things are currently used? (I'm fine either way fwiw)

The goal here really was to be able to build "any" downstream image with the default debos container rather than having to use a separate container. In practice that seems to work quite well ;-).

sjoerdsimons:
> > If pacman-package-manager is indeed needed in the container; I'd suggest creating a seperate build (which you can probably do with the same Dockerfile) for a `debos-testing` or `debos-bookworm` or whatever container, which would be beneficial for our overall testing anyway. And in that container we can then install the bits needed for the arch tooling as well.
> > fwiw i actually had expected we would get an arch based docker container to use the arch tools, but i guess from this that's not how things are currently used? (I'm fine either way fwiw)
> 
> The goal here really was to be able to build "any" downstream image with the default debos container rather than having to use a separate container. In practice that seems to work quite well ;-).

That's the ideal indeed; But not at the potential cost of stability of said container especially when it's for new features :) That said orthogonally to this there are benefits of building a container against debian testing as part of our CI loop as well so we can catch regressions in a future Debian release early as well.

So my suggestion would be still, for now, lets build a second container that we test with the arch bits included. Then we can merge all the arch tooling and have a container available for people to build arch systems with even if it's not the default debos container (which really i don't think is a big inconvenience).

Then orthogonally someone could request the pacman maintainer to do a backport as well at which point we can incude the tools in the main container trivially as well; But even then we can keep both containers to improve our testing, so it's not a lost effort anyway

evelikov:
The CI in upstream pacman ensures that it just works with Arch, Fedora and Debian, by churning through it's very extensive test suite.

So I don't know if repeating the pacstrap/pacman testing on many distributions will bring value - not against it though. Mostly concerned about the CI wait times and Github getting angry at us, for burning well over ~2 hours of CPU time per PR.

Above aside: we could keep the current Debian image at bullseye and manually `git close && meson install` pacman. Not pretty, but it will unblock us until the Debian pacman-package-manager maintainer adds it to bullseye-backports.

Overall I don't have strong opinion - whatever works for you guys, I'm happy to give it a try.

obbardc:
> In case it wasn't clear - I'm comfortable with implementing any option, just let me know which one you prefer.

Since we don't wanna bump to bookworm, let's close this one and continue the discussion in #377 :-)

PR #387, opened 2022-12-07
ci: remove deprecated set-output

evelikov:
The option is deprecated and we should be using $GITHUB_OUTPUT instead. See the official announcement for details:

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


evelikov:
The Debian failures are unrelated. Should we have weekly/monthly builds to catch such issues?

PR #390, opened 2023-01-03
Limit old suite workaround

andhe:
The workaround for https://github.com/go-debos/debos/issues/361 that was applied in https://github.com/go-debos/debos/commit/b3c1f76bcc1dbd55fef584b8ddbda33f12733116 breaks recipes for bookworm and newer.

This change makes the workaround less damaging (should atleast work with official debian releases for the forseeable future), while still far from perfect (any derivate distribution that is bookworm-based will fail).

### DIFF
--- a/actions/debootstrap_action.go
+++ b/actions/debootstrap_action.go
@@ -52,6 +52,7 @@ package actions
 import (
 	"fmt"
 	"io"
+	"log"
 	"os"
 	"path"
 	"strings"
@@ -157,6 +158,24 @@ func (d *DebootstrapAction) RunSecondStage(context debos.DebosContext) error {
 	return err
 }
 
+// Guess if suite is something before usr-is-merged was introduced
+func (d *DebootstrapAction) isLikelyOldSuite() bool {
+	switch strings.ToLower(d.Suite) {
+	case "sid", "unstable":
+		return false
+	case "testing":
+		return false
+	case "bookworm":
+		return false
+	case "trixie":
+		return false
+	case "forky":
+		return false
+	default:
+		return true
+	}
+}
+
 func (d *DebootstrapAction) Run(context *debos.DebosContext) error {
 	d.LogStart()
 	cmdline := []string{"debootstrap"}
@@ -203,7 +222,12 @@ func (d *DebootstrapAction) Run(context *debos.DebosContext) error {
 		cmdline = append(cmdline, fmt.Sprintf("--variant=%s", d.Variant))
 	}
 
-	cmdline = append(cmdline, "--exclude=usr-is-merged")
+	// workaround for https://github.com/go-debos/debos/issues/361
+	if d.isLikelyOldSuite() {
+		log.Println("excluding usr-is-merged as package is not in suite")
+		cmdline = append(cmdline, "--exclude=usr-is-merged")
+	}
+
 	cmdline = append(cmdline, d.Suite)
 	cmdline = append(cmdline, context.Rootdir)
 	cmdline = append(cmdline, d.Mirror)


andhe:
Potential further improvements:
* Print a warning when usr-is-merged is excluded! (to give users a hint of what's going on rather than just the obscure `apt install --fix-broken` output.)
* Introduce explicit command line switches to enable/disable the workaround.
* Consider switching the defaults around (assuming bookworm-based releases, allow overriding via command line switch as suggesten in previous bullet point).

WDYT?

actions/debootstrap_action.go
```diff
@@ -157,6 +157,26 @@ func (d *DebootstrapAction) RunSecondStage(context debos.DebosContext) error {
 	return err
 }
 
+func (d *DebootstrapAction) isLikelyOldSuite() bool {
+	// Guess if suite is something before usr-is-merged was introduced
```
obbardc, line 161: Can you move this to the function declaration so that go doc will work? e.g.

```suggestion
// Guess if suite is something before usr-is-merged was introduced
func (d *DebootstrapAction) isLikelyOldSuite() bool {
```

actions/debootstrap_action.go
```diff
@@ -157,6 +157,26 @@ func (d *DebootstrapAction) RunSecondStage(context debos.DebosContext) error {
...
+	switch strings.ToLower(d.Suite) {
+	// these are known to have usr-is-merged.
```
obbardc, line 163: remove comment

actions/debootstrap_action.go
```diff
@@ -157,6 +157,26 @@ func (d *DebootstrapAction) RunSecondStage(context debos.DebosContext) error {
...
+		return false;
+	case "testing":
+		return false;
+	case "bookworm":
+		return false;
+	case "trixie":
+		return false;
+	case "forky":
+		return false;
+	// we can't list all derivate distributions suites, so use default.
```
obbardc, line 174: remove comment

actions/debootstrap_action.go
```diff
@@ -157,6 +157,26 @@ func (d *DebootstrapAction) RunSecondStage(context debos.DebosContext) error {
 	return err
 }
 
+func (d *DebootstrapAction) isLikelyOldSuite() bool {
+	// Guess if suite is something before usr-is-merged was introduced
+	switch strings.ToLower(d.Suite) {
+	// these are known to have usr-is-merged.
+	case "sid", "unstable":
+		return false;
```
obbardc, line 165: remove the `;` from each line as go doesnt use them

obbardc (changes requested):

actions/debootstrap_action.go
```diff
@@ -203,7 +223,11 @@ func (d *DebootstrapAction) Run(context *debos.DebosContext) error {
 		cmdline = append(cmdline, fmt.Sprintf("--variant=%s", d.Variant))
 	}
 
-	cmdline = append(cmdline, "--exclude=usr-is-merged")
+	// workaround for https://github.com/go-debos/debos/issues/361
+	if d.isLikelyOldSuite() {
+		cmdline = append(cmdline, "--exclude=usr-is-merged")
```
obbardc, line 228: can we print a warning here ?

e.g. `log.Printf("excluding usr-is-merged as package is not in suite\n")`

andhe: All comments addressed. Compile-tested.

actions/debootstrap_action.go
```diff
@@ -157,6 +158,24 @@ func (d *DebootstrapAction) RunSecondStage(context debos.DebosContext) error {
...
 }
 
+// Guess if suite is something before usr-is-merged was introduced
+func (d *DebootstrapAction) isLikelyOldSuite() bool {
+	switch strings.ToLower(d.Suite) {
+	case "sid", "unstable":
+		return false
+	case "testing":
+		return false
+	case "bookworm":
```
andhe, line 168: Note: In 2023 bookworm will be stable, and with the code as is using suite name "stable" will then start to fail.

obbardc: Maybe we should leave the stable case out of this PR?

obbardc:
Manually tested PR and it fixes testing/unstable/bullseye cases and also works with Apertis (a downstream based on buster) \o/

actions/debootstrap_action.go
andhe: Yes, it is left out. What I'm saying is that "stable" needs to be added once bookworm becomes stable (and all users needs to upgrade their debos binary again).

andhe: Hopefully everyone uses release names rather than stable/testing/unstable though....

obbardc: That's something we could add a warning for (i'd accept that as a PR on top) :-)

PR #399, opened 2023-02-24
Include non-free-firmware component in Simple example

### LINKED ISSUE #398: Package 'firmware-linux' has no installation candidate
The readme file contains a [Simple example](https://github.com/go-debos/debos#simple-example) for new users to get started.

It currently fails with the following error message:
```
2023/02/24 15:20:00 apt | Package firmware-linux is not available, but is referred to by another package.
2023/02/24 15:20:00 apt | This may mean that the package is missing, has been obsoleted, or
2023/02/24 15:20:00 apt | is only available from another source
2023/02/24 15:20:00 apt | 
2023/02/24 15:20:00 apt | E: Package 'firmware-linux' has no installation candidate
2023/02/24 15:20:00 Action `apt` failed at stage Run, error: exit status 100
```

This is related to https://lists.debian.org/debian-devel-announce/2022/10/msg00001.html which implies that the `firmware-linux` package has moved to the `non-free-firmware` component. The fix should therefore be to add it to the components like this:
```
actions:
  - action: debootstrap
    suite: bookworm
    components:
      - main
      - non-free-firmware
```

Penlect:
Fixes #398

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ make a tarball.
         suite: bookworm
         components:
           - main
+          - non-free-firmware
         mirror: https://deb.debian.org/debian
         variant: minbase
 


obbardc:
thanks for your patch!

bors r+

PR #417, opened 2023-07-11
action: Remove *BaseAction.LogStart() function

obbardc:
The LogStart() function was manually called at the start of each Run() function for each action to print the initial log line for each action (to print the action name/description). Move this logic to the main loop so that we don't need to reply on an action to remember to call this function.

(cc @evelikov)

### DIFF
--- a/action.go
+++ b/action.go
@@ -3,7 +3,6 @@ package debos
 import (
 	"bytes"
 	"github.com/go-debos/fakemachine"
-	"log"
 )
 
 type DebosState int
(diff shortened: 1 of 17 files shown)


evelikov:
It's been a long day, yet I'm struggling to see how this MR is different from the one I've opened.

obbardc:
> It's been a long day, yet I'm struggling to see how this MR is different from the one I've opened.

Sorry, long day for me also. Happy to close this and let you continue.

Issue #378, opened 2022-10-31
Add lint workflow


evelikov:
We should have a lint stage catching commit mistakes like the https://github.com/go-debos/fakemachine/pull/78 one.

In the past I've tried adding one, alas things blew-up due to our ostree fork - https://github.com/evelikov/debos/commits/lint-wo-ostree.
Thinking our loud, for the short term perhaps we should move that fork into the go-debos namespace?

PR #423, opened 2023-07-26
ci: Add linter stage to workflow

obbardc:
Ensure that submitted merge requests are run through a linter check to keep high code quality.

Initially run just a formatting check due to code quality issues. Once merged we can enable useful linters one-by-one after fixing each issue.

Closes: #378

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,6 +17,35 @@ on:
   workflow_dispatch:
 
 jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+    container:
+      image: ghcr.io/go-debos/test-containers/trixie:main
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+        filter: tree:0
+
+    - name: Add cwd as git safe directory
+      run: git config --global --add safe.directory $(pwd)
+
+    - name: Download dependencies
+      run: go mod tidy
+
+    - name: Pre-build otbuiltin package
+      run: go build github.com/sjoerdsimons/ostree-go/pkg/otbuiltin
+
+    - name: golangci-lint
+      uses: golangci/golangci-lint-action@v8
+      with:
+        version: v2.3.1
+
   test:
     strategy:
       fail-fast: false
@@ -59,7 +88,9 @@ jobs:
   build:
     name: Build Docker container
     runs-on: ubuntu-latest
-    needs: test
+    needs:
+      - golangci
+      - test
     steps:
       - name: Repository checkout
         uses: actions/checkout@v4
@@ -277,6 +308,7 @@ jobs:
     name: allgreen
     if: always()
     needs:
+      - golangci
       - unit-tests
       - recipe-tests
       - example-recipes
--- a/actions/apt_action.go
+++ b/actions/apt_action.go
@@ -3,14 +3,14 @@ Apt Action
 
 Install packages and their dependencies to the target rootfs with 'apt'.
 
- # Yaml syntax:
- - action: apt
-   recommends: bool
-   unauthenticated: bool
-   update: bool
-   packages:
-     - package1
-     - package2
+	# Yaml syntax:
+	- action: apt
+	  recommends: bool
+	  unauthenticated: bool
+	  update: bool
+	  packages:
+	    - package1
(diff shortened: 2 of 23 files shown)


evelikov:
I've played around with this a while ago (see https://github.com/go-debos/debos/issues/378) but didn't manage to get past the lint issues in ostree. How did you manage to handle those?

sjoerdsimons:
@obbardc any plans to update this :)

obbardc:
> I've played around with this a while ago (see #378) but didn't manage to get past the lint issues in ostree. How did you manage to handle those?

The lint "issues" with ostree should be fixed with https://github.com/sjoerdsimons/ostree-go/pull/6 ; it is simply because there is no go.mod file there.

obbardc:
> > I've played around with this a while ago (see #378) but didn't manage to get past the lint issues in ostree. How did you manage to handle those?
> 
> The lint "issues" with ostree should be fixed with [sjoerdsimons/ostree-go#6](https://github.com/sjoerdsimons/ostree-go/pull/6) ; it is simply because there is no go.mod file there.

Hm, seems like it is more than just that. I still get `failed to load package otbuiltin: could not load export data: no export data for \"github.com/sjoerdsimons/ostree-go/pkg/otbuiltin\""`

Maybe we need to disable linting that package OR fix the issue.

obbardc:
> > > I've played around with this a while ago (see #378) but didn't manage to get past the lint issues in ostree. How did you manage to handle those?
> > 
> > 
> > The lint "issues" with ostree should be fixed with [sjoerdsimons/ostree-go#6](https://github.com/sjoerdsimons/ostree-go/pull/6) ; it is simply because there is no go.mod file there.
> 
> Hm, seems like it is more than just that. I still get `failed to load package otbuiltin: could not load export data: no export data for \"github.com/sjoerdsimons/ostree-go/pkg/otbuiltin\""`
> 
> Maybe we need to disable linting that package OR fix the issue.

OK, turns out the workaround here is to build that package first. Not sure if it's proper or not.

.github/workflows/ci.yaml
```diff
@@ -17,6 +17,35 @@ on:
...
+    - name: Download dependencies
+      run: go mod tidy
+
+    - name: Pre-build otbuiltin package
+      run: go build github.com/sjoerdsimons/ostree-go/pkg/otbuiltin
+
+    - name: golangci-lint
+      uses: golangci/golangci-lint-action@v8
+      with:
+        version: v2.3.0
```
obbardc: need to update to latest linter v2.3.1

.github/workflows/ci.yaml
```diff
@@ -17,6 +17,35 @@ on:
...
+        filter: tree:0
+
+    - name: Add cwd as git safe directory
+      run: git config --global --add safe.directory $(pwd)
+
+    - name: Download dependencies
+      run: go mod tidy
+
+    - name: Pre-build otbuiltin package
+      run: go build github.com/sjoerdsimons/ostree-go/pkg/otbuiltin
```
obbardc, line 42: need to add a doc on how to run the linter manually, something like this works:

```
$ docker run --rm -it -v $(pwd):/app -w /app golangci/golangci-lint:v2.3.1 bash
$ apt update && apt install --yes libostree-dev
$ go build github.com/sjoerdsimons/ostree-go/pkg/otbuiltin
$ golangci-lint run
```

actions/apt_action.go
```diff
@@ -3,14 +3,14 @@ Apt Action
...
 
- # Yaml syntax:
- - action: apt
-   recommends: bool
-   unauthenticated: bool
-   update: bool
-   packages:
-     - package1
-     - package2
+	# Yaml syntax:
```
obbardc, line 6: Need to see if this modifies the godoc behaviour.

obbardc: I checked this locally with `go install golang.org/x/pkgsite/cmd/pkgsite@latest && pkgsite` and the resulting HTML was fine !

obbardc:
> as a nitpick (apart from the conflict) can you order the commit to fix gofmt before introducing gofmt to keep the tree bisectable? Otherwise looks great

@sjoerdsimons I've done this. But I wanted to merge this once some of the other PRs from other contributors are solved so we don't block on weird git conflicts.

.github/workflows/ci.yaml
obbardc: This should be added to a CONTRIBUTING doc, i guess.

PR #424, opened 2023-08-04
actions: debootstrap: Add debootstrap-suite and debootstrap-script properties

### LINKED ISSUE #361: New debootstrap is unstable to bootstrap pre-bookworm debian derivatives

obbardc:


sjoerdsimons:
Would be great if you could add a test for this case

obbardc:
> Would be great if you could add a test for this case
as discussed, next step is to add some CI tests to build old/new debian/ubuntu/kali releases :-)

obbardc:
cc @elboulangero for the Kali rolling tests. Is that good enough ?

elboulangero:
> cc @elboulangero for the Kali rolling tests. Is that good enough ?

It's excellent, thanks very much for this work. Tested it, works as expected for Kali.

tests/kali/test.yaml
```diff
@@ -0,0 +1,23 @@
...
+# to ensure bootstrapping suites that debootstrap won't internally know about
+# works
+{{- $architecture := or .architecture "amd64" }}
+
+architecture: {{ $architecture }}
+
+actions:
+  - action: debootstrap
+    suite: kali-rolling
+    parent-suite: unstable
```
elboulangero, line 12: Nitpick, for correctness: Kali's `parent-suite` is `testing`.

tests/kali/test.yaml
```diff
@@ -0,0 +1,23 @@
...
+    components:
+      - main
+    mirror: https://http.kali.org/
```
elboulangero, line 15: Another nitpick: `mirror: https://http.kali.org/kali/`

actions/debootstrap_action.go
```diff
@@ -242,7 +246,16 @@ func (d *DebootstrapAction) Run(context *debos.DebosContext) error {
 	cmdline = append(cmdline, d.Suite)
 	cmdline = append(cmdline, context.Rootdir)
 	cmdline = append(cmdline, d.Mirror)
-	cmdline = append(cmdline, "/usr/share/debootstrap/scripts/unstable")
+
+	if len(d.Script) > 0 {
+		if _, err := os.Stat(d.Script); err != nil {
```
sjoerdsimons, line 258: this should be done in verify not at run time

.github/workflows/ci.yaml
```diff
@@ -136,6 +136,8 @@ jobs:
             test: { name: "arch", case: "arch" }
           - backend: kvm
             test: { name: "apertis", case: "apertis" }
+          - backend: kvm
+            test: { name: "kali", case: "kali" }
```
sjoerdsimons, line 143: Does kali not build without this? Fwiw i'm wary about adding rolling releases as they might break CI too easily

.github/workflows/ci.yaml
```diff
@@ -138,6 +138,10 @@ jobs:
             test: { name: "apertis", case: "apertis" }
           - backend: kvm
             test: { name: "kali", case: "kali" }
+          - backend: kvm
+            test: { name: "apertis v2022", case: "apertis", variables: "-t suite:v2022" }
```
sjoerdsimons, line 145: 2022 is just went EOL ; so lets not have it there; if you want something bullseye based v2023 is the right one

.github/workflows/ci.yaml
```diff
@@ -138,6 +138,10 @@ jobs:
...
+          - backend: kvm
+            test: { name: "apertis v2024dev3", case: "apertis", variables: "-t suite:v2024dev3 -t parent_suite:bookworm" }
```
sjoerdsimons, line 147: v2024 at this point

.github/workflows/ci.yaml
```diff
@@ -125,9 +125,12 @@ jobs:
...
           - { name: "templating", case: "templating", variables: " -t escaped:\\$ba\\'d\\$gers\\ snakes" }
-          - { name: "debian (amd64)", case: "debian", variables: "-t architecture:amd64" }
-          - { name: "debian (arm64)", case: "debian", variables: "-t architecture:arm64" }
-          - { name: "debian (armhf)", case: "debian", variables: "-t architecture:armhf" }
+          - { name: "debian (bookworm amd64)", case: "debian", variables: "-t architecture:amd64 -t suite:bookworm" }
+          - { name: "debian (bookworm arm64)", case: "debian", variables: "-t architecture:arm64 -t suite:bookworm" }
+          - { name: "debian (bookworm armhf)", case: "debian", variables: "-t architecture:armhf -t suite:bookworm" }
+          - { name: "debian (trixie amd64)", case: "debian", variables: "-t architecture:amd64 -t suite:trixie" }
+          - { name: "debian (trixie arm64)", case: "debian", variables: "-t architecture:arm64 -t suite:trixie" }
+          - { name: "debian (trixie armhf)", case: "debian", variables: "-t architecture:armhf -t suite:trixie" }
```
sjoerdsimons, line 133: this massively extends the number of test jobs; Does testing trixie on arm64/armhf give us extra data? Also i'm a tad wary about it breaking CI non-deterministically

sjoerdsimons (commented):
lots of questins; need some pondering if this is the right shape

.github/workflows/ci.yaml
elboulangero: Hello Sjoerd,

> Does kali not build without this

Nope, it doesn't, since it's based on Debian testing. I tried with latest debos from the Docker Hub, just to be sure:

```
$ cat recipe-kali.yaml 
architecture: amd64
actions:
  - action: debootstrap
    mirror: http://kali.download/kali/
    suite: kali-rolling
    components: [ main ]
    keyring-file: kali-archive-keyring.gpg
    keyring-package: kali-archive-keyring
  - action: apt
    packages: [ procps ]

$ podman run --rm -it --device /dev/kvm --workdir /recipes \
  --mount "type=bind,source=$(pwd),destination=/recipes" \
  --security-opt label=disable \
  godebos/debos:main-20240113 recipe-kali.yaml

Running /debos --artifactdir /recipes /recipes/recipe-kali.yaml using kvm backend
2024/01/15 04:49:58 ==== debootstrap ====
2024/01/15 04:49:58 excluding usr-is-merged as package is not in suite
2024/01/15 04:49:58 Debootstrap | I: Retrieving InRelease 
[...]
2024/01/15 04:51:41 Debootstrap | I: Base system installed successfully.
2024/01/15 04:51:42 ==== apt ====
2024/01/15 04:51:42 apt | Hit:1 http://kali.download/kali kali-rolling InRelease
2024/01/15 04:51:43 apt | Reading package lists...
2024/01/15 04:51:43 apt | Reading package lists...
2024/01/15 04:51:43 apt | Building dependency tree...
2024/01/15 04:51:43 apt | procps is already the newest version (2:4.0.4-2+b1).
2024/01/15 04:51:43 apt | You might want to run 'apt --fix-broken install' to correct these.
2024/01/15 04:51:43 apt | The following packages have unmet dependencies:
2024/01/15 04:51:43 apt |  init-system-helpers : Depends: usrmerge but it is not going to be installed or
2024/01/15 04:51:43 apt |                                 usr-is-merged
2024/01/15 04:51:43 apt | E: Unmet dependencies. Try 'apt --fix-broken install' with no packages (or specify a solution).
2024/01/15 04:51:43 Action `apt` failed at stage Run, error: exit status 100
```

I suppose all derivatives based on Debian testing/sid are affected.

Note that we can (and we do) workaround by adding this just after the debootstrap step:

```
  - description: "Install usr-is-merged (cf. debos #361 and #362)"
    action: apt
    packages: [ usr-is-merged ]
```

So it's not a dealbreaker, we can live without the MR. But it's nicer when we don't have to use cryptic workarounds.

> i'm wary about adding rolling releases as they might break CI too easily

Debootstrapping Kali rolling is just like debootstrapping Debian testing, only two packages in the set are forked, the rest is just Debian testing. Usually `kali-dev` (more or less equivalent to Debian sid) breaks every now and then, but we have solid QA that keeps `kali-rolling` working.

You might want to use the mirror `kali.download` instead of `http.kali.org`, so you'll hit Cloudflare CDN rather than our redirector, for maximum reliability.

Thanks

sjoerdsimons (changes requested):
So spending a bunch of time today thinking about it and remembering what the issue is again for background

Fundamentally the main part of this problem happens because the newer debootstrap scripts for Debian forced installing usr-is-merged (and to not install usrmerge), unless it detected a known "old" debian suite. Note that this didn't generally impact e.g. Ubuntu, tanglu or trisquel (as their debootstrap scripts don't do that).

Fundamentally the fix we did in debos was to exclude usr-is-merged unless we know the bootstrap distro was a *newer* debian to not break existing image builds for stable debian releases. For extra fun this never impacted apertis v2024 (bookworm based) as its init-system-helpers hard depend on usr-is-merged and the `--exclude` switch only applies to the initial package selection not on the expanded dependency list (cause usr-is-merged to be installed on bootstrap in apertis even though it was excluded). On Debian >= bookworm and some derivatives init-system-helpers depends on `usrmerge | usr-is-merged` , which with the exclusion by both debos and the debootstrap scripts meant neither got installed hence the issues :).

However it should be noted that this kind of gymnastics from debootstrap is exceedingly rare as should be debos doing these workaround.

Slightly orthogonally debos always forced the Debian debootstrap scripts (unstable specifically but they're all the same for any recentish debian release). Even though e.g. kali, ubuntu, tanglu, etc do have their own custom (to a smaller or bigger extend) debootstrap scripts.

So overall i think the step forward for debos *should* be to be able to use more appropriate debootstrap scripts (e.g. the ubuntu ones for ubuntu); e.g. either matched on the suite (if it's has it's own script directly) or by a property that indicates which suite to use by debootstrap (usually the direct parent, but there could be a longer lineage) or for more custom handling a deboostrap script as part of the recipe (for those needing a custom scripts, but don't have it available in debootstrap).
 
Most of that matches the patches you've already done; Though to paint some bikesheds:
* Change `parent-suite` into `debootstrap-suite` instead, with the semantic that the configured argument is what "suite" to use for debootstrap purposes (which primarily selects the right "builtin" debootstrap script but optionally does workarounds).  
* Change `script` into `debootstrap-script` with the argument being the path to the debootstrap script inside the recipe origin rather then a full path.

Having both a `debootstrap-script` and `debootstrap-suite` should cause an error as it makes no real sense. If you use a script and need workarounds they should be done in the custom script instead at which point the `debootstrap-suite` setting has no meaning anymore :)

Also at the vary least if the `suite` has a builtin debootstrap script then having a `debootstrap-suite` should at the very least be a warning but i'd even error on the side of it being an error as it's both confusing and a very weird situation.

Having a `debootstrap-script` in that situation should not be as problematic as e.g. someone is building a variant not supported in the debootstrap scripts. But the output should clearly indicate a custom script is used (probably in the debootstrap output)

For the current behaviour; we should use the current default "unstable" script as always (but also indicate the fallback in the output and potentially suggest to user to set `debootstrap-suite`) if `suite` doesn't have it's own script and neither of the new properties were given. In other words automatic fallback only happens if the user didn't explicitly ask for a specific suite or script for debootstrapping

For the usr-is-merged workaround we should only apply it when we detect a known *old* distribution (based on the suite used debootstrapping) but assume anything not specifically recognized doesn't need this workaround. This is similar to how debootstrap does this (though it only goes from the remote codename rather then the script name) and will be more future proof. 

Your current patchset already gets quite closed to this so the changes should be minimal. For e.g. kali that means there is no reason to set a `debootstrap-suite` as it already has its scripts in debootstrap and we'll pick those up now. For apertis v2024 the same goes (though building will correctly suggest it should be for consistency), while apertis v2023 (and older) now need a `debootstrap-suite` setting to build image on a newer debian base/debos, but I think that's fine.

.github/workflows/ci.yaml
obbardc: I'd like to keep the Kali test in - with the mirror change @elboulangero mentions above.

@sjoerdsimons are you happy with that ?

.github/workflows/ci.yaml
obbardc: yeah probably best to remove trixie

.github/workflows/ci.yaml
obbardc: I would like to remove the kali test since it currently complains about the keyring.

elboulangero: The new keyring is at https://archive.kali.org/archive-keyring.gpg, that should fix the test. I'm happy to do it, but I'm not sure I have the rights to add commits to this MR

obbardc: @elboulangero thanks, I have updated it and it builds again.

obbardc:
Looks like I messed up the rebase; took some time and sorted it out. I will look at the remaining comments next.

obbardc:
I rebased on main; fixed lint issues and removed the CI tests (for now). Will look at solving review comments and best way to add CI tests next.

.github/workflows/ci.yaml
obbardc: I removed the CI jobs completely. Not sure they add much value really.

obbardc:
@sjoerdsimons I squashed the commits and updated the commit message to hopefully be a bit more clear. Let me know if you want any changes.

actions/debootstrap_action.go
```diff
@@ -163,9 +178,9 @@ func (d *DebootstrapAction) RunSecondStage(context debos.DebosContext) error {
 	return err
 }
 
-// Guess if suite is something before usr-is-merged was introduced
-func (d *DebootstrapAction) isLikelyOldSuite() bool {
-	switch strings.ToLower(d.Suite) {
+// Check if suite is something before usr-is-merged was introduced
+func shouldExcludeUsrIsMerged(suite string) bool {
```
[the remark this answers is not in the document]
obbardc: I left the list the same as debootstrap; leaving jessie in. It doesn't hurt ;-)

obbardc:
@sjoerdsimons I found some time to rework this very old PR. Review appreciated.

obbardc:
Also, I wonder if we need to do something similar for the `mmdebstrap` action as part of this PR? I didn't check that action yet.

PR #433, opened 2023-08-15
actions: run: Set postprocess working directory to artifact directory

### LINKED ISSUE #345: The working directory will be set to the artifact directory
This is what is documented in https://pkg.go.dev/github.com/go-debos/debos/actions#hdr-Run_Action:

> - postprocess -- if set script or command is executed after all other commands and has access to the recipe directory ($RECIPEDIR) and the artifact directory ($ARTIFACTDIR). The working directory will be set to the artifact directory. 

In practice, it doesn't seem to be true, the working directory seem to be RECIPEDIR in any case.

Recipe:

```
architecture: amd64

actions:
  - action: run
    postprocess: true
    command: pwd
```

Output:

```
$ debos recipe.yaml 
[...]
2022/05/30 12:34:03 pwd | /home/me/recipes
[...]

$ mkdir d
$ debos --artifactdir=d recipe.yaml 
[...]
2022/05/30 12:34:31 pwd | /home/me/recipes
[...]
```

obbardc:
Currently any run action tagged as postprocess will be ran in the recipe
directory. The artifacts are stored in this directory by default, unless
--artifactdir is passed to Debos to change where the artifacts are stored.

The run action documentation states:

  postprocess -- if set script or command is executed after all other
  commands and has access to the recipe directory ($RECIPEDIR) and the
  artifact directory ($ARTIFACTDIR). The working directory will be set to
  the artifact directory.

But this is wrong; currently the working directory of postprocess commands
is set to the recipe directory. Set the working directory to the artifact
directory instead to allow postprocess commands to be ran in the correct
location.

Fixes: https://github.com/go-debos/debos/issues/345
Closes: #617 
Closes: #618 
Closes: #619

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -249,6 +249,11 @@ func (cmd Command) Run(label string, cmdline ...string) error {
 		exe.Env = append(os.Environ(), cmd.extraEnv...)
 	}
 
+	// Set working directory
+	if len(cmd.Dir) > 0 {
+		exe.Dir = cmd.Dir
+	}
+
 	// Disable services start/stop for commands running in chroot
 	if cmd.ChrootMethod != ChrootMethodNone {
 		services := ServiceHelper{cmd.Chroot}
--- a/commands.go
+++ b/commands.go
@@ -251,6 +251,11 @@ func (cmd Command) Run(label string, cmdline ...string) error {
 		exe.Env = append(os.Environ(), cmd.extraEnv...)
 	}
 
+	// Set working directory
+	if len(cmd.Dir) > 0 {
+		exe.Dir = cmd.Dir
+	}
+
 	// Disable services start/stop for commands running in chroot
 	if cmd.ChrootMethod != ChrootMethodNone {
 		services := ServiceHelper{cmd.Chroot}
--- a/tests/partitioning/test.yaml
+++ b/tests/partitioning/test.yaml
@@ -65,3 +65,18 @@ actions:
       jq . ${RECIPEDIR}/expected.json > ${SCRATCHDIR}/expected.json
       jq . ${RECIPEDIR}/actual.json > ${SCRATCHDIR}/actual.json
       diff -u ${SCRATCHDIR}/expected.json ${SCRATCHDIR}/actual.json
+
+  # Verify that postprocess commands have cwd set to artifact directory
+  # by ensuring:
+  #  - test.img exists in the current directory,
+  #  - the working directory is $ARTIFACTDIR,
+  #  - that $ARTIFACTDIR is different from $RECIPEDIR.
+  #
+  # This test should be ran with `--artifactdir` different to `--recipedir`.
+  - action: run
+    postprocess: true
+    description: Verify postprocess cwd is artifact directory
+    command: |
+      test -f test.img
+      test "${ARTIFACTDIR}" = "$(pwd)"
+      test "${ARTIFACTDIR}" != "${RECIPEDIR}"
--- a/tests/partitioning/test.yaml
+++ b/tests/partitioning/test.yaml
@@ -54,4 +54,7 @@ actions:
   - action: run
     description: Compare expected and actual
     chroot: false
-    command: bash -c 'diff -u <(jq . ${RECIPEDIR}/expected.json) <(jq . ${RECIPEDIR}/actual.json)'
+    command: |
+      jq . ${RECIPEDIR}/expected.json > ${SCRATCHDIR}/expected.json
+      jq . ${RECIPEDIR}/actual.json > ${SCRATCHDIR}/actual.json
+      diff -u ${SCRATCHDIR}/expected.json ${SCRATCHDIR}/actual.json
--- a/tests/partitioning/test.yaml
+++ b/tests/partitioning/test.yaml
@@ -15,6 +15,7 @@ actions:
         fs: ext2
         start: 0%
         end: 256M
+        partuuid: 7BA1B99D-7942-450A-921B-F394A0A065AF
       - name: system
         fs: ext4
         start: 256m
@@ -24,26 +25,32 @@ actions:
         fs: ext4
         start: 2G
         end: 3G
+        partuuid: 713CA942-5DB4-41F8-A365-E8F1CA967D7A
       - name: data1
         fs: ext4
         start: 3G
         end: 4G
+        partuuid: 05D13CA0-7EA3-48D2-9824-8C94263E5692
       - name: data2
         fs: ext4
         start: 4G
         end: 5G
+        partuuid: C86924DB-49D2-47D8-BBB7-BB56426A8934
       - name: data3
         fs: ext4
         start: 5G
         end: 6G
+        partuuid: B0F4D633-B66A-44B6-933A-E77A0144B275
       - name: data4
         fs: ext4
         start: 6G
         end: 7G
+        partuuid: 8BDAE85E-E473-4D77-B09D-43E1D6F741DA
       - name: data5
         fs: ext4
         start: 7G
         end: 8G
+        partuuid: F9272482-3F7E-44E3-8D99-C24023EB3317
 
   - action: run
     chroot: false
--- a/tests/partitioning/test.yaml
+++ b/tests/partitioning/test.yaml
@@ -65,3 +65,18 @@ actions:
       jq . ${RECIPEDIR}/expected.json > /tmp/expected.json
       jq . ${RECIPEDIR}/actual.json > /tmp/actual.json
       diff -u /tmp/expected.json /tmp/actual.json
+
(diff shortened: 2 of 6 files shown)


commands.go
```diff
@@ -254,6 +254,11 @@ func (cmd Command) Run(label string, cmdline ...string) error {
 		exe.Env = append(os.Environ(), cmd.extraEnv...)
 	}
 
+	// Allow working directory to be set for commands not running in chroot
+	if len(cmd.Dir) > 0 && cmd.ChrootMethod == CHROOT_METHOD_NONE {
```
sjoerdsimons, line 258: why only for CHROOT_METHOD_NONE

sjoerdsimons (changes requested):
We should really ensure this get exercised in at least one of the tests

sjoerdsimons:
@obbardc will you have to time to nudge this forward? the outstanding comments were quite small

obbardc:
> @obbardc will you have to time to nudge this forward? the outstanding comments were quite small

yes, I will be able to take a look at the beginning of March, along with my other outstanding PRs :-)

commands.go
obbardc: Not sure why it was only for that, I can't remember my thinking at the time. It should for for all chroot methods right ? Maybe I never needed the other CHROOT_METHOD at the time :-)

obbardc: Done.

obbardc:
Not sure what test I can write for this. Other than a run action with postprocess confirming the artifact dir. Any ideas @sjoerdsimons @daissi ?

sjoerdsimons:
> Not sure what test I can write for this. Other than a run action with postprocess confirming the artifact dir. Any ideas @sjoerdsimons @daissi ?

I think that's it right? Just confirming the postprocess cwd matches what you'd expect ? e.g. by writing an artifact  in an action and then checking you can access in a postprocessing step

tests/run/test.yaml
```diff
@@ -0,0 +1,14 @@
+architecture: amd64
+
+actions:
+  # Check the pwd is set to the artifact directory
+  # TODO Set artifactdir to somewhere other than recipedir and check it
+  # or  confirming the postprocess cwd matches what you'd expect ? e.g. by writing an artifact in an action and then checking you can access in a postprocessing step
```
sjoerdsimons: Fwiw you could do an image-partition action  to get debos to create an artifact in the artifactdir; and then as the postprocess command run `test -f <image file>` which should succeed with and without artifactdir set on the command line

sjoerdsimons: @copilot Could you create a test for this PR using the recipe based testing strategy, implementation suggestion above

obbardc: @sjoerdsimons I am happy to cherry-pick the Copilot commits from #617 to this PR & fixup, if that makes sense ?

obbardc: OK, I did that and kept the authorship. Let me know if there are any changes you'd like. I'd like to get this finished in the next week.

tests/run/test.yaml
```diff
@@ -0,0 +1,18 @@
+architecture: amd64
+
+actions:
+  # Create an artifact (test.img) in the artifact directory
+  - action: image-partition
+    description: Create test image in artifact directory
+    imagename: test.img
+    imagesize: 1M
+    partitiontype: msdos
```
obbardc: maybe better to use a run command with `dd`

tests/run/test.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+  # Verify that postprocess commands have cwd set to artifact directory
+  # by checking if test.img exists in the current directory.
+  # This test is run with --artifactdir=run so the artifact directory
+  # is different from the recipe directory.
+  - action: run
+    postprocess: true
+    description: Verify postprocess cwd is artifact directory
```
obbardc: something like; "Verify file exists" and update comment too

tests/run/test.yaml
sjoerdsimons: I like using image-partition as that double-checks a few things as in the end the main use-case for post-processing tends to be to do something with the image; 

That said would probably make sense to add that it to an existing test that creates images, to keep the number of jobs down.  Probably something we for fun could ask @copilot as well, but i suspect it'll do the creating a new branch/PR thing again.. Lets see what it does

tests/run/test.yaml
sjoerdsimons: Ack and for extra doublecheck we could test if e.g. ${ARTIFACTDIR} == $(pwd) or something silly like that.  (cc @copilot )

obbardc:
I've added the changes from #618 and #619 as a draft commit to see if the test works fine in CI. It will need squashing once it works.

tests/run/test.yaml
obbardc: done; makes sense!

tests/run/test.yaml
obbardc: done

tests/partitioning/test.yaml
```diff
@@ -55,3 +55,13 @@ actions:
     description: Compare expected and actual
     chroot: false
     command: bash -c 'diff -u <(jq . ${RECIPEDIR}/expected.json) <(jq . ${RECIPEDIR}/actual.json)'
+
+  # Verify that postprocess commands have cwd set to artifact directory
+  # by checking if test.img exists in the current directory and that
+  # the current directory equals $ARTIFACTDIR.
+  # This test is run with --artifactdir=run so the artifact directory
```
sjoerdsimons: utter nitpick: This is a lie as its with run with artifactdir=scratch.

wonder if to be sure we should enforce in this test that ARTIFACTDIR != RECIPEDIR

obbardc: good catch; i updated the test

sjoerdsimons (dismissed):

obbardc:
@sjoerdsimons added some fixes for the partitioning test; hope it's ok. Disabled auto merge so you can review before merging... Cheers!

obbardc:
For some reason it fails as the image can't be found in artifactdir in CI (but of course it works locally). Added some test commits to check...


hmm, in the CI run, `ls -la $ARTIFACTDIR` returns:

```
2025/12/07 23:19:30 ==== test ====
2025/12/07 23:19:30 ls -la ${ARTIFACTDIR} | total 4
2025/12/07 23:19:30 ls -la ${ARTIFACTDIR} | drwxrwxrwt  4 root root   80 Dec  7 23:19 .
2025/12/07 23:19:30 ls -la ${ARTIFACTDIR} | drwxrwxrwt 15 root root  440 Dec  7 23:19 ..
2025/12/07 23:19:30 ls -la ${ARTIFACTDIR} | drwxr-xr-x  3 root root 4096 Dec  7 23:19 mnt
2025/12/07 23:19:30 ls -la ${ARTIFACTDIR} | drwxr-xr-x  2 root root   40 Dec  7 23:19 root
```

obbardc:
> For some reason it fails as the image can't be found in artifactdir in CI (but of course it works locally). Added some test commits to check...
> 
> hmm, in the CI run, `ls -la $ARTIFACTDIR` returns:
> 
> ```
> 2025/12/07 23:19:30 ==== test ====
> 2025/12/07 23:19:30 ls -la ${ARTIFACTDIR} | total 4
> 2025/12/07 23:19:30 ls -la ${ARTIFACTDIR} | drwxrwxrwt  4 root root   80 Dec  7 23:19 .
> 2025/12/07 23:19:30 ls -la ${ARTIFACTDIR} | drwxrwxrwt 15 root root  440 Dec  7 23:19 ..
> 2025/12/07 23:19:30 ls -la ${ARTIFACTDIR} | drwxr-xr-x  3 root root 4096 Dec  7 23:19 mnt
> 2025/12/07 23:19:30 ls -la ${ARTIFACTDIR} | drwxr-xr-x  2 root root   40 Dec  7 23:19 root
> ```

OK, fixed it!

obbardc:
I moved the changes for fixing the partitioning test into #696, they probably need to go into their own PR anyway.

obbardc:
> I moved the changes for fixing the partitioning test into #696, they probably need to go into their own PR anyway.

OK, rebased onto https://github.com/go-debos/debos/pull/708 instead.

tests/partitioning/test.yaml
```diff
@@ -65,3 +65,18 @@ actions:
       jq . ${RECIPEDIR}/expected.json > /tmp/expected.json
       jq . ${RECIPEDIR}/actual.json > /tmp/actual.json
       diff -u /tmp/expected.json /tmp/actual.json
+
+  # Verify that postprocess commands have cwd set to artifact directory
```
obbardc, line 69: maybe this should move into its own test rather than hijacking the partitioning test.

PR #444, opened 2023-11-02
docker: update to Debian 12 (Bookworm)

kgoger:
Bookworm is stable since June. So let's update the Dockerfile to use it so we can get rid of all the backports.
This required to add the systemd-resolved package as it was split into a separate package from systemd-networkd.

Motivation for the upgrade was have dpkg with zstd support so we can debootstrap Ubuntu again.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,7 +3,7 @@ ARG DEBIAN_FRONTEND=noninteractive
 ARG GOPATH=/usr/local/go
 
 ### first stage - builder ###
-FROM debian:bullseye-slim as builder
+FROM debian:bookworm-slim as builder
 
 ARG DEBIAN_FRONTEND
 ARG GOPATH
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -96,26 +80,16 @@ RUN apt-get update && \
...
-# lacks pactrap.
-# Install Arch dependencies
-RUN apt-get update && \
-    apt-get install -y --no-install-recommends \
+        zip \
         makepkg \
-        pacman-package-manager && \
-    apt-get install -y --no-install-recommends \
-        -t bullseye-backports \
+        pacman-package-manager \
```
obbardc, line 91: as a follow-up note we should remove the arch packages from the debian-based images and build arch-based images separately (#445).

obbardc (approved):

sjoerdsimons (approved):

PR #446, opened 2023-11-10
Update new fakemachine and shell escaping 

sjoerdsimons:
Draft for now until fakemachine v0.0.7 is released


actions/recipe.go
```diff
@@ -29,7 +29,13 @@ Comments are allowed and should be prefixed with '#' symbol.
      # Use value of variable 'Var' defined above
      property2: {{$Var}}
 
-Mandatory properties for receipt:
+
+Two custom template functions are available
```
obbardc, line 33: nit: don't limit it to two, more may be available later,
nit: append :

obbardc (approved):
looks good.

PR #451, opened 2023-11-24
Add slim-sprig functions to template language

obbardc:


actions/recipe.go
```diff
@@ -34,6 +34,9 @@ The following custom template functions are available
 
 - sector: Returns the argument * 512 (convential sector size) e.g. `{{ sector 64 }}`
 - escape: Shell escape the  argument `{{ escape $var }}`
+- decimal: Returns the integer argument as a decimal integer e.g. `{{ decimal 0x40 }} => 64`
+- add: Returns the calculation of two integer arguments `a + b`, e.g. `{{ add 1 1 }} => 2`
+- sub: Returns the calculation of two integer arguments `a - b`, e.g. `{{ sub 10 1 }} => 9`
```
sjoerdsimons, line 39: If we go this route i'd rather pull in exist library with template functions; doing some google makes sprig come up (https://pkg.go.dev/github.com/go-task/slim-sprig/v3#section-readme) with includes those and many more integer operations etc.

sjoerdsimons: that's the full sprig mind ; slim-spring limits sprig to not depend on extra dependencies (which may avoid the CI failure you're seeing)..

Also can you clean up you branch so it doesn't have commits that get converted later and drop the decimal function ? that should be unneeded (int should do the trick there)

sjoerdsimons: Can you also add a minimal test case to ensure the functions are available? Also it seems to fail to build on bullseye, though i'm ok to drop support for that assuming it's due to a golang version issue.

PR #455, opened 2023-12-05
docker: Install archlinux-keyring from source

obbardc:
The arch linux tests require a newer version of archlinux-keyring than what is in Debian stable; so revert back to pulling the latest version.

This partially reverts 36cf33366bbe0dd602185a87a241ff070517f868.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -26,6 +26,24 @@ COPY . $GOPATH/src/github.com/go-debos/debos
 WORKDIR $GOPATH/src/github.com/go-debos/debos/cmd/debos
 RUN go install ./...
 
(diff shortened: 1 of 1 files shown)


obbardc:
> Should probably add a scarier inline message, otherwise the next person will remove the code.
Do you have a suggestion?

`# WARNING: removing this will most likely break archlinux support` is not good enough ?

obbardc:
@evelikov Can we pull the master branch of the keyring ? Seems like the tag you originally used is outdated (before the version in Debian, which is most lilkely why it was replaced with the keyring pacakge in Debian).

obbardc:
@evelikov How does this look now then?

PR #459, opened 2023-12-08
Build example recipes in GitHub Actions

obbardc:
To ensure that we don't ship broken example recipes (as reported in some
previous bugs), we should build the example recipes and make sure they
build before pushing new container images.


.github/workflows/ci.yaml
```diff
@@ -236,13 +236,50 @@ jobs:
...
+          path: /tmp
+      - name: Load image
+        run: |
+          docker load --input /tmp/debos-image.tar
+
+      - name: Build example Debos recipe ${{matrix.example}}
+        run:
+          docker run
+          --cgroupns=private
+          -v $(pwd)/tests:/tests
```
daissi:
```suggestion
          -v $(pwd)/doc/examples:/tests
```
I guess this is the reason of the CI failure since it fails with:
`2025/06/27 09:56:17 stat /tests/ospack-debian/ospack-debian.yaml: no such file or directory`

daissi:
Thanks @obbardc I like it! Just found a small issue :)

.github/workflows/ci.yaml
obbardc: Think I have solved it.

PR #469, opened 2024-01-11
doc/man: fix typos

jwilk:


obbardc:
please fix readme then regenerate man page :-)

PR #478, opened 2024-01-30
Add option to quiet fakemachine

bergmx:
Hi! Thanks for the work on this useful project.

We have successfully been using debos for some time, but it always prints the whole debos invocation, which causes secrets passed into debos as variables (e.g. account passwords, private keys) to be written to our CI logs. I looked into this and found that it was because fakemachine logs the command it runs by default, though it has an option to suppress this. I saw no way of setting this option through debos, though, hence this pull request.

This pull request adds the flag "--quiet-fakemachine" because it was the least intrusive change, but on a second thought, it looks a bit odd. I think it might make more sense to use the quiet option in fakemachine by default, and let the existing debos flag "--verbose" disable it instead. What do you think? Any approach that lets us pass a variable without having it written to log is fine with us.

### DIFF
--- a/cmd/debos/debos.go
+++ b/cmd/debos/debos.go
@@ -316,6 +316,9 @@ func main() {
 			}
 		}
 
+		// Silence extra output from fakemachine unless the --verbose flag was passed.
+		m.SetQuiet(!options.Verbose)
+
 		exitcode, err = m.RunInMachineWithArgs(args)
 		if err != nil {
 			fmt.Println(err)


obbardc:
@bergmx Thanks for this, I think i'd rather disable the fakemachine invocation log message by default (e.g. `SetQuiet(false)`) & instead enable it with `debos --verbose`.

It may be nicer to disable the message by default and have `--verbose` and `machine.SetVerbose(true|false)` on the fakemachine side, but that is hindsight ;-).

bergmx:
Sounds good to me! I made new commit with that change.

doc/man/debos.1
```diff
@@ -1,5 +1,5 @@
 '\" t
-.\" Automatically generated by Pandoc 2.17.1.1
+.\" Automatically generated by Pandoc 3.1.3
```
obbardc, line 2: drop this change

cmd/debos/debos.go
```diff
@@ -317,7 +316,8 @@ func main() {
 			}
 		}
 
-		m.SetQuiet(options.QuietFakeMachine)
+		// Silence extra output from fakemachine unless the --verbose flag was passed.
```
obbardc, line 319: can you please squash the two commits & use the top commit as the commit message ?

obbardc (changes requested):

doc/man/debos.1
bergmx: Should already be gone, since I undid all documentation changes

cmd/debos/debos.go
bergmx: Sure! Done.

bergmx:
Hi again @obbardc ! Looks like the merge is blocked. I can't see anything I can do on my end, so maybe there's a button to press on your end? Let me know if there is something I have to do.

obbardc (approved):

obbardc:
I had to review it again manually! Should be all done now though. Thanks again for your contribution!

Issue #483, opened 2024-02-26
create separate Arch container


obbardc:
Create separate Arch container

PR #484, opened 2024-02-26
docker: Remove Arch support from Debian-based container

obbardc:
It turns out that trying to bootstrap an Arch system from Debian-packaged tools is quite hard since Arch moves much quicker than Debian Bookworm. Any upstream Arch churn can cause the Debos container build to break (this has happened twice now).

Since this really is an anti-pattern in Debos - we suggest building an OS image from the packages which make that OS - remove the Arch bits from the Debian Dockerfile to stop breakages. The hope is as time goes on we can create a separate Arch-based Dockerfile for Arch support.

Link: https://github.com/go-debos/debos/issues/483

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -26,24 +26,6 @@ COPY . $GOPATH/src/github.com/go-debos/debos
 WORKDIR $GOPATH/src/github.com/go-debos/debos/cmd/debos
 RUN go install ./...
 
-# Install the latest archlinux-keyring, since the one in Debian is bound
-# to get outdated sooner or later.
-# WARNING: returning to the debian package will break the pacstrap action
-RUN apt-get update && \
-    apt-get install -y --no-install-recommends \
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -26,24 +26,6 @@ COPY . $GOPATH/src/github.com/go-debos/debos
...
-RUN apt-get update && \
-    apt-get install -y --no-install-recommends \
-        pkgconf \
-        python3-all \
-        sq \
-        systemd \
-        make && \
-    rm -rf /var/lib/apt/lists/*
-
-RUN git clone https://gitlab.archlinux.org/archlinux/archlinux-keyring && \
```
sjoerdsimons, line 41: the problem is not so much arch; the problem is trying to rebuild their keyring. The debian packaging for this simply uses the prebuild keyrings instead (part of every release)

PR #490, opened 2024-03-14
partitioning: Add partattrs property to partition

nfraprado:
Add a 'partattrs' property to partitions. It allows the GPT partition attribute bits to be set for a partition. Notably this is needed to set bits 48 and 56, which are part of the GUID specific range, to allow a ChromeOS Kernel partition to be booted on Chromebooks.

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -58,6 +58,7 @@ a 32 bits hexadecimal number (e.g. '1234ABCD' without any dash separator).
 	   fsck: bool
 	   fsuuid: string
 	   partuuid: string
+	   partattrs: list of partition attribute bits to set
 
 Mandatory properties:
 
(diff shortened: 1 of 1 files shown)


actions/image_partition_action.go
```diff
@@ -95,6 +96,9 @@ for partition.
 - flags -- list of additional flags for partition compatible with parted(8)
 'set' command.
 
+- partattrs -- list of GPT partition attribute bits. See description for
```
obbardc, line 99: Can you actually add the usage notes (and some examples?) here, rather than pointing to sfdisk manpage ?

I'd like to understand if adding a new parameter is the right way to go.

nfraprado: @obbardc Added now. Let me know what you think.

actions/image_partition_action.go
```diff
@@ -58,6 +58,7 @@ a 32 bits hexadecimal number (e.g. '1234ABCD' without any dash separator).
 	   fsck: bool
 	   fsuuid: string
 	   partuuid: string
+	   partattrs: list of partition attributes
```
sjoerdsimons, line 61:
```suggestion
	   partattrs: list of partition attributes bits to set
```
Would be good to have the `Verify` function check it's only bit numbers in a valid range

sjoerdsimons (changes requested):
One small note; would also be great to extend the paritioning tests to test partition attributes

actions/image_partition_action.go
nfraprado: Sorry for the delay. I've implemented verification now.

However I've noticed that bits 0-2 despite being documented in the specification and in the sfdisk man page, don't seem to be supported:

```
2024/05/31 19:42:27 sfdisk | unsupported GPT attribute bit '2'
2024/05/31 19:42:27 sfdisk | sfdisk: /dev/disk/by-fakemachine-label/fakedisk-0: partition 1: failed to set partition attributes
```

Should we forbid these values in verification too? Given it's documented it might be something specific to my setup. Unfortunately I don't have a lot of time to look into this.

daissi: However, it's possible to set them using their name i.e. `LegacyBIOSBootable` instead of `2` in this case.

sjoerdsimons:
Closing as it's been integrated through #554

PR #491, opened 2024-03-19
actions: image-partition: enable creation of FAT{12|16|32} partitions

obbardc:
see commit for detail

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -342,8 +342,21 @@ func (i ImagePartitionAction) formatPartition(p *Partition, context debos.DebosC
 
 	cmdline := []string{}
 	switch p.FS {
-	case "vfat":
-		cmdline = append(cmdline, "mkfs.vfat", "-F32", "-n", p.Name)
+	case "fat", "fat12", "fat16", "fat32", "msdos", "vfat":
+		cmdline = append(cmdline, "mkfs.vfat", "-n", p.Name)
+
+		switch p.FS {
+		case "fat12":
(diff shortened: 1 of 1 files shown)


actions/image_partition_action.go
```diff
@@ -342,8 +342,17 @@ func (i ImagePartitionAction) formatPartition(p *Partition, context debos.DebosC
 
 	cmdline := []string{}
 	switch p.FS {
+	case "fat16":
+	case "fat32":
```
sjoerdsimons, line 346: probably makes sense to add fat12 as well for completeness and maybe fat which relies on the mkfs autoselection?

obbardc: yeah adding `fat` here changes behaviour so I decided to keep this commit minimal. I will add it.

actions/image_partition_action.go
```diff
@@ -498,7 +511,11 @@ func (i ImagePartitionAction) Run(context *debos.DebosContext) error {
 
 		command := []string{"parted", "-a", "none", "-s", "--", context.Image, "mkpart", name}
 		switch p.FS {
-		case "vfat":
+		case "fat16":
+			command = append(command, "fat16")
+		case "fat", "fat12", "fat32", "msdos", "vfat":
+			/* TODO: Not sure if this is correct. Perhaps
+			   fat12 should be treated the same as fat16 ? */
```
sjoerdsimons, line 518: fat12 has it's own unique partition id on msdos labels ; but parted seems to only recognize this, you can't create one. So :man_shrugging:

PR #497, opened 2024-05-13
download: Actually validate empty pathnames

fooishbar:
From the path module doc:
    If the path is empty, Base returns ".". If the path consists
    entirely of slashes, Base returns "/".

Actually validate that we provide a filename when one is required, rather than failing with an error that /scratch exists but is not a file.


fooishbar:
The CI failing isn't my fault, but I can't see how to retry.

obbardc (approved):

PR #501, opened 2024-05-21
actions: fix fstab generated in case of FAT{12|16|32}.

azayet01:
Recent changes introduced a tuned formatting of the partitions created with FAT{12|16|32}. The issue is that fstab is generated using the fs property which won't match the system expectations.

If the fs property is "fat", "fat12", "fat16", "fat32" or "msdos", then fstab must use "vfat".

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -270,8 +270,17 @@ func (i *ImagePartitionAction) generateFSTab(context *debos.DebosContext) error 
 				fs_passno = 2
 			}
 		}
+
+		fsType := m.part.FS
+		switch m.part.FS {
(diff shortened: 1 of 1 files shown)


obbardc:
Thank you for this fix!

azayet01:
@obbardc ,
You are welcome ! thanks for reviewing and reacting fast ! Very much appreciated. There are the two critical stuff that I have. But for the remaining contribution , it's not critical so I will be pushing some proposals but not critical to have.

Have a great day !

obbardc:
@azayet01 Please feel free to open issues/PRs about anything you wish to discuss :-). Thank you again for your contributions!

PR #505, opened 2024-06-11
docker: Add arm64-based Dockerfile

tboehler1:
I needed an arm64-based Docker image to run debos but it seems like you don't provide one.

This PR contains the Dockerfile I use to build images on arm64. Arch Linux specific things (notably the archlinux-keyring) aren't installed and handled at all, but seeing that https://github.com/go-debos/debos/issues/483 is open this might be put into a separate conatiner for building Arch Linux images anyway?

Feel free to merge if you're okay with the Dockerfile and want to support arm64-based Docker images upstream. Of course I'm open to improve this as well.

Thanks!

### DIFF
--- a/docker/Dockerfile.arm64
+++ b/docker/Dockerfile.arm64
@@ -1,0 +1,98 @@
+# Global ARGs shared by all stages
+ARG DEBIAN_FRONTEND=noninteractive
+ARG GOPATH=/usr/local/go
+
+### first stage - builder ###
+FROM arm64v8/debian:bookworm-slim as builder
+
+ARG DEBIAN_FRONTEND
+ARG GOPATH
+ENV GOPATH=${GOPATH}
+
+# install debos build and unit-test dependencies
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        ca-certificates \
+        gcc \
+        git \
+        golang-go \
+        libc6-dev \
+        libostree-dev \
+        unzip && \
+    rm -rf /var/lib/apt/lists/*
+
+# Build debos
+COPY . $GOPATH/src/github.com/go-debos/debos
+WORKDIR $GOPATH/src/github.com/go-debos/debos/cmd/debos
+RUN go install ./...
+
+### second stage - runner ###
+FROM arm64v8/debian:bookworm-slim as runner
+
+ARG DEBIAN_FRONTEND
+ARG GOPATH
+
+# Set HOME to a writable directory in case something wants to cache things
+ENV HOME=/tmp
+
+LABEL org.label-schema.name "debos"
+LABEL org.label-schema.description "Debian OS builder"
+LABEL org.label-schema.vcs-url = "https://github.com/go-debos/debos"
+LABEL org.label-schema.docker.cmd 'docker run \
+  --rm \
+  --interactive \
+  --tty \
+  --device /dev/kvm \
+  --user $(id -u) \
+  --workdir /recipes \
+  --mount "type=bind,source=$(pwd),destination=/recipes" \
+  --security-opt label=disable'
+
+# debos runtime dependencies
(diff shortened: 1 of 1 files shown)


obbardc:
This isn't built in ci? :-)

Also, would it be possible to just extend the actions to set the build arch to arm64 with the same Dockerfile ?

tboehler1:
With actions you mean the steps running in the CI? I'm not familiar with the GitHub CI at all, I'm not sure if I can be of help there.

I imagine using `ARG` in the Dockerfile and passing `arm64` with `--build-arg ARCH=arm64` doesn't work because there is *some* logic involved with the different architectures, i.e. `arm64` would imply using `arm64v8/debian` but it would also imply installing `linux-image-arm64` and not installing `user-mode-linux` (as it's not available for arm64). I'm not aware of Docker being able to handle logic this in Dockerfiles but do correct me if I'm wrong.

nbuchwitz:
How about something like 

```
diff --git a/docker/Dockerfile b/docker/Dockerfile
index dd4d3ab..89e2de5 100644
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,9 +1,10 @@
 # Global ARGs shared by all stages
+ARG ARCH=
 ARG DEBIAN_FRONTEND=noninteractive
 ARG GOPATH=/usr/local/go
 
 ### first stage - builder ###
-FROM debian:bookworm-slim as builder
+FROM ${ARCH}debian:bookworm-slim AS builder
 
 ARG DEBIAN_FRONTEND
 ARG GOPATH
@@ -34,7 +35,20 @@ COPY docker/get-archlinux-keyring.sh /
 RUN /get-archlinux-keyring.sh /arch-keyring
 
 ### second stage - runner ###
-FROM debian:bookworm-slim as runner
+FROM ${ARCH}debian:bookworm-slim AS runner-amd64
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        linux-image-amd64 && \
+        user-mode-linux && \
+
+FROM ${ARCH}debian:bookworm-slim AS runner-arm64
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        linux-image-arm64 && \
+
+FROM runner-${TARGETARCH} AS runner
 
 ARG DEBIAN_FRONTEND
 ARG GOPATH
@@ -78,7 +92,6 @@ RUN apt-get update && \
         pigz \
         libostree-1-1 \
         libslirp-helper \
-        linux-image-amd64 \
         openssh-client \
         parted \
         pkg-config \
@@ -91,7 +104,6 @@ RUN apt-get update && \
         systemd-resolved \
         u-boot-tools \
         unzip \
-        user-mode-linux \
         xfsprogs \
         xz-utils \
         zip \
```

This would allow to override certain arch specific stuff in a separate stage. Even diffrent base images depending on the target arch are possible. 

Build is done with explicit target arch:
```
docker buildx build --platform linux/arm64/v8 -f docker/Dockerfile -t godebos/debos .
```

and without:
```
docker buildx build -f docker/Dockerfile -t godebos/debos .
```

Even multiple architectures can be build at once:
```
docker buildx build --platform linux/arm/v7,linux/arm64/v8,linux/amd64 -f docker/Dockerfile -t godebos/debos .
```

obbardc:
@nbuchwitz I agree with your solution. Can you open a PR?

nbuchwitz:
Sure, @obbardc. I've created a feature branch in my fork https://github.com/nbuchwitz/debos/tree/devel/add-arm64-support

Need to rework some of the piplines to ensure that the multiarch container is tested properly (mostly migrate the build artifacts to oci exports). After that I will open the PR (probably later today or tomorrow).

@tboehler1 I've added you as co autor on the commit message.

obbardc:
Fixed in #513

Issue #509, opened 2024-06-20
Using "--memory" breaks fakemachine backend


tboehler1:
Some time in the last 2 months I've been unable to use debos with the "--memory" flag. The fakemachine backend keeps running into OOM errors, hangs completely, and times out after a few minutes.

To reproduce:

```
$ cat repro.yaml
architecture: arm64

actions:
  - action: run
    command: "echo 'Hello World'"
```
Run this using `debos --memory 8192 repro.yaml`

Outputs:

- [`debos --memory 8192 repro.yaml`](https://github.com/user-attachments/files/15911718/normal_output.txt)
- [`debos --show-boot --verbose --memory 8192 repro.yaml`](https://github.com/user-attachments/files/15911720/verbose_output.txt)

flxzt:
Did you try using `8192MB` ? Without a unit I think it parses as only 8192 bytes (see [here](https://github.com/go-debos/debos/blob/fb02f98c6d327dadeae375e51d4f45dd393da233/cmd/debos/debos.go#L254))

tboehler1:
That was indeed the problem and when specifying `8192MB` it works. Sorry for the noise!

obbardc:
I created https://github.com/go-debos/debos/pull/510 to hopefully warn when a user forgets to add a suffix and not change existing behaviour. :-)

PR #510, opened 2024-07-01
Warn if parsed memory/scratchsize may be too small

obbardc:
Since the --memory and --scratchsize arguments are parsed with human-readable suffixes into bytes (e.g 2048MB or 2GB), if a user does not put a suffix the arguments are parsed as bytes which can cause issues later when a user sets the memory to be 2048 assuming it will be parsed as MB.

Change the documentation to match reality and also add a warning if the user has set either parameter less than the recommended default.

Closes: #509

### DIFF
--- a/cmd/debos/debos.go
+++ b/cmd/debos/debos.go
@@ -68,9 +68,9 @@ func main() {
 		TemplateVars  map[string]string `short:"t" long:"template-var" description:"Template variables (use -t VARIABLE:VALUE syntax)"`
 		DebugShell    bool              `long:"debug-shell" description:"Fall into interactive shell on error"`
 		Shell         string            `short:"s" long:"shell" description:"Redefine interactive shell binary (default: bash)" optionsl:"" default:"/bin/bash"`
-		ScratchSize   string            `long:"scratchsize" description:"Size of disk backed scratch space"`
+		ScratchSize   string            `long:"scratchsize" description:"Size of disk-backed scratch space (parsed with human-readable suffix; assumed bytes if no suffix)"`
 		CPUs          int               `short:"c" long:"cpus" description:"Number of CPUs to use for build VM (default: 2)"`
-		Memory        string            `short:"m" long:"memory" description:"Amount of memory for build VM (default: 2048MB)"`
+		Memory        string            `short:"m" long:"memory" description:"Amount of memory for build VM (parsed with human-readable suffix; assumed bytes if no suffix. default: 2Gb)"`
 		ShowBoot      bool              `long:"show-boot" description:"Show boot/console messages from the fake machine"`
 		EnvironVars   map[string]string `short:"e" long:"environ-var" description:"Environment variables (use -e VARIABLE:VALUE syntax)"`
 		Verbose       bool              `short:"v" long:"verbose" description:"Verbose output"`
@@ -265,7 +265,12 @@ func main() {
 			exitcode = 1
 			return
 		}
-		m.SetMemory(int(memsize / 1024 / 1024))
+
+		memsizeMB := int(memsize / 1024 / 1024)
+		if memsizeMB < 256 {
+			log.Printf("WARNING: Memory size of %dMB is less than recommended minimum 256MB\n", memsizeMB)
+		}
+		m.SetMemory(memsizeMB)
 
 		if options.CPUs == 0 {
 			// Set default CPU count for fakemachine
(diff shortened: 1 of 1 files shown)


daissi:
Something is wrong with --scratchsize:
```
./debos ../debos-recipes/rpi3/debimage-rpi3.yaml --scratchsize 100MB -m 100MB
2024/11/14 18:35:32 WARNING: Memory size of 100MB is less than recommended 2048MB
2024/11/14 18:35:32 WARNING: Scratch size of 95MB is less than recommended 2048MB
```
While is works for --memory, the conversion for scratchsize doesn't give us the same value 100MB vs 95MB

obbardc:
> Something is wrong with --scratchsize:
> 
> ```
> ./debos ../debos-recipes/rpi3/debimage-rpi3.yaml --scratchsize 100MB -m 100MB
> 2024/11/14 18:35:32 WARNING: Memory size of 100MB is less than recommended 2048MB
> 2024/11/14 18:35:32 WARNING: Scratch size of 95MB is less than recommended 2048MB
> ```
> 
> While is works for --memory, the conversion for scratchsize doesn't give us the same value 100MB vs 95MB

I guess that is because for scratchsize we convert to bytes with `units.FromHumanSize` but for memory we do not :-).

Perhaps we should unify this by:

1) using `units.FromHumanSize` to convert for both scratchsize & memory.
2) convert to MB using forumula `/ 1000 / 1000` for both variables instead of `/ 1024 / 1024`. That should make the printed number the same.

To be honest, I threw this together quickly just for some feedback. Happy to rework it.

daissi:
I would say let's use for both `units.FromHumanSize` for consistency :) This should avoid confusion.

sjoerdsimons:
> I would say let's use for both `units.FromHumanSize` for consistency :) This should avoid confusion.

go-units isn't great `FromHumanSize` is really more for disk sizes (decimal units) while `RamInBytes`  follows binary units which is more suitable for memory (see https://en.wikipedia.org/wiki/Byte#Multiple-byte_units).  However annoyingly it will accept both MiB and MB for either, but will parse them according to the predetermined table.... So if you do `FromHumanSize("100Mib")` you'll get `100 * 1000**2` 

For this PR assuming a MB for disk is `1000**2` while a MB for ram is `1024**2`  would be the correct solution. Ideally FromHumanSize to recognize MiB as binary, but doesn't look like that will happen ([go-units issue](https://github.com/docker/go-units/issues/31)). However the expectation of what an `MB` is sadly just is different depending on the context :man_shrugging:

sjoerdsimons (changes requested):
see comment :)

obbardc:
@sjoerdsimons made teh changes according to your suggestions :-)

cmd/debos/debos.go
```diff
@@ -265,7 +265,12 @@ func main() {
 			exitcode = 1
 			return
 		}
-		m.SetMemory(int(memsize / 1024 / 1024))
+
+		memsizeMB := int(memsize / 1024 / 1024)
+		if memsizeMB < 2048 {
+			log.Printf("WARNING: Memory size of %dMB is less than recommended 2048MB\n", memsizeMB)
```
sjoerdsimons, line 271: Fwiw less then  2G is fine; It's probably when it's less then say 512m or even 256 i'd start really questioning  things.

cmd/debos/debos.go
```diff
@@ -280,6 +285,11 @@ func main() {
 				exitcode = 1
 				return
 			}
+
+			scratchsizeMB := int(size / 1000 / 1000)
+			if scratchsizeMB < 2048 {
+				log.Printf("WARNING: Scratch size of %dMB is less than recommended 2048MB\n", scratchsizeMB)
```
sjoerdsimons, line 291: fwiw scratchsize is in particular surprising when it's e.g. less then the vm's memory as that makes it somewhat pointless.  2G is quite a bit for most systems we tend to be build, for most 1G or even 512M would still be fine.

sjoerdsimons (changes requested):
Sorry i should have looked at the sizes more closely yesterday, i just ended up being in the 1024 vs 1000 rathole too much ;) TL;DR I think giving warnings is great, but the current threshold seems a bit low, if you run on a small VM (e.g. 1 or 2G memory instance) then having a smaller then 2G memory size is fine.  Especially if you use a scratchdisk, where the VMs memory mostly just needs to be enough to run apt.

So lets be a bit less trigger-happy on these warnings, as you said it's mostly for people forgetting to add `M`. Which we'll still easily filter out even with a very conservative "accepted" minimum

cmd/debos/debos.go
obbardc: changed it to say "recommended minimum" and set that limit to be 256M

cmd/debos/debos.go
obbardc: changed it to say "recommended minimum" and set that limit to be 512M

obbardc: I could check if its smaller than RAM here, but i don't think it's worth it.

PR #513, opened 2024-08-19
docker: Add arm64 builds to Dockerfile and CI

nbuchwitz:
Follow-up of @tboehler1 's #505 as suggested by @obbardc 

Convert Dockerfile so multiple target architecturebuilds are possible. Also modify the CI pipeline, in order to built a single multi-platform container.

Note that at the current time-being docker cannot export multi-platform containers without an intermediate registry. Thus only the amd64 image is exported for the test stage. Nevertheless the arm64 image is built in the pipeline and also published later to the registry.

If the go-debos project has access to GitHub large runners, the pipeline could be reworked in order to build the container on a native arm runner. This would bring a significant speed improvement and also would allow for testing the arm container. But for know this solution should be sufficient as it tests the build on both architectures and runs the test suite on amd64.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -133,7 +133,7 @@ jobs:
 
       - name: Run unit tests
         run: |
-          docker-compose -f docker/unit-tests.test.yml \
+          docker compose -f docker/unit-tests.test.yml \
             up --exit-code-from=sut
 
   recipe-tests:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,6 +2,7 @@ name: Build and Test
 
 env:
   GITHUB_TAG: ghcr.io/${{ github.repository }}
+  DOCKER_TARGET_PLATFORMS: linux/amd64,linux/arm64
 
 on:
   push:
@@ -51,6 +52,9 @@ jobs:
       - name: Repository checkout
         uses: actions/checkout@v4
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
       - name: Setup Docker buildx
         uses: docker/setup-buildx-action@v3
 
@@ -67,6 +71,26 @@ jobs:
           push: false
           pull: true
           file: docker/Dockerfile
+          platforms: ${{ env.DOCKER_TARGET_PLATFORMS }}
+          tags: debos
+          cache-from: type=local,src=/tmp/.build-cache
+          cache-to: type=local,dest=/tmp/.build-cache,mode=max
+
+      # WORKAROUND:
+      # Docker buildx cannot export multi-platform images with type=docker,
+      # but only with type=oci. The latter cannot be imported without an
+      # intermediate registry. Thus export only the amd64 image for usage
+      # in the test stage.
+      #
+      # see https://docs.docker.com/reference/cli/docker/buildx/build/#docker
+      - name: Export amd64 Docker image for later usage
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          push: false
+          pull: true
+          file: docker/Dockerfile
+          platforms: linux/amd64
           tags: debos
           cache-from: type=local,src=/tmp/.build-cache
           cache-to: type=local,dest=/tmp/.build-cache,mode=max
@@ -158,6 +182,7 @@ jobs:
         with:
           name: debos-image
           path: /tmp
+
       - name: Load image
         run: |
           docker load --input /tmp/debos-image.tar
@@ -241,6 +266,7 @@ jobs:
         with:
           context: .
           push: true
+          platforms: ${{ env.DOCKER_TARGET_PLATFORMS }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
           file: docker/Dockerfile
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -67,6 +70,26 @@ jobs:
           push: false
           pull: true
           file: docker/Dockerfile
+          platforms: linux/amd64,linux/arm64
+          tags: debos
+          cache-from: type=local,src=/tmp/.build-cache
+          cache-to: type=local,dest=/tmp/.build-cache,mode=max
+
+      # WORKAROUND:
```
obbardc, line 78: In future, could we work around this with a registry in the runner ? We should really run tests on both docker images :-)

.github/workflows/ci.yaml
```diff
@@ -241,6 +265,7 @@ jobs:
         with:
           context: .
           push: true
+          platforms: linux/amd64,linux/arm64
```
obbardc, line 268: expand to a variable e.g `DOCKER_PLATFORMS` ?

docker/Dockerfile
```diff
@@ -1,9 +1,10 @@
 # Global ARGs shared by all stages
+ARG ARCH=
```
obbardc, line 2: I think this needs to be more descriptive e.g `IMAGE_PREFIX_ARCH`

docker/Dockerfile
```diff
@@ -34,7 +35,24 @@ COPY docker/get-archlinux-keyring.sh /
...
+        qemu-system-x86 \
+        user-mode-linux && \
+    rm -rf /var/lib/apt/lists/*
+
+FROM ${ARCH}debian:bookworm-slim AS runner-arm64
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        linux-image-arm64 \
+        qemu-system-arm \
+       # fixes: qemu-system-aarch64: failed to find romfile "efi-virtio.rom"
```
obbardc, line 51: spacing to be consistent with these two lines :-)

obbardc (changes requested):

obbardc:
We only have amd64 runners, I think.

.github/workflows/ci.yaml
nbuchwitz: Makes sense. I moved the supported platforms into `DOCKER_TARGET_PLATFORMS`

docker/Dockerfile
nbuchwitz: Actually this isn't really needed as we use buildkit and a base image which already has support for the target platforms. Thus I have removed the `ARCH` prefix in the latest series. Damn musle memory from times before multi platform support in Docker :-)

.github/workflows/ci.yaml
nbuchwitz: Yes this would resolve the issue. It also makes it really easy for a reviewer to test a PR's content within a container. With a proper retention policy the containers can be cleaned up afterwards. Alternatively a hook on PR close/merge could do the cleanup.

nbuchwitz:
All of the review feedback should have been addressed with the latest patch series. I was also able to figure out, why the unit tests failed earlier. Turns out this was unrelated to my modifications and more an issue with GitHub switching over from v1 to v2 compose api. Fixed this in another commit. 

CI works as expected (in my fork): https://github.com/nbuchwitz/debos/actions/runs/10537327064

.github/workflows/ci.yaml
obbardc: I created https://github.com/go-debos/debos/issues/514 to track the unit tests. I don't think we should block on that for merging this PR.

obbardc (approved):

nbuchwitz:
The failed job in https://github.com/go-debos/debos/actions/runs/10537327383/job/29218707965?pr=513#step:6:3778 looks like the arch server was unresponsive during curl run. Could you please restart the job?

obbardc:
> The failed job in https://github.com/go-debos/debos/actions/runs/10537327383/job/29218707965?pr=513#step:6:3778 looks like the arch server was unresponsive during curl run. Could you please restart the job?

Sorry @nbuchwitz I haven't been able to dedicate any time to debos recently. I can't seem to run the job. Can you navigate to 
 https://github.com/nbuchwitz/debos/actions/workflows/ci.yaml and run the workflow manually for your branch ?

Thank you!

nbuchwitz:
I've ran the CI as requested here: https://github.com/nbuchwitz/debos/actions/runs/11581471675

obbardc:
@nbuchwitz it still shows as not passing on the PR and I can't see how to force it. Perhaps you have to force-push a new commit ?

nbuchwitz:
The pipeline status in this pull request in inherited from merge request job and not the job which ran in my fork. Thus the job needs to be retriggered here. Just did a force push, guess you will have to start the job

obbardc:
> The pipeline status in this pull request in inherited from merge request job and not the job which ran in my fork. Thus the job needs to be retriggered here. Just did a force push, guess you will have to start the job

I've approved it, it's running! Apologies for all of that.

nbuchwitz:
Fingers crossed :smile:

nbuchwitz:
All green! Looking forward to point our build servers to the native arm64 releases after merge ! :partying_face:

obbardc:
Happy with this PR. Thanks for your contribution @nbuchwitz

PR #515, opened 2024-08-25
actions: Add mmdebstrap action

nbuchwitz:
Add mmdebstrap [1] as a faster and more flexible alternative to debootstrap.

[1] https://gitlab.mister-muffin.de/josch/mmdebstrap

I am not really familar with go (day to day language is Python or C). Therefore I ask for your patience :-)

# Comparison to deboostrap

Example recipe on a Raspberry Pi 5 in docker container with KVM enabled:
```
---
{{- $architecture := or .architecture "arm64"}}
{{- $tool := or .tool "mmdebstrap" }}
architecture: {{$architecture}}

actions:
  - action: {{ $tool }}
    suite: bookworm
    variant: minbase

  - action: apt
    description: Install some base packages
    packages:
      - procps
```

With `deboostrap`:
```
real    0m53.590s
user    0m38.792s
sys     0m25.345s

```

With `mmdebstrap`:
```
real    0m33.400s
user    0m24.239s
sys     0m8.838s
```

=> 20 seconds faster. I plan to build a kinda build-on-demand service around debos, so every second matters :smile:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -151,7 +151,8 @@ jobs:
           - { name: "templating", case: "templating", variables: " -t escaped:\\$ba\\'d\\$gers\\ snakes" }
           - { name: "partitioning", case: "partitioning" }
           - { name: "msdos partitioning", case: "msdos" }
-          - { name: "debian (amd64)", case: "debian", variables: "-t architecture:amd64" }
+          - { name: "debian (amd64, debootstrap)", case: "debian", variables: "-t architecture:amd64" }
+          - { name: "debian (amd64, mmdebstrap)", case: "debian", variables: "-t architecture:amd64 -t tool:mmdebstrap" }
         exclude:
           - backend: nofakemachine
             test: { name: "partitioning", case: "partitioning" }
@@ -159,9 +160,13 @@ jobs:
             test: { name: "msdos partitioning", case: "msdos" }
         include:
           - backend: kvm
-            test: { name: "debian (arm64)", case: "debian", variables: "-t architecture:arm64" }
+            test: { name: "debian (arm64, debootstrap)", case: "debian", variables: "-t architecture:arm64" }
           - backend: kvm
-            test: { name: "debian (armhf)", case: "debian", variables: "-t architecture:armhf" }
+            test: { name: "debian (arm64, mmdebstrap)", case: "debian", variables: "-t architecture:arm64 -t tool:mmdebstrap" }
+          - backend: kvm
+            test: { name: "debian (armhf, debootstrap)", case: "debian", variables: "-t architecture:armhf" }
+          - backend: kvm
+            test: { name: "debian (armhf, mmdebstrap)", case: "debian", variables: "-t architecture:armhf -t tool:mmdebstrap" }
           - backend: kvm
             test: { name: "arch", case: "arch" }
           - backend: kvm
--- a/actions/mmdebstrap_action.go
+++ b/actions/mmdebstrap_action.go
@@ -17,6 +17,7 @@ and this may lead to incorrect configuration when becoming part of the created r
    keyring-files:
    include:
    dpkg-opts:
+   apt-opts:
 
 Mandatory properties:
 
@@ -45,6 +46,8 @@ Example:
 
 - dpkg-opts -- list of arbitrary options to dpkg.
 
+- apt-opts -- list of arbitrary options to apt.
+
 */
 package actions
 
@@ -69,6 +72,7 @@ type MmdebstrapAction struct {
 	MergedUsr        *bool `yaml:"merged-usr"`
 	Include          []string
 	DpkgOpts         []string `yaml:"dpkg-opts"`
+	AptOpts          []string `yaml:"apt-opts"`
 }
 
 func NewMmdebstrapAction() *MmdebstrapAction {
@@ -160,6 +164,12 @@ func (d *MmdebstrapAction) Run(context *debos.DebosContext) error {
 	if d.DpkgOpts != nil {
 		for _, opt := range d.DpkgOpts {
 			cmdline = append(cmdline, fmt.Sprintf("--dpkgopt='%s'", opt))
+		}
+	}
+
+	if d.AptOpts != nil {
+		for _, opt := range d.AptOpts {
+			cmdline = append(cmdline, fmt.Sprintf("--aptopt='%s'", opt))
 		}
 	}
 
--- a/actions/mmdebstrap_action.go
+++ b/actions/mmdebstrap_action.go
@@ -15,6 +15,7 @@ and this may lead to incorrect configuration when becoming part of the created r
    variant: "name"
    keyring-packages:
    keyring-files:
+   include:
 
 Mandatory properties:
 
@@ -39,6 +40,8 @@ Example:
 
 - merged-usr -- use merged '/usr' filesystem, fallback to distribution default if not set.
 
+- include -- list of packages to install during bootstrap.
+
 */
 package actions
 
@@ -61,6 +64,7 @@ type MmdebstrapAction struct {
 	KeyringFiles     []string `yaml:"keyring-files"`
 	Components       []string
 	MergedUsr        *bool `yaml:"merged-usr"`
+	Include          []string
 }
 
 func NewMmdebstrapAction() *MmdebstrapAction {
@@ -142,6 +146,11 @@ func (d *MmdebstrapAction) Run(context *debos.DebosContext) error {
 
 	if d.Variant != "" {
 		cmdline = append(cmdline, fmt.Sprintf("--variant=%s", d.Variant))
+	}
+
+	if d.Include != nil {
+		s := strings.Join(d.Include, ",")
+		cmdline = append(cmdline, fmt.Sprintf("--include=%s", s))
 	}
 
 	cmdline = append(cmdline, d.Suite)
--- a/actions/mmdebstrap_action.go
+++ b/actions/mmdebstrap_action.go
@@ -16,6 +16,7 @@ and this may lead to incorrect configuration when becoming part of the created r
    keyring-packages:
    keyring-files:
    include:
+   dpkg-opts:
 
 Mandatory properties:
 
@@ -42,6 +43,8 @@ Example:
 
 - include -- list of packages to install during bootstrap.
 
+- dpkg-opts -- list of arbitrary options to dpkg.
+
 */
 package actions
 
@@ -65,6 +68,7 @@ type MmdebstrapAction struct {
 	Components       []string
 	MergedUsr        *bool `yaml:"merged-usr"`
 	Include          []string
+	DpkgOpts         []string `yaml:"dpkg-opts"`
 }
 
 func NewMmdebstrapAction() *MmdebstrapAction {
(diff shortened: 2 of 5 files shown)


obbardc:
looks fine to me, but I remember @sjoerdsimons had some reservations about ai ply integrating another tool and would rather rewrite debootstrap in go utilising more efficient system libs (debootstrap is single threaded, assumes not much is installed on the target system etc etc)

nbuchwitz:
> looks fine to me, but I remember @sjoerdsimons had some reservations about ai ply integrating another tool and would rather rewrite debootstrap in go utilising more efficient system libs (debootstrap is single threaded, assumes not much is installed on the target system etc etc)

Thanks for the fast response and the heads up. `mmdebstrap` deals with some major issues of `deboostrap` (some of them are even wontfix, like multiple mirrors etc). So maybe it is worth considering it (as an intermediate solution). Looking forward to @sjoerdsimons feedback

iwamatsu:
Hi, 

If you plan to apply this feature, please consider mmdebstrap's include variable feature.
I create a patch for this  feature:
  https://github.com/iwamatsu/debos/commit/b9f3f28e78ebbd3b6d4b79682f7d230bf07e4365

actions/mmdebstrap_action.go
```diff
@@ -0,0 +1,184 @@
...
+
+	err := debos.Command{}.Run("Mmdebstrap", cmdline...)
+
+	/* Cleanup resolv.conf after mmdebstrap */
+	resolvconf := path.Join(context.Rootdir, "/etc/resolv.conf")
+	if _, err = os.Stat(resolvconf); !os.IsNotExist(err) {
+		if err = os.Remove(resolvconf); err != nil {
+			return err
+		}
+	}
```
tboehler1, line 190: Making this part of a defer at the beginning of the function avoids problems down the line if more lines were added after the mmdebstrap call and before the removal of the resolv.conf that would exit our early on error. That way, if the resolv.conf exists, it will *always* be removed, and there's no chance for mistakes.

tboehler1 (changes requested):

actions/mmdebstrap_action.go
tboehler1: It's idiomatic go to put function cleanup for both normal return and error in a defer and should something like recovering from an error here ever be implemented this trivial change could prevent a potential bug. But it's your choice.

sjoerdsimons:
> > Hi,
> > If you plan to apply this feature, please consider mmdebstrap's include variable feature. I create a patch for this feature: [iwamatsu@b9f3f28](https://github.com/iwamatsu/debos/commit/b9f3f28e78ebbd3b6d4b79682f7d230bf07e4365)
> 
> Thanks! I've picked your commit into this PRs branch

Fwiw this makes the settings different from debootstrap; With debootstrap there were a whole bunch of issues around doing includes this early, though hopefully mmdebstrap fixes that because it actually uses apt. 

That said i'm not sure about this change; Are there use-cases where this has real benefits over mmdebstrap + apt actions to install the extra packages?

actions/mmdebstrap_action.go
```diff
@@ -0,0 +1,193 @@
...
+	Components       []string
+	MergedUsr        *bool `yaml:"merged-usr"`
+	Include          []string
+}
+
+func NewMmdebstrapAction() *MmdebstrapAction {
+	d := MmdebstrapAction{}
+	// Be secure by default
+	// Use main as default component
+	d.Components = []string{"main"}
```
sjoerdsimons, line 75: MergdUsr should be true by default

sjoerdsimons: ah you did make it a pointer nvm :)

actions/mmdebstrap_action.go
```diff
@@ -0,0 +1,193 @@
...
+	if d.Components != nil {
+		s := strings.Join(d.Components, ",")
+		cmdline = append(cmdline, fmt.Sprintf("--components=%s", s))
+	}
+
+	/* Only works for amd64, arm64 and riscv64 hosts, which should be enough */
+	foreign := context.Architecture != runtime.GOARCH
+
+	if foreign {
+		cmdline = append(cmdline, fmt.Sprintf("--architectures=%s", context.Architecture))
```
sjoerdsimons, line 151: with mmdebstrap you can avoid this hack; just always set the `--architecture` option

actions/mmdebstrap_action.go
sjoerdsimons: This is the same flow as we have for debootstrap; so it's fine by me; 

Fwiw like in the debootstrap command if we run commands in the chroot later, you most likely actually want to have `resolv.conf` cleaned up at that point (as our chroot command helper has it's only `resolv.conf` handling).  In that case a defer would actually be wrong :)

That said this code is currently masking the `err` return from the mmdebstrap command, which is something that really should be fix

.github/workflows/ci.yaml
```diff
@@ -152,6 +152,7 @@ jobs:
           - { name: "partitioning", case: "partitioning" }
           - { name: "msdos partitioning", case: "msdos" }
           - { name: "debian (amd64)", case: "debian", variables: "-t architecture:amd64" }
+          - { name: "debian with mmdebstrap (amd64)", case: "debian", variables: "-t architecture:amd64 -t tool:mmdebstrap" }
```
sjoerdsimons, line 155: Also add tests for arm64 and armhf to ensure different archs also work.. Feel free to change the commends to `debian (amd64, debootstrap)` and `debian (amd64, mmdebstrap)` rather then making mmdebstrap special

sjoerdsimons (changes requested):
Some small improvements; but overall looking nice..

Fwiw i've always been against other bootstrapping tools directly in debos; However it seems mmdebstrap actually got enough traction (with being used on the buildds now), that it seems a safe bet for a long-term viable tool

actions/mmdebstrap_action.go
nbuchwitz: Well spotted with the masked `err`, I've fixed it and will leave the rest of the logic as is in `debootstrap`.

actions/mmdebstrap_action.go
nbuchwitz: Fixed in latest force push

.github/workflows/ci.yaml
nbuchwitz: Fixed in latest force push

nbuchwitz:
Pipeline is also fixed, so this PR is ready for another round of feedback

actions/mmdebstrap_action.go
```diff
@@ -0,0 +1,177 @@
...
+	Variant          string
+	KeyringPackages  []string `yaml:"keyring-packages"`
+	KeyringFiles     []string `yaml:"keyring-files"`
+	Components       []string
+	MergedUsr        *bool `yaml:"merged-usr"`
+}
+
+func NewMmdebstrapAction() *MmdebstrapAction {
+	d := MmdebstrapAction{}
+	// Be secure by default
```
sjoerdsimons, line 80: Drop the leftover comment

actions/mmdebstrap_action.go
```diff
@@ -0,0 +1,177 @@
...
+ If no components are specified debos will use main as default.
+
+Example:
+ components: [ main, contrib ]
+
+- keyring-packages -- list of keyrings for package validation.
+
+- keyring-files -- list keyring files for repository validation.
+
+- merged-usr -- use merged '/usr' filesystem, true by default.
```
sjoerdsimons, line 43: As i understand it correctly; it's not true by default but rather follows the distribution default; so this seems incorrect

actions/mmdebstrap_action.go
```diff
@@ -0,0 +1,177 @@
...
+
+	return nil
+}
+
+func (d *MmdebstrapAction) Run(context *debos.DebosContext) error {
+	cmdline := []string{"mmdebstrap"}
+
+	if d.MergedUsr != nil {
+		if *d.MergedUsr {
+			cmdline = append(cmdline, "--include=usrmerge")
```
sjoerdsimons, line 133: Open question whether this should be the method rather then using `/usr/share/mmdebstrap/hooks/merged-usr`  (which follows how deboostrap sets things up).

sjoerdsimons (changes requested):

sjoerdsimons:
@nbuchwitz Just a small nitpicks left and a question; Fwiw please don't add more features in this MR, lets focus on getting it merged  :) Extra features can easily come as a follow-up

actions/mmdebstrap_action.go
nbuchwitz: I really don't have strong feelings for either solution. Nevertheless, after some thinking,  `/usr/share/mmdebstrap/hooks/merged-usr` seems to be a good compromise, as a user can choose the other way (thanks to @iwamatsu's patch) by adding the `usrmerge` package to the `include` list.

actions/mmdebstrap_action.go
nbuchwitz: I rephrased the help text in order to point out, that if the attribute is not specified it falls back to the distributions default.

actions/mmdebstrap_action.go
nbuchwitz: Hm, tests are failed with the hooks script: I can reproduce the same issue with plain `mmdebstrap` and bullseye suite:

```
$ mmdebstrap --hook-dir=/usr/share/mmdebstrap/hooks/no-merged-usr --components=main --architectures=amd64 --variant=minbase bullseye rootfs.tar   
I: automatically chosen mode: unshare
I: amd64 cannot be executed natively, but transparently using qemu-user binfmt emulation
I: finding correct signed-by value...
done
I: automatically chosen format: tar
I: using /tmp/mmdebstrap.d4bPhIhmgf as tempdir
dpkg: warning: failed to open configuration file '/home/nbw/.dpkg.cfg' for reading: Permission denied
I: running --setup-hook directly: /usr/share/mmdebstrap/hooks/no-merged-usr/setup00.sh /tmp/mmdebstrap.d4bPhIhmgf
Warning: starting with Debian 12 (Bookworm), systems without merged-/usr are not supported anymore
dpkg-deb: building package 'usr-is-merged' in '/tmp/mmdebstrap.d4bPhIhmgf/tmp/tmp.zBxyHOzSaK/usr-is-merged.deb'.
dpkg: warning: failed to open configuration file '/home/nbw/.dpkg.cfg' for reading: Permission denied
Selecting previously unselected package usr-is-merged.
(Reading database ... 0 files and directories currently installed.)
Preparing to unpack .../usr-is-merged.deb ...
Unpacking usr-is-merged (1) ...
Setting up usr-is-merged (1) ...
I: running apt-get update...
done
I: downloading packages with apt...
done
I: extracting archives...
done
I: installing essential packages...
done
I: running --essential-hook directly: /usr/share/mmdebstrap/hooks/no-merged-usr/essential00.sh /tmp/mmdebstrap.d4bPhIhmgf
Reading package lists... Done
Building dependency tree... Done
usr-is-merged is already the newest version (1).
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
E: setup failed: E: command failed: /usr/share/mmdebstrap/hooks/no-merged-usr/essential00.sh
I: main() received signal PIPE: waiting for setup...
I: removing tempdir /tmp/mmdebstrap.d4bPhIhmgf...
```

nbuchwitz: Looks like `/usr/share/mmdebstrap/hooks/maybe-merged-usr` is the safer hook: It will not try to merge if unsupported and otherwise merge. I refactored the logic so `MergedUsr` is now a bool (like in debootstrap) which defaults to `true`. With this both debos actions should behave similar.

sjoerdsimons: Well if we set merged-user to true; it should really either fails or end up with merged user rather then a maybe-merged-usr :). @obbardc  was looking at improving the per-suite handling in #424, but yeah it's all a bit of nightmare.

I'll try to have a further look tonight as well; That said if mmdebstrap setting `merged-usr: true` only really works from bookworm onwards that's not the end of the world (something we can document). And as you said including `usrmerge` is a workaround for older ones?

nbuchwitz: Ok then let's keep the current approach (haven't force pushed my changed described above). In order to make the tests work again, we should change the `suite` to `bookworm` (makes sense anyway as it is stable and will become oldstable this year ...)?

sjoerdsimons (approved):

Issue #430, opened 2023-08-13
add --version flag which prints version


jerkstorecaller:
Self-explanatory.

PR #517, opened 2024-10-30
cmd: Add 'version' option

daissi:
This 'version' option prints the debos version. Fix: #430

### DIFF
--- a/cmd/debos/debos.go
+++ b/cmd/debos/debos.go
@@ -15,6 +15,8 @@ import (
 	"github.com/jessevdk/go-flags"
 )
 
+const Version = "1.1.4"
+
 func checkError(context *debos.DebosContext, err error, a debos.Action, stage string) int {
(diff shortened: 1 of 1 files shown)


cmd/debos/debos.go
```diff
@@ -110,6 +113,11 @@ func main() {
 		}
 	}
 
+	if options.Version {
+		fmt.Printf("debos %s\n", Version)
```
obbardc, line 117:
```suggestion
		fmt.Printf("debos v%s\n", Version)
```

daissi: Thank you! Fixed :)

obbardc:
Can you do the same thing for fakemachine ? There's an open issue, i think.

daissi:
> Can you do the same thing for fakemachine ? There's an open issue, i think.

Sure, I will do it tomorrow.

PR #521, opened 2024-11-18
docker: install qemu from backports

obbardc:
Seems that cross-arch Python scripts are failing inside fakemachine
due to qemu-user issue. Install latest qemu from backports to work
around this issue.


tests/debian/test.yaml
```diff
@@ -4,7 +4,7 @@ architecture: {{$architecture}}
 
 actions:
   - action: debootstrap
-    suite: bullseye
+    suite: bookworm
```
obbardc, line 7: The bullseye bootstrapping test on arm64 fails with newer qemu...

sjoerdsimons:
@obbardc is this still required or can this be closed?

obbardc:
> @obbardc is this still required or can this be closed?

Yes, seems like we need newer qemu. I'll look closer in the new year once I am back to work.

obbardc:
@sjoerdsimons The reason this is needed is due to running amd64 python bits in a fakemachine. apt seems to fall over when instlaling packages which call e.g. a python script during postinst.

For now we have worked around this for our nightly build by using a custom-built docker container from this branch; but it'd be good to fix it properly. I think using qemu-user-static from backports is a bit of a "hack"

I will add a test in to check this issue and try to reproduce it first.

obbardc:
hmm, seems like UML is borked. maybe this is the time to disable it.

EDIT: that is tracked as https://github.com/go-debos/fakemachine/issues/241

sjoerdsimons:
Can you clean up the commit history a bit; it's rather confusing atm :)

obbardc:
> Can you clean up the commit history a bit; it's rather confusing atm :)

yeah, i plan to do so this week && properly disable UML (as it is broken). still a draft :-)

tests/exit_test/exit_test.sh
```diff
@@ -54,20 +54,20 @@ fi
 expect_success debos --help
 expect_failure debos --not-a-valid-option
 expect_failure debos
-expect_failure debos good.yaml good.yaml
+expect_failure debos --fakemachine-backend=qemu good.yaml good.yaml
```
sjoerdsimons: Does the autodetection not work as expected?

docker/Dockerfile
```diff
@@ -35,15 +35,14 @@ COPY docker/get-archlinux-keyring.sh /
...
 
 ### second stage - runner ###
-FROM debian:bookworm-slim AS runner-amd64
+FROM debian:trixie-slim AS runner-amd64
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
         linux-image-amd64 \
-        qemu-system-x86 \
-        user-mode-linux && \
+        qemu-system-x86 && \
```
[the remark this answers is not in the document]
obbardc: It is documented in the commit message :-)

tests/exit_test/exit_test.sh
obbardc: i can't remember why this commit was needed; that's a benefit of adding a clear commit message ;-). let me revert it and see.

obbardc: I will leave this PR as a draft until I figure out what is going on with the exit tests.

obbardc: right ok this was a workaround because we haven't tagged a new fakemachine with https://github.com/go-debos/fakemachine/commit/cd96d69e17d0b0cd52318beb676a2dadf8003111 yet ; i will go and do that now & remove these commits.

obbardc: sorted that out, works fine now

PR #536, opened 2025-02-18
Add support for alternate sector size

julien-massot:
Depends on: https://github.com/go-debos/fakemachine/pull/222

Newer storage, such as UFS or NVMe disk, may not support the
default 512 sector size.

For these devices, the partition table and the partition offset
need to be aligned with sector size boundaries. The GPT partition
table is 34 sectors in length, so we can't have compatible disk images
between 512 and 4096 sector-size devices.

This patch introduces a major breakage with the {{ sector 64 }}
template.
At template processing, we can't guess the appropriate sector size
for the image since this value can be defined as a YAML header.
Instead of multiplying the sector number by the default 512 bytes
boundary, append a 's' suffix. This defers the work of computing
the value to the different Actions.
The 's' suffix is chosen to be compatible with parted (ImagePartition)
and support has been added to raw action.

My main concern is about the 'sector' template function:

- We can either find a way to know the 'sectorsize' header at template processing
  time.
 - Keep the current behavior sector X -> X * 512 but mention that this is broken for non-default
   sector size
 -  Compute the value thanks to the 's' suffix  in any appropriate actions


cmd/debos/debos.go
```diff
@@ -272,6 +273,7 @@ func main() {
 			options.CPUs = 2
 		}
 		m.SetNumCPUs(options.CPUs)
+		m.SetSectorSize(r.SectorSize)
```
obbardc, line 276: you need to mention a dependency on the PR in fakemachine && also update go.mod to your WIP branch

obbardc (changes requested):

cmd/debos/debos.go
julien-massot: Ah yes so obvious Thanks

julien-massot: should be ok now, but I'm not familiar with go, so let see :)

.github/workflows/ci.yaml
```diff
@@ -162,10 +163,16 @@ jobs:
...
           - backend: kvm
             test: { name: "debian (armhf)", case: "debian", variables: "-t architecture:armhf" }
+          - backend: kvm
+            test: { name: "raw (4096 sector size)", case: "raw", variables: "-t sectorsize:4096" }
           - backend: kvm
             test: { name: "arch", case: "arch" }
           - backend: kvm
             test: { name: "apertis", case: "apertis" }
+          - backend: kvm
+            test: { name: "512 sector size GPT partition table", case: "partitionning-sector-size", variables: "-t sectorsize:512" }
```
nbuchwitz, line 173:
```suggestion
            test: { name: "512 sector size GPT partition table", case: "partitioning-sector-size", variables: "-t sectorsize:512" }
```

.github/workflows/ci.yaml
```diff
@@ -162,10 +163,16 @@ jobs:
...
+          - backend: kvm
+            test: { name: "4096 sector size GPT partition table", case: "partitionning-sector-size", variables: "-t sectorsize:4096" }
```
nbuchwitz, line 175:
```suggestion
            test: { name: "4096 sector size GPT partition table", case: "partitioning-sector-size", variables: "-t sectorsize:4096" }
```

actions/recipe.go
```diff
@@ -43,6 +44,11 @@ Mandatory properties for recipe:
 
 - actions -- at least one action should be listed
 
+Optionnal properties for recipe:
```
nbuchwitz, line 47:
```suggestion
Optional properties for recipe:
```

julien-massot: thanks

.github/workflows/ci.yaml
julien-massot: fixed

julien-massot:
Exclude raw test case on 'nofakemachine' backend since it depends on partitioning.

sjoerdsimons (changes requested):
Some small changes; Also the no fakemachine case should either be supported or give an error if the recipe has an unsupported sector size.. Unfortunately it looks like the go-losetup mod doesn't support setting the sector size (and looks  unmaintained), so i guess it should be the latter.

julien-massot:
> Some small changes; Also the no fakemachine case should either be supported or give an error if the recipe has an unsupported sector size.. Unfortunately it looks like the go-losetup mod doesn't support setting the sector size (and looks unmaintained), so i guess it should be the latter.

Thanks for the review,

Indeed, the go-losetup looks unmaintained. We are already using some commands, so I decided to directly use
the 'losetup' command to change the loop file sector size.
Let me know if this solution sounds good enough to you, 'losetup' comes with the 'mount' package (on Debian)
so it should be available on most OS.

actions/recipe.go
```diff
@@ -43,6 +44,11 @@ Mandatory properties for recipe:
 
 - actions -- at least one action should be listed
 
+Optional properties for recipe:
+
+- sectorsize: Override the default 512 sectorsize, mandatory for device using
+4k block size such as UFS or NVMe storage.
```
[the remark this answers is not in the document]
julien-massot: done

actions/raw_action.go
```diff
@@ -18,9 +18,10 @@ Mandatory properties:
 
 Optional properties:
 
-- offset -- offset in bytes for output image file.
-It is possible to use internal templating mechanism of debos to calculate offset
-with sectors (512 bytes) instead of bytes, for instance: '{{ sector 256 }}'.
+- offset -- offset in bytes or in sector number e.g 256s.
+The sector size is either the recipe header 'sectorsize' or the default 512 sector
+size.
+Internal templating mechanism will append the 's' suffix, for instance: '{{ sector 256 }}' will be converted to '256s'.
```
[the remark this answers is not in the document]
julien-massot: done

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/go-debos/debos
 
 go 1.23
 
+replace github.com/go-debos/fakemachine => github.com/julien-massot/fakemachine v0.0.0-20250217102638-89bdc3eff408
```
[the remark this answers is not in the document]
julien-massot: Replaced by the main branch while waiting for a release.

actions/recipe.go
```diff
@@ -43,6 +44,12 @@ Mandatory properties for recipe:
 
 - actions -- at least one action should be listed
 
+Optional properties for recipe:
+
+- sectorsize: Overrides the default 512 sectorsize, mandatory for device using
```
obbardc, line 49:
```suggestion
- sectorsize: Overrides the default 512 byte sectorsize, mandatory for device using
```

actions/image_partition_action.go
```diff
@@ -481,6 +482,15 @@ func (i *ImagePartitionAction) PreNoMachine(context *debos.DebosContext) error {
 	if err != nil {
 		return fmt.Errorf("Failed to setup loop device: %v", err)
 	}
+
+	if context.SectorSize != 512 {
+		command := []string{"losetup", "--sector-size", strconv.Itoa(context.SectorSize), i.loopDev.Path()}
```
obbardc, line 487: It'd be good to add a comment as to why you're not using a library function from losetup & possibly a link to an issue requesting the feature ?

actions/recipe.go
julien-massot: Hi fixed thanks.

actions/image_partition_action.go
julien-massot: Just added a comment and openned a pull request for go-losetup

go.mod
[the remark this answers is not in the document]
sjoerdsimons: Just release 0.0.11 :)

sjoerdsimons:
nggg; i pushed main to the source repo before doing a re-base ; but helpfully that caused github to close the PR and thus revoking my permission to push to the source branch of an open PR :see_no_evil:  

@julien-massot could you repush, then we can re-open the PR :) Fwiw this is why it's better to create a named branch in a fork when you do changes, so there is no confusion as to what `main` refers to

sjoerdsimons:
a

julien-massot:
Hi @sjoerdsimons I repushed my changes, can you re-open the PR? I will choose a better branch name next time.
Thanks

sjoerdsimons:
> Hi @sjoerdsimons I repushed my changes, can you re-open the PR? I will choose a better branch name next time. Thanks

Done thanks; can you update it to use the new  fakemachine version and drop your hack :)

julien-massot:
> Done thanks; can you update it to use the new fakemachine version and drop your hack :)
Done update to fakemachine release 0.0.11

sjoerdsimons (approved):

go.sum
```diff
@@ -9,37 +9,36 @@ github.com/docker/go-units v0.5.0 h1:69rxXcBk27SvSaaxTtLh/8llcHD8vYHT7WSdRZ/jvr4
 github.com/docker/go-units v0.5.0/go.mod h1:fgPhTUdO+D/Jk86RDLlptpiXQzgHJF7gydDDbaIK4Dk=
 github.com/freddierice/go-losetup/v2 v2.0.1 h1:wPDx/Elu9nDV8y/CvIbEDz5Xi5Zo80y4h7MKbi3XaAI=
 github.com/freddierice/go-losetup/v2 v2.0.1/go.mod h1:TEyBrvlOelsPEhfWD5rutNXDmUszBXuFnwT1kIQF4J8=
-github.com/go-debos/fakemachine v0.0.10 h1:z1ghnWFvQXGOWUIhTVINVmuUq4/idUBcBwF33vXdIek=
-github.com/go-debos/fakemachine v0.0.10/go.mod h1:ebPDs2GaUrsUtEdnlOEJrsAOg6EZBfFQwafMp/XRKrM=
+github.com/go-debos/fakemachine v0.0.11 h1:pXDtOEOjzeK2b3YnrcOx+hJp1B3rbRmxy8v+iD4H6lQ=
```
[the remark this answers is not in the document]
sjoerdsimons: Seems dependabot needed some prodding as it thought we didn't like it anymore. Ofcourse not a blocker for this PR :)

nbuchwitz (approved):
LGTM

obbardc (approved):

PR #539, opened 2025-03-03
Fix mmdebstrap action

iwamatsu:
The mmdebstrap process fails because the format contains single quotation.
To fix this, remove single quotation (') from the append strings of dpkgopt and aptopt.

Debos yaml file:
```
architecture: amd64

actions:
  - action: mmdebstrap
    suite: sid
    variant: minbase
    merged-usr: true
    dpkg-opts:
      - path-exclude=/usr/share/man/*
```

Error message:
```
[...]
2025/03/03 06:59:12 Mmdebstrap | dpkg: error: configuration error: /etc/dpkg/dpkg.cfg.d/99mmdebstrap:1: unknown option ''
[...]
```

From debug message of mmdebstrap:
```
[...]
D: 1 3180 content of /etc/dpkg/dpkg.cfg.d/99mmdebstrap:
'path-exclude=/usr/share/man/*' <--here
apt 2.9.30 (amd64)
Supported modules:
*Ver: Standard .deb
[...]
```

### DIFF
--- a/tests/debian/test.yaml
+++ b/tests/debian/test.yaml
@@ -9,6 +9,12 @@ actions:
     suite: {{ $suite }}
     variant: minbase
     merged-usr: true
+{{- if eq $tool "mmdebstrap" }}
+    dpkg-opts:
+      - path-exclude=/usr/share/man/*
+    apt-opts:
+      - Apt::Install-Recommends "true"
+{{- end }}
 
   - action: apt
     description: Install some base packages
--- a/actions/mmdebstrap_action.go
+++ b/actions/mmdebstrap_action.go
@@ -163,13 +163,13 @@ func (d *MmdebstrapAction) Run(context *debos.DebosContext) error {
 
 	if d.DpkgOpts != nil {
 		for _, opt := range d.DpkgOpts {
-			cmdline = append(cmdline, fmt.Sprintf("--dpkgopt='%s'", opt))
+			cmdline = append(cmdline, fmt.Sprintf("--dpkgopt=%s", opt))
 		}
 	}
 
 	if d.AptOpts != nil {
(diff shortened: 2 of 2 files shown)


iwamatsu:
@nbuchwitz Could you check this PR?

nbuchwitz:
Hm looks like something is double-escaping the string. I've tested your patches and can confirm that this fixes the issue. Thanks for also adding tests for this! -> LGTM

nbuchwitz (approved):

nbuchwitz:
@sjoerdsimons unfortunately this slipped trough in my initial MR. This should go into the release too

obbardc:
@sjoerdsimons looks good to me. I will let you review too

tests/debian/test.yaml
```diff
@@ -9,6 +9,12 @@ actions:
     suite: {{ $suite }}
     variant: minbase
     merged-usr: true
+{{- if eq $tool "mmdebstrap" }}
+    dpkg-opts:
+      - path-exclude=/usr/share/man/*
+    apt-opts:
+      - Apt::Install-Recommends "true"
+{{- end }}
```
sjoerdsimons, line 17: Ideally for a more complete test you should verify if these options were really applied/ended up in the expected config files

sjoerdsimons:
> @sjoerdsimons looks good to me. I will let you review too

thanks looks like we'll need to fix the arch build to get the pipeline green against first

tests/debian/test.yaml
sjoerdsimons: @iwamatsu I'll merge this now so we don't block a potential next release; but would be great if you could do a PR later on with a few more tests for this functionality

PR #549, opened 2025-03-21
Add configurable default chroot method

Witcher01:
By default debos runs chroot commands using systemd-nspawn. This hasn't been configurable. In order for systems that don't have systemd-nspawn to be able to build images with recipes using the chroot keyword for run actions using debos, the default chroot method must be configurable. Since systemd-nspawn is an improved version of chroot, just using a chroot instead should work fine.

Expose a command line option "--default-chroot-method" that allows setting the default method to one that isn't systemd-nspawn.

### DIFF
--- a/cmd/debos/debos.go
+++ b/cmd/debos/debos.go
@@ -50,37 +50,36 @@ func warnLocalhost(variable string, value string) {
 		    Consider using an address that is valid on your network.`
 
 	if strings.Contains(value, "localhost") ||
-	   strings.Contains(value, "127.0.0.1") ||
-	   strings.Contains(value, "::1") {
+		strings.Contains(value, "127.0.0.1") ||
+		strings.Contains(value, "::1") {
 		log.Printf(message, variable)
 	}
 }
 
-
 func main() {
-	context := debos.DebosContext { &debos.CommonContext{}, "", "", 512 }
 	var options struct {
-		Backend       string            `short:"b" long:"fakemachine-backend" description:"Fakemachine backend to use" default:"auto"`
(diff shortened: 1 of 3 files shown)


cmd/debos/debos.go
```diff
@@ -122,6 +124,24 @@ func main() {
 		return
 	}
 
+
+	var chrootMethod debos.ChrootEnterMethod
+	switch options.ChrootMethod {
```
obbardc, line 129: code repetition here from https://github.com/go-debos/debos/blob/main/commands.go#L18; wonder if there's anything that can be done about that ? or maybe a function in that file e.g. `ChrootMethodFromString` ?

commands.go
```diff
@@ -70,7 +71,7 @@ func (w *commandWrapper) flush() {
 }
 
 func NewChrootCommandForContext(context DebosContext) Command {
```
obbardc, line 73: wonder if we should detect if the chrootmethod is supported here and error if not ?

obbardc (changes requested):

sjoerdsimons:
I don't really want to have this as an option as it does change the behaviour of debos for  anything that runs inside a chroot. The chroot chroot method really is only meant to support `debootstrap` specifically due to debootstrap trying to create device nodes.

The systemd tools don't rely on using systemd as pid 1; So nspawn could be packaged for alpine in principle (looks like systemd-boot is for example).

In general fakemachine does assume the host system has systemd available for its functionality (doesn't need to be pid 1 on the host); So packaging debos for systems without systemd will limit its functionality quite a bit

PR #550, opened 2025-04-11
actions: pack/unpack: Add support for zstd compression

julien-massot:
Debian Bookworm includes a version of tar with native support for Zstandard compression.
Add mapping to enable zstd in pack and unpack operations.


obbardc (approved):

actions/pack_action.go
```diff
@@ -77,17 +78,18 @@ func (pf *PackAction) Run(context *debos.DebosContext) error {
...
 	command := []string{"tar"}
+	command = append(command, "cf")
+	command = append(command, outfile)
+
 	if usePigz == true {
-		command = append(command, "cf")
+		command = append(command, "--use-compress-program=pigz")
 	} else {
-		command = append(command, "cf" + tarOpts[pf.Compression])
+		command = append(command, "--" + tarOpts[pf.Compression])
```
julien-massot, line 87: may break none case will check

actions/pack_action.go
```diff
@@ -32,9 +32,10 @@ import (
 )
 
 var tarOpts = map[string]string{
-	"gz":    "z",
-	"bzip2": "j",
-	"xz":    "J",
+	"gz":    "gzip",
+	"bzip2": "bzip2",
+	"xz":    "xz",
+	"zstd":  "zstd",
```
sjoerdsimons, line 38: any reason to not make the options be e.g. `--gzip` to be more in line with the unpackTarOpts in archive?

actions/pack_action.go
```diff
@@ -77,16 +78,14 @@ func (pf *PackAction) Run(context *debos.DebosContext) error {
...
-	} else {
-		command = append(command, "cf" + tarOpts[pf.Compression])
-	}
+	command = append(command, "cf")
 	command = append(command, outfile)
 	command = append(command, "--xattrs")
 	command = append(command, "--xattrs-include=*.*")
 	if usePigz == true {
 		command = append(command, "--use-compress-program=pigz")
+	} else if pf.Compression != "none" {
```
sjoerdsimons, line 87: rather then checking for None; maybe check if `tarOpts[pf.Compression] != ""` as that's what you're really after

actions/pack_action.go
```diff
@@ -77,16 +78,14 @@ func (pf *PackAction) Run(context *debos.DebosContext) error {
...
-		command = append(command, "cf" + tarOpts[pf.Compression])
-	}
+	command = append(command, "cf")
 	command = append(command, outfile)
 	command = append(command, "--xattrs")
 	command = append(command, "--xattrs-include=*.*")
 	if usePigz == true {
 		command = append(command, "--use-compress-program=pigz")
+	} else {
+		command = append(command, tarOpts[pf.Compression])
```
sjoerdsimons, line 88: Shouldn't there be a check for an empty  tarOpts (e.g. when compression is `none`)

sjoerdsimons (approved):

PR #554, opened 2025-05-28
partitioning: Add partattrs property to partition

daissi:
This PR is based on https://github.com/go-debos/debos/pull/490 but adds support of bits 0 to 2 using their bit names.

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -59,6 +59,7 @@ a 32 bits hexadecimal number (e.g. '1234ABCD' without any dash separator).
 	   fsck: bool
 	   fsuuid: string
 	   partuuid: string
+	   partattrs: list of partition attribute bits to set
 
 Mandatory properties:
 
@@ -101,6 +102,16 @@ for partition.
 
 - flags -- list of additional flags for partition compatible with parted(8)
 'set' command.
+
+- partattrs -- list of GPT partition attribute bits to set, as defined in
+https://uefi.org/specs/UEFI/2.10/05_GUID_Partition_Table_Format.html#defined-gpt-partition-entry-attributes.
+Bit 0: "Required Partition", bit 1: "No Block IO Protocol", bit 2: "Legacy BIOS
+Bootable". Bits 3-47 are reserved. Bits 48 - 63 are GUID specific. For example,
+ChromeOS Kernel partitions (GUID=fe3a2a5d-4f32-41a7-b725-accc3285a309) use bit
+56 for "successful boot" and bits 48-51 for "priority", where 0 means not
+bootable, thus bits 56 and 48 need to be set through this property in order to
+be able to boot a ChromeOS Kernel partition on a Chromebook, like so:
+'partattrs: [56, 48]'.
 
 - fsck -- if set to `false` -- then set fs_passno (man fstab) to 0 meaning no filesystem
 checks in boot time. By default is set to `true` allowing checks on boot.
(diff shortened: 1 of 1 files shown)


daissi:
cc @obbardc

obbardc:
@daissi unrelated; but can you look at pushing to existing PRs rather than creating new ones ? I think this is an option for maintainers and would reduce the amount of churn.

daissi:
> @daissi unrelated; but can you look at pushing to existing PRs rather than creating new ones ? I think this is an option for maintainers and would reduce the amount of churn.

That would be a  lot cleaner indeed, but when I tried for the previous one, I didn't have permission. I need write access to `go-debos/debos` to push in existing PRS. Which makes sens to avoid people pushing in repo of other users :)

daissi:
The pipeline fails on Bookworm because `slices`  was added [in Go 1.21](https://tip.golang.org/doc/go1.21#slices) whereas in Bookworm we have Go 1.19. Thus, should we still keep supporting the old one in Bookworm or is it okay to bump the minimum required Go version? @sjoerdsimons what is your opinion?

sjoerdsimons:
> The pipeline fails on Bookworm because `slices` was added [in Go 1.21](https://tip.golang.org/doc/go1.21#slices) whereas in Bookworm we have Go 1.19. Thus, should we still keep supporting the old one in Bookworm or is it okay to bump the minimum required Go version? @sjoerdsimons what is your opinion?

Given trixie is almost out; i'd be happy to bump go to the version that ships

actions/image_partition_action.go
```diff
@@ -102,6 +103,19 @@ for partition.
 - flags -- list of additional flags for partition compatible with parted(8)
 'set' command.
 
+- partattrs -- list of GPT partition attribute bits to set, as defined in
+https://uefi.org/specs/UEFI/2.10/05_GUID_Partition_Table_Format.html#defined-gpt-partition-entry-attributes.
+Bit 0: "Required Partition", bit 1: "No Block IO Protocol", bit 2: "Legacy BIOS
+Bootable". Bits 0-2 have to be passed using their bit names i.e. (RequiredPartition,
+NoBlockIOProtocol or LegacyBIOSBootable) and not as bit numbers due to a
+libfdisk's limitation. Bits 3-47 are reserved. Bits 48-63 are GUID specific.
```
sjoerdsimons: we can do the conversion from the number to the name as required by libfdisk so we don't end up pushing limitations back to the user though.

sjoerdsimons:
> > > @daissi unrelated; but can you look at pushing to existing PRs rather than creating new ones ? I think this is an option for maintainers and would reduce the amount of churn.
> > 
> > 
> > That would be a lot cleaner indeed, but when I tried for the previous one, I didn't have permission. I need write access to `go-debos/debos` to push in existing PRs. Which makes sense to avoid people pushing in repo of other users :)
> 
> can you ask @sjoerdsimons if he agrees ? I think it would make sense to add you.

tbh pushing back to a existing PR is a PITA as a maintainer (i managed to get it wrong a few times) - and only works if the PR opener has allowed it (which is the default but still).  So i'm not sure that actually is a better workflow

THat said i don't mind adding @daissi to the team

actions/image_partition_action.go
daissi: Fixed with 63e05fe1ea4e9b4108157951d4da3115939f7896

daissi:
Bumping to Trixie is no longer required since I removed the use of `slices`.

sjoerdsimons (approved):

PR #555, opened 2025-06-19
Refactor error handling

daissi:
This PR is based on https://github.com/go-debos/debos/pull/303 and tries to address remaining feedback.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -149,6 +149,11 @@ jobs:
           docker compose -f docker/unit-tests.test.yml \
             up --exit-code-from=sut
 
+      - name: Run exitcode tests
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -149,6 +149,11 @@ jobs:
           docker compose -f docker/unit-tests.test.yml \
             up --exit-code-from=sut
 
+      - name: Run exitcode tests
+        run: |
+          docker compose -f docker/exitcode-test.yml \
```
obbardc, line 154: the filename is really `exitcode-test.yaml` which causes this test to fail.
suggest renaming the file `yml` to match others.

daissi: Oups, fixed! Thank you.

sjoerdsimons (approved):

Issue #426, opened 2023-08-09
Create artifactdir if it does not exist


obbardc:
We should attempt to create artifactdir if it is passed to debos and doesn't exist.

Currently it will fails with something similar to:
```
obbardc@obbardc-t14:~/projects/debos/debos-recipes$ debos --artifactdir=out rpi64/debimage-rpi64.yaml 
2023/08/09 12:34:00 Action `image-partition` failed at stage PreMachine, error: open /home/obbardc/projects/debos/debos-recipes/out/debian-rpi64.img: no such file or directory
```

PR #429, opened 2023-08-11
Check if artifactdir exists early

obbardc:
Currently debos fails if an action attempts to use artifactdir and it doesn't exist on the host:

  $ debos --artifactdir=out rpi64/debimage-rpi64.yaml
  Action `image-partition` failed at stage PreMachine, error: open /home/obbardc/projects/debos/debos-recipes/out/debian-rpi64.img: no such file or directory

Check that artifactdir exists early and return an error in case that it does not exist.

Closes: #426

PR #556, opened 2025-06-26
Check if artifactdir exists early

daissi:
Currently debos fails if an action attempts to use artifactdir and it doesn't exist on the host:

  $ debos --artifactdir=out rpi64/debimage-rpi64.yaml
  Action `image-partition` failed at stage PreMachine, error: open /home/obbardc/projects/debos/debos-recipes/out/debian-rpi64.img: no such file or directory

Check that artifactdir exists early and return an error in case that it does not exist.

Closes: #426

This PR is based on https://github.com/go-debos/debos/pull/429 but using `context.State` to track errors since https://github.com/go-debos/debos/pull/555 is now merged.

### DIFF
--- a/cmd/debos/debos.go
+++ b/cmd/debos/debos.go
@@ -238,6 +238,11 @@ func main() {
 		context.Artifactdir, _ = os.Getwd()
 	}
 	context.Artifactdir = debos.CleanPath(context.Artifactdir)
+	if dirInfo, err := os.Stat(context.Artifactdir); err != nil || !dirInfo.IsDir() {
+		log.Printf("Artifact Directory %s does not exist or is not a directory\n", context.Artifactdir)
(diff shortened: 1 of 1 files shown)


cmd/debos/debos.go
```diff
@@ -238,6 +238,11 @@ func main() {
 		context.Artifactdir, _ = os.Getwd()
 	}
 	context.Artifactdir = debos.CleanPath(context.Artifactdir)
+	if dirInfo, err := os.Stat(context.Artifactdir); err != nil || !dirInfo.IsDir() {
+		log.Printf("artifactdir does not exist or is not a directory")
```
obbardc:
```suggestion
		log.Printf("Artifact Directory %s does not exist or is not a directory\n", context.Artifactdir)
```

possibly

PR #557, opened 2025-07-27
Add qemu mapping for sh4 architecture

cepawiel:
Was getting an error about not knowing the correct qemu executable. Testing using the SH4 Debian Port (http://deb.debian.org/debian-ports/)

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -327,6 +327,10 @@ func newQemuHelper(c Command) (*qemuHelper, error) {
 		if runtime.GOARCH != "amd64" {
 			q.qemusrc = "/usr/bin/qemu-x86_64-static"
 		}
+	case "sh4":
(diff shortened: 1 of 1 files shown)


obbardc:
Looks good to me for the debos side. Did you test with [fakemachine](https://github.com/go-debos/fakemachine/) or just `--disable-fakemachine` ?
Some things will probably need to be fixed in fakemachine I guess, since it is only built for amd64 and arm64.

obbardc:
note: once merged, we will need to build debian package for sh4 arch.

obbardc (approved):
OK

obbardc:
ok, i was wrong with last comment. this has nothing to do with deb package as its the arch you choose to build at runtime. oops.

thanks for contribution @cepawiel

PR #571, opened 2025-08-09
Add debos logo

daissi:
Add a debos logo drawn by ChatGPT in PNG and converted in SVG using Inkscape.

There must be a way to get better ones.


obbardc:
I like it, but we should use it for inspiration for a proper logo as I am not sure on reuse licence of AI generated images?

daissi:
Based on https://help.openai.com/en/articles/6783457-what-is-chatgpt#h_4fb39a00bd, it should be fine. But the question remain for the use of Gopher which seems to be under [CC-BY-4.0](https://go.dev/brand). Anyway you are right, let's try to draw a better one ourselves.

obbardc:
I really do like the logo concept! I think next steps you should ask Mark F to ask his graphic designer to come up with a proper logo & mark as a draft until then :-)

obbardc (changes requested):
The SVG isn't a real vector - it is a PNG image embedded inside an SVG. Might as well remove it :-).

sjoerdsimons:
fwiw unsure about having the go gopher that prominent though; I don't think that's a key aspect of debos.. And yeah something not AI created would be nice

daissi:
I also had doubts about the Gopher, so let's try some other proposals:
- with a crane:
ChatGPT Image Aug 12, 2025, 02_06_27 PM

- with a cowboy (no it's not Woody :-P)
ChatGPT Image Aug 12, 2025, 02_25_45 PM

I'll try to reach a consensus before asking a real artist to do a drawing :)

Other ideas on how to represent a builder in a nice way? :thinking:

daissi:
I tried it with an astronaut who looked like Buzz, but the result was not satisfactory.

daissi:
And now with a cowboy builder :smile: : 
ChatGPT Image Aug 14, 2025, 09_20_16 AM

obbardc:
other characters welcome :P

obbardc:
@daissi did you ask Mark F about this ? I think it would be nice to get a proper SVG icon in :-)

daissi:
@obbardc Not yet, but I thought/tried something simpler:

debos

PR #576, opened 2025-08-11
Remove user-mode-linux support from container & remove related tests 

obbardc:
Since we plan to remove user-mode-linux support from fakemachine (see https://github.com/go-debos/fakemachine/issues/193); don't install the packages in the container & don't run the user-mode-linux tests in CI.

Depends on https://github.com/go-debos/fakemachine/pull/244

### DIFF
--- a/tests/exit_test/exit_test.sh
+++ b/tests/exit_test/exit_test.sh
@@ -54,20 +54,20 @@ fi
 expect_success debos --help
 expect_failure debos --not-a-valid-option
 expect_failure debos
-expect_failure debos good.yaml good.yaml
+expect_failure debos --fakemachine-backend=qemu good.yaml good.yaml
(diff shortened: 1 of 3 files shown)


tests/exit_test/exit_test.sh
```diff
@@ -54,20 +54,20 @@ fi
 expect_success debos --help
 expect_failure debos --not-a-valid-option
 expect_failure debos
-expect_failure debos good.yaml good.yaml
-expect_failure debos --disable-fakemachine --fakemachine-backend=uml good.yaml
-expect_failure debos non-existent-file.yaml
-expect_failure debos garbled.yaml
+expect_failure debos --fakemachine-backend=qemu good.yaml good.yaml
```
obbardc, line 57: This commit can go away once https://github.com/go-debos/fakemachine/pull/244 is merged.

obbardc:
Actually; this makes sense to be part of #521 so closing.

PR #585, opened 2025-08-26
feat: Optional sha256sum for download action

lool:
Setting this property will check the SHA256 sum of the downloaded file
and delete it on mismatch.


lool:
This is inspired by Yocto's ability to declare expected SHA256 sums next to upstream URLs. This can of course we achieved with the `run` action like [here](https://github.com/qualcomm-linux/qcom-deb-images/commit/953683028e9a2f67acffcce93222abed7c78688c) but it would be prettier to just have a property for this.

actions/download_action.go
```diff
@@ -29,14 +29,21 @@ See the 'Unpack' action for more information.
 
 - compression -- optional hint for unpack allowing to use proper compression method.
 See the 'Unpack' action for more information.
+
+- sha256sum -- optional expected SHA256 sum of the downloaded file
```
sjoerdsimons: would be wroth noting that this is expected to be a hex string of 64 characters (i assume it can't use the 0x prefix but is not case-sensitive?)

actions/download_action.go
```diff
@@ -147,6 +164,19 @@ func (d *DownloadAction) Run(context *debos.DebosContext) error {
 		return fmt.Errorf("Unsupported URL is provided: '%s'", url.String())
 	}
 
+	if len(d.Sha256sum) > 0 {
+		data, err := ioutil.ReadFile(filename)
+		if err != nil {
+			return fmt.Errorf("Failed to read download file %s: %v", filename, err)
+		}
+		hash := sha256.Sum256(data)
+		actualSha256sum := hex.EncodeToString(hash[:])
```
sjoerdsimons: Can be later/nice to have; But i wonder if it makes sense now that we have that code to always print the sha256 sum, so it can be traced even if it's not specified

sjoerdsimons:
Seems like a quite worthwhile features to have indeed; Would be cool to also have a `download` test be part of the recipe test, but not a blocker (and raises the question what's a good stable test target to download)

actions/download_action.go
lool: Indeed, that's not case sensitive; I've clarified the description

We could strip a leading 0x, but I personally would prefer if we stick to a single way of writing these

actions/download_action.go
lool: Good idea; changed!

lool:
> Seems like a quite worthwhile features to have indeed; Would be cool to also have a `download` test be part of the recipe test, but not a blocker (and raises the question what's a good stable test target to download)

Not sure what you mean, the tests setup a mini HTTP server and do an actual download of contents that we control?

sjoerdsimons:
> > Seems like a quite worthwhile features to have indeed; Would be cool to also have a `download` test be part of the recipe test, but not a blocker (and raises the question what's a good stable test target to download)
> 
> Not sure what you mean, the tests setup a mini HTTP server and do an actual download of contents that we control?

Maybe :) But yeah it's either setting up a local http server to make sure we control everything *or* downloading some known content from the internet (makes the test simpler, but potentially more fragile).

actions/download_action.go
sjoerdsimons: Oh yeah; no need to allow a leading 0x , nobody writes hashes with them :) It was all just about clarifying the documentation string :)

actions/download_action.go
```diff
@@ -147,6 +165,21 @@ func (d *DownloadAction) Run(context *debos.DebosContext) error {
 		return fmt.Errorf("Unsupported URL is provided: '%s'", url.String())
 	}
 
+	data, err := ioutil.ReadFile(filename)
```
sjoerdsimons: Didn't catch this yesterday; while we normally expect the download action to be fore small files; people can do silly things, so just reading all of it to memory is not ideal. Better to do a streaming hash. The Hash interface does have an ioWriter, so you should be able to use some of the io copy utilities for that.

lool: That's a really good point, in fact I did hit memory pressure situations in building debos recipes; how does this now look?

lool:
Rebased as there were some conflicts with the staticcheck branch that landed

PR #598, opened 2025-10-07
Use Verify to check required commands are present

### LINKED ISSUE #280: actions: check for presence of required commands in Verify()

daissi:
For each action, check that as many commands it needs are available as possible. This can't be meaningfully extended to commands that run in the chroot, or might run in the chroot.

This follows up from comments on https://github.com/go-debos/debos/pull/276 which were captured as issue https://github.com/go-debos/debos/issues/280.

Closes: https://github.com/go-debos/debos/issues/280

Based on: https://github.com/go-debos/debos/pull/283

### DIFF
--- a/action.go
+++ b/action.go
@@ -54,8 +54,12 @@ func (c *Context) Origin(o string) (string, bool) {
 }
 
 type Action interface {
+	// Verify runs first
 	/* FIXME verify should probably be prepare or somesuch */
 	Verify(context *Context) error
+	// CheckEnvironment runs in the environment that Run will execute in,
+	// prior to Run commencing
+	CheckEnvironment(context *Context) error
 	PreMachine(context *Context, m *fakemachine.Machine, args *[]string) error
 	PreNoMachine(context *Context) error
 	Run(context *Context) error
@@ -74,7 +78,8 @@ type BaseAction struct {
 	Description string
 }
 
-func (b *BaseAction) Verify(_ *Context) error { return nil }
+func (b *BaseAction) Verify(_ *Context) error           { return nil }
+func (b *BaseAction) CheckEnvironment(_ *Context) error { return nil }
 func (b *BaseAction) PreMachine(_ *Context,
 	_ *fakemachine.Machine,
 	_ *[]string) error {
--- a/actions/debootstrap_action.go
+++ b/actions/debootstrap_action.go
@@ -127,6 +127,15 @@ func (d *DebootstrapAction) Verify(context *debos.Context) error {
 	return nil
 }
 
+func (d *DebootstrapAction) CheckEnvironment(_ *debos.Context) error {
+	// Check that debootstrap is available
+	cmd := debos.Command{}
+	if err := cmd.CheckExecutableExists("debootstrap"); err != nil {
+		return err
+	}
+	return nil
+}
+
 func (d *DebootstrapAction) PreMachine(context *debos.Context, m *fakemachine.Machine, _ *[]string) error {
 	mounts := d.listOptionFiles(context)
 
--- a/cmd/debos/debos.go
+++ b/cmd/debos/debos.go
@@ -300,7 +300,36 @@ func main() {
 		return
 	}
 
+	if !runInFakeMachine && !fakemachine.InMachine() {
+		// We will run on the host
+		for _, a := range r.Actions {
+			// Stack PostMachineCleanup methods
+			defer func(action debos.Action) {
+				_ = action.PostMachineCleanup(&context)
+			}(a)
+
+			err = a.PreNoMachine(&context)
+			if handleError(&context, err, a, "PreNoMachine") {
+				return
+			}
+		}
+	}
+
+	if !runInFakeMachine {
+		// Either we're on the host and intend to run on it,
+		// or we're in the fake machine and intend to run on
+		// it. In the former case, PreNoMachine has just run;
+		// in the latter PreMachine has already run.
+		for _, a := range r.Actions {
+			err = a.CheckEnvironment(&context)
+			if handleError(&context, err, a, "CheckEnvironment") {
+				return
+			}
+		}
+	}
+
 	if runInFakeMachine {
+		// We're on the host and intend to run on the fakemachine
 		var args []string
 
 		if options.Memory == "" {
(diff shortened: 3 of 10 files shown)


obbardc (approved):
looks fine to me

commands.go
```diff
@@ -289,6 +289,21 @@ func (cmd Command) Run(label string, cmdline ...string) error {
 	return nil
 }
 
+func (cmd Command) Run(label string, cmdline ...string) error {
+	w := newCommandWrapper(label, true)
+	defer w.flush()
+	return cmd.run(label, w, cmdline)
+}
+
+func (cmd Command) CheckExists(label string, cmdline ...string) error {
```
obbardc: perhaps we should just check if the command exists in the path rather than running the command ?
if the command fails later (e.g. built for wrong arch), it's fine, since that'll be picked up later.
not sure what benefit running these commands has over checking if it exists in $path.
that would allow most of the cruft to go away in this file.

obbardc (changes requested):
nitpick

commands.go
daissi: I simplified this part.

actions/debootstrap_action.go
```diff
@@ -127,6 +127,15 @@ func (d *DebootstrapAction) Verify(context *debos.Context) error {
 	return nil
 }
 
+func (d *DebootstrapAction) CheckEnv(_ *debos.Context) error {
+	// Check that debootstrap is available
+	cmd := debos.Command{}
+	if err := cmd.CheckExists("Debootstrap", "debootstrap", "--version"); err != nil {
```
obbardc: why call `debootstrap --version` ?
I think `cmd.CheckExecutableExists("debootstrap")` is a better API

(same for all other consumers)

action.go
```diff
@@ -74,7 +78,8 @@ type BaseAction struct {
 	Description string
 }
 
-func (b *BaseAction) Verify(_ *Context) error { return nil }
+func (b *BaseAction) Verify(_ *Context) error   { return nil }
+func (b *BaseAction) CheckEnv(_ *Context) error { return nil }
```
obbardc:
```suggestion
func (b *BaseAction) CheckEnvironment(_ *Context) error { return nil }
```

Environment seems better to me

cmd/debos/debos.go
```diff
@@ -300,7 +300,36 @@ func main() {
 		return
 	}
 
+	if !runInFakeMachine && !fakemachine.InMachine() {
```
obbardc, line 303: are these few lines a merge error ? they dont look like they belong here

cmd/debos/debos.go
```diff
@@ -300,7 +300,36 @@ func main() {
...
+
+			err = a.PreNoMachine(&context)
+			if handleError(&context, err, a, "PreNoMachine") {
+				return
+			}
+		}
+	}
+
+	if !runInFakeMachine {
+		// Either we're on the host and intend to run on it,
```
obbardc, line 319: can you just call `CheckEnvironment` independently of all of this ? in same area as `Verify` ?

commands.go
obbardc: I think `CheckExecutableExists` sounds better name

go.mod
```diff
@@ -24,5 +24,6 @@ require (
 	github.com/surma/gocpio v1.1.0 // indirect
 	github.com/ulikunitz/xz v0.5.14 // indirect
 	golang.org/x/sys v0.22.0 // indirect
+	golang.org/x/text v0.22.0 // indirect
```
obbardc: what is this for ? sounds like its separate to this change ?

go.sum
```diff
@@ -45,6 +45,8 @@ github.com/ulikunitz/xz v0.5.14/go.mod h1:nbz6k7qbPmH4IRqmfOplQw/tblSgqTqBwxkY0o
 golang.org/x/sys v0.21.0/go.mod h1:/VUhepiaJMQUp4+oa/7Zr1D23ma6VTLIYjOOTFZPUcA=
 golang.org/x/sys v0.22.0 h1:RI27ohtqKCnwULzJLqkv897zojh5/DwS/ENaMzUOaWI=
 golang.org/x/sys v0.22.0/go.mod h1:/VUhepiaJMQUp4+oa/7Zr1D23ma6VTLIYjOOTFZPUcA=
+golang.org/x/text v0.22.0 h1:bofq7m3/HAFvbF51jz3Q9wLg3jkvSPuiZu/pD1XwgtM=
```
obbardc: again, what is this for ? sounds like its separate to this change ?

obbardc (changes requested):

commands.go
daissi: Renamed.

actions/debootstrap_action.go
daissi: Oups fixed

action.go
daissi: Renamed.

go.sum
daissi: Removed

go.mod
daissi: Removed

sjoerdsimons (changes requested):
Need to verify the mmdebstrap action as well

sjoerdsimons:
@daissi is there a reason for this to be a draft?

daissi:
> Need to verify the mmdebstrap action as well

Fixed!

daissi:
> @daissi is there a reason for this to be a draft?

Just because I haven't addressed all the comments, but let's remove the draft status.

obbardc:
@daissi this needs rebasing on main & couple of simple comments actioning :-)

cmd/debos/debos.go
daissi: This comes from the previous PR and discussed/explained here https://github.com/go-debos/debos/pull/283#discussion_r689369941. I think it's still relevant.

cmd/debos/debos.go
daissi: From my understanding this is also required to ensure we are running it from the host and not from the fakemachine.

cmd/debos/debos.go
obbardc: This possibly belongs in its own commit with rationale ?

obbardc:
@daissi gentle ping. Think it is almost ready for merge apart from small comments above :-)

PR #599, opened 2025-10-08
image_partition: add ability to skip "start" and specify relative offsets

daissi:
Based on https://github.com/go-debos/debos/pull/285

Just a rebase for now.

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -786,7 +786,7 @@ func ParseOffset(offset string) (int, string, error) {
 	typ := string(t)
 
 	if err != nil {
-		return 0, "", fmt.Errorf("Can not parse %s to integer", string(v))
+		return 0, "", fmt.Errorf("can not parse %s to integer", string(v))
 	}
 
 	return val, typ, nil
@@ -800,7 +800,7 @@ func CalculateOffset(start, end string) (string, error) {
 		valStart, typeStart, errEnd := ParseOffset(start)
 
 		if typeStart != typeEnd {
-			return "", fmt.Errorf("Relative offset types are not consistent")
+			return "", fmt.Errorf("relative offset types are not consistent")
 		}
 
 		if errStart != nil {
@@ -820,7 +820,7 @@ func ValidPercentage(offset string) error {
 		return err
 	}
 	if typ == "%" && val > 100 {
-		return fmt.Errorf("Size can not exceed 100%")
+		return fmt.Errorf("size can not exceed 100%%")
 	}
 	return nil
 }


actions/image_partition_action.go
```diff
@@ -769,6 +770,61 @@ func (i ImagePartitionAction) PostMachineCleanup(context *debos.Context) error {
...
+		case l >= 'A' && l <= 'Z', l >= 'a' && l <= 'z', l == '%':
+			t = append(t, l)
+		}
+	}
+
+	val, err := strconv.Atoi(string(v))
+	typ := string(t)
+
+	if err != nil {
+		return 0, "", fmt.Errorf("Can not parse %s to integer", string(v))
```
obbardc:
```suggestion
		return 0, "", fmt.Errorf("can not parse %s to integer", string(v))
```
nitpick; lowercase for errors

actions/image_partition_action.go
```diff
@@ -769,6 +770,61 @@ func (i ImagePartitionAction) PostMachineCleanup(context *debos.Context) error {
...
+
+func CalculateOffset(start, end string) (string, error) {
+	/* Do addition if end value uses a '+' prefix */
+	if strings.HasPrefix(end, "+") {
+		end = strings.Split(end, "+")[1]
+		valEnd, typeEnd, errStart := ParseOffset(end)
+		valStart, typeStart, errEnd := ParseOffset(start)
+
+		if typeStart != typeEnd {
+			return "", fmt.Errorf("Relative offset types are not consistent")
```
obbardc:
```suggestion
			return "", fmt.Errorf("relative offset types are not consistent")
```
nitpick: lower for errors

actions/image_partition_action.go
```diff
@@ -769,6 +770,61 @@ func (i ImagePartitionAction) PostMachineCleanup(context *debos.Context) error {
...
+	return end, nil
+}
+
+func ValidPercentage(offset string) error {
+	val, typ, err := ParseOffset(offset)
+	if err != nil {
+		return err
+	}
+	if typ == "%" && val > 100 {
+		return fmt.Errorf("Size can not exceed 100%")
```
obbardc:
```suggestion
		return fmt.Errorf("size can not exceed 100%%")
```
nitpick: lowercase errors && proper escape for `%` character

lool:
I wonder if this should be a "size" rather than a relative offset. This would also for more explicit handling: either this is a size based partition request, or it's fixed offsets (start/end).

sjoerdsimons:
> I wonder if this should be a "size" rather than a relative offset. This would also for more explicit handling: either this is a size based partition request, or it's fixed offsets (start/end).

I like that better indeed; As it's more obvious to the user rather then having the `+`.  There is a bigger topic around improving our partitioning but this seems a reasonable small step to have that should well in the current setup

obbardc:
@daissi do you plan to rebase on main & address comments in the next month or so ? :-)

PR #605, opened 2025-10-25
Docker: Speed up runner builds by avoiding initramfs

sjoerdsimons:
The runner images install a kernel, which pulls in initramfs-tools to create an initramfs. For debos we don't need to have an initramfs, we just need the kernel image. Generating this on the native architecture (amd64) is reasonably fast. However in emulation, when building for arm64, this is *very* slow.

In current github actions this change seems to save about 20 minutes per docker build. For PR triggered runs this (which don't upload the containers) this should save 20 minutes end-to-end (one build on the critical path), while for runs that do it should save about 40 minutes end to end (three builds, of which 2 on the critical path).


obbardc (approved):
This has been on the annoyance list for ages! Thanks!

PR #636, opened 2026-01-09
pacstrap_action: add an option to install extra packages

### LINKED ISSUE #481: pacstrap: add option to install extra packages
Currently we implicitly install only "base". This may be an issue for some Arch-based distros.

Needs to be confirmed first.

daissi:
When building Arch-based rootfs, we sometimes have to install packages early, so using the "pacman" action just after the "pacstrap" action is not suitable for instance if we have an extra keyring.

Fixes #481

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -223,6 +223,8 @@ jobs:
           - backend: kvm
             test: { name: "arch", case: "arch" }
           - backend: kvm
+            test: { name: "arch (pacstrap with extra packages)", case: "arch-pacstrap-extra-packages" }
+          - backend: kvm
             test: { name: "apertis", case: "apertis" }
           - backend: kvm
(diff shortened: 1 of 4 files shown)


daissi:
I am wondering if we should also install the `base` metapackage when the list of packages to install is used. Currently, adding packages into the list no longer installs the `base` package, so we have to manually add it into the list.

obbardc:
> I am wondering if we should also install the `base` metapackage when the list of packages to install is used. Currently, adding packages into the list no longer installs the `base` package, so we have to manually add it into the list.

I think leaving that up to the user with the comment you have around `packages` is fine. But I am open to ideas.

daissi:
I agree, in this way it provides more flexibility if we do not want to install the base metapackage. I have updated the comment to add this clarification.

sjoerdsimons:
Changes look fine to me; @daissi  would be nice to have a test for this as well ;)

PR #642, opened 2026-01-19
docker: Install qemu-user instead of qemu-user-static

obbardc:


obbardc:
This fails with debootstrap actions . Not sure why yet.

obbardc:
Also need to install this package in test-containers / fakemachine.

Issue #23, opened 2017-09-29
action: overlay: add verification either 'source' or 'origin' has been set


d4s:
Source or origin must be set to prevent copying of recipedir

PR #644, opened 2026-01-22
actions: overlay: verify either source or origin set

tobsan:
This check avoids accidental overlaying the recipe directory, which the origin defaults to, in case source isn't set.

Fixes: #23

### DIFF
--- a/actions/overlay_action.go
+++ b/actions/overlay_action.go
@@ -25,6 +25,7 @@ If destination isn't set '/' of the rootfs will be used.
 package actions
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"os"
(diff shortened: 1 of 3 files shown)


obbardc:
Small nitpick: commit message to be "actions: overlay: verify either source or origin set"

actions/overlay_action.go
```diff
@@ -45,6 +46,10 @@ func (overlay *OverlayAction) Verify(context *debos.Context) error {
 		return err
 	}
 
+	if len(overlay.Source) == 0 && len(overlay.Origin) == 0 {
+		return errors.New("'source' and 'origin' properties can't both be empty");
```
obbardc: nitpick: lint fails:
```
  actions/overlay_action.go:50:1: File is not properly formatted (gofmt)
  		return errors.New("'source' and 'origin' properties can't both be empty");
```

remove the `;`

tobsan: The C programmer inside me strikes again. Fixed.

PR #645, opened 2026-01-22
Add mappings between GPT parttype names and UUIDs

### LINKED ISSUE #393: image-partition: Define constants for GPT partition types
We can set GPT partition types based on [The Discoverable Partitions Specification](https://uapi-group.org/specifications/specs/discoverable_partitions_specification/), for a full example see: https://salsa.debian.org/Mobian-team/mobian-recipes/-/blob/master/include/partition-gpt.yaml#L54

We should expose these as constants or provide some kind of helper function to make the syntax easier.

tobsan:
Closes #393

### DIFF
--- a/tests/gptnames/test.yaml
+++ b/tests/gptnames/test.yaml
@@ -1,0 +1,53 @@
+{{- $architecture := or .architecture "amd64"}}
+
+architecture: {{ $architecture }}
+
+actions:
+  - action: image-partition
+    description: Partition the image
+    imagename: test.img
+    imagesize: 4G
+    partitiontype: gpt
+    diskid: 12345678-1234-1234-1234-123456789012
+    mountpoints:
+      - mountpoint: /
+        partition: system
+      - mountpoint: /usr
+        partition: usr
+    partitions:
+      - name: system
+        fs: ext4
+        start: 0%
+        end: 1G
+        partuuid: 87654321-1234-5678-9012-345678901234
+        parttype: root
+      - name: usr
+        fs: ext4
+        start: 1G
+        end: 2G
+        partuuid: 87654321-1234-5678-9012-345678901235
+        parttype: usr
+      - name: root_verity
+        fs: ext4
+        start: 2G
+        end: 3G
+        partuuid: 87654321-1234-5678-9012-345678901236
+        parttype: root-verity
+      - name: other_part
+        fs: ext4
+        start: 3G
+        end: 4G
+        partuuid: 87654321-1234-5678-9012-345678901237
+        # Explicitly setting a type UUID also works
+        parttype: 4982F05A-8DD2-44D3-83C8-03EEA8B3478F
+
+  - action: run
+    chroot: false
+    command: >
+      cd ${ARTIFACTDIR};
+      sfdisk -J test.img | tee ${RECIPEDIR}/actual.json
+
+  - action: run
+    description: Compare expected and actual
+    chroot: false
+    command: bash -c 'diff -u <(jq . ${RECIPEDIR}/expected-{{$architecture}}.json) <(jq . ${RECIPEDIR}/actual.json)'
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -216,7 +216,8 @@ jobs:
           - { name: "templating", case: "templating", variables: " -t escaped:\\$ba\\'d\\$gers\\ snakes" }
           - { name: "partitioning", case: "partitioning" }
           - { name: "msdos partitioning", case: "msdos" }
-          - { name: "GPT partition names", case: "gptnames" }
+          - { name: "GPT partition names (amd64)", case: "gptnames", variables: "-t architecture:amd64" }
+          - { name: "GPT partition names (arm64)", case: "gptnames", variables: "-t architecture:arm64" }
           - { name: "debian (amd64, debootstrap)", case: "debian", variables: "-t architecture:amd64" }
           - { name: "debian (amd64, mmdebstrap)", case: "debian", variables: "-t architecture:amd64 -t tool:mmdebstrap" }
           - { name: "raw", case: "raw" }
@@ -225,6 +226,10 @@ jobs:
             test: { name: "partitioning", case: "partitioning" }
           - backend: nofakemachine
             test: { name: "msdos partitioning", case: "msdos" }
+          - backend: nofakemachine
+            test: { name: "GPT partition names (amd64)", case: "gptnames" }
+          - backend: nofakemachine
+            test: { name: "GPT partition names (arm64)", case: "gptnames" }
           - backend: nofakemachine
             test: { name: "raw", case: "raw" }
         include:
--- a/tests/exit_test/exit_test.sh
+++ b/tests/exit_test/exit_test.sh
@@ -72,6 +72,7 @@ expect_failure debos overlay-missing-source.yaml
 expect_failure debos overlay-no-source.yaml
 expect_failure debos missing-mountpoint.yaml
 expect_failure debos missing-partition.yaml
+expect_failure debos partition-bad-gpt.yaml
 expect_failure rename_command NOT_DEBOS debos good.yaml
 
 expect_failure $SUDO debos missing-file.yaml --disable-fakemachine
--- a/tests/exit_test/partition-bad-gpt.yaml
+++ b/tests/exit_test/partition-bad-gpt.yaml
@@ -1,0 +1,31 @@
+architecture: amd64
+
+actions:
+  - action: image-partition
+    description: Partition the image
+    imagename: test.img
+    imagesize: 3G
+    partitiontype: gpt
+    diskid: 12345678-1234-1234-1234-123456789012
+    mountpoints:
+      - mountpoint: /
+        partition: system
+      - mountpoint: /usr
+        partition: usr
+    partitions:
+      - name: system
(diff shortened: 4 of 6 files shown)


actions/image_partition_action.go
```diff
@@ -58,6 +58,7 @@ a 32 bits hexadecimal number (e.g. '1234ABCD' without any dash separator).
 		   flags: list of flags
 		   fsck: bool
 		   fsuuid: string
+		   parttype: string
```
obbardc, line 61: would it not make sense to use `partuuid` and check if it is a valid UUID and if not, then check contents of `GPTPartitionTypeUUIDs`?

actions/image_partition_action.go
```diff
@@ -201,6 +202,126 @@ import (
 	"github.com/go-debos/debos"
 )
 
+func findPartTypeUUID(friendlyName string) (string, error) {
+
+	var GPTPartitionTypeUUIDs = map[string]string{
```
tobsan: Tbh, not sure we have to add ALL of these here, but otoh it becomes difficult to decide what goes and what doesn't.

obbardc (changes requested):
small nitpick; also suggest moving the other unrelated commit into a separate PR.

actions/image_partition_action.go
obbardc: I'd be happy with a commented link to the systemd.io webpage & the same list of uuids.

actions/image_partition_action.go
tobsan: I'm not sure that makes sense. But maybe what you mean is to check that the `parttype` is a valid UUID, which currently isn't checked until sfdisk checks it? `partuuid` isn't involved here at all afaik.

actions/image_partition_action.go
```diff
@@ -880,15 +1001,25 @@ func (i *ImagePartitionAction) Verify(_ *debos.Context) error {
...
-			var partTypeLen int
-			switch i.PartitionType {
-			case "gpt":
-				partTypeLen = 36
-			case "msdos":
-				partTypeLen = 2
-			}
-			if len(p.PartType) != partTypeLen {
-				return fmt.Errorf("incorrect partition type for %s, should be %d characters", p.Name, partTypeLen)
+			if i.PartitionType == "gpt" && strings.HasPrefix(p.PartType, "SD_GPT") {
```
sjoerdsimons: why the SD_GPT prefix? that's just the defined in the systemd api, it doesn't mean anything outside of that. Eg. an ESP is not defined by DPS  in itself (https://uapi-group.org/specifications/specs/discoverable_partitions_specification/)   

also for usability it would be nice to be able to use more meaning full names so just ESP (or esp?) for well that and root, usr etc (with it automatically picking the right partition for the architecture).

sjoerdsimons: For naming it would be good to follow e.g. repart.d  Type definition https://www.freedesktop.org/software/systemd/man/latest/repart.d.html

sjoerdsimons (changes requested):
good start; not great usability, see comments ;)

actions/image_partition_action.go
sjoerdsimons: partuuid is the partition identifier, *not* the type; The parttype is well the type, which is a uuid as well for GPT just to confuse you

actions/image_partition_action.go
tobsan: The simple answer is because I just mirrored the UAPI spec. We don't *have* to do that though, and picking up the right arch for those constants that need it would be awesome actually, it didn't really cross my mind.

actions/image_partition_action.go
obbardc: oops, i got confused. sorry @tobsan

actions/image_partition_action.go
tobsan: Added a link and included all (I think) archs that are supported through qemu and exists in the spec.

actions/image_partition_action.go
tobsan: Updated with what I hope is greatly improved usability, at least it looks good in the tests :)

actions/image_partition_action.go
```diff
@@ -879,16 +997,22 @@ func (i *ImagePartitionAction) Verify(_ *debos.Context) error {
...
-			case "gpt":
-				partTypeLen = 36
-			case "msdos":
-				partTypeLen = 2
-			}
-			if len(p.PartType) != partTypeLen {
-				return fmt.Errorf("incorrect partition type for %s, should be %d characters", p.Name, partTypeLen)
+		if len(p.PartType) > 0 {
+			if i.PartitionType == "gpt" {
+				_, err := uuid.Parse(p.PartType)
```
sjoerdsimons, line 1002: from the documentaton of uuid.Parse (https://pkg.go.dev/github.com/google/uuid#Parse): Parse should not be used to validate strings as it parses non-standard encodings as indicated above.

There is a Validate as well ; However that also parses a few different uuid types not just the GUID type, which is what we require and i think the tooling expects.

obbardc: we use uuid.Parse elsewhere in actions/image_partition_action.go; perhaps we need to change the usage of those first :/. @tobsan do you want to handle that (in separate PR) or do you want me to have a go ?

tobsan: Maybe a reasonable way would be to call Validate first. If that succeeds, run Parse and then string on the UUID object, which according to its docs always outputs in the format we (and sfdisk) expects?

tests/gptnames/test.yaml
```diff
@@ -0,0 +1,53 @@
...
+        end: 3G
+        partuuid: 87654321-1234-5678-9012-345678901236
+        parttype: root-verity
+      - name: other_part
+        fs: ext4
+        start: 3G
+        end: 4G
+        partuuid: 87654321-1234-5678-9012-345678901237
+        # Explicitly setting a type UUID also works
+        parttype: 4982F05A-8DD2-44D3-83C8-03EEA8B3478F
```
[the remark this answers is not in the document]
tobsan: Fixed.

actions/image_partition_action.go
```diff
@@ -201,6 +205,120 @@ import (
...
+
+	arch := context.Architecture
+	switch arch {
+	case "armhf", "armel":
+		arch = "arm"
+	}
+
+	submap, ok := GPTPartitionTypeByArch[arch]
+	if !ok {
+		return "", fmt.Errorf("no GPT type identifiers for arch %s", context.Artifactdir)
```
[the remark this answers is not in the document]
tobsan: Fixed

actions/image_partition_action.go
```diff
@@ -201,6 +205,120 @@ import (
...
+	/*
+		Not an exhaustive list, mostly limited by the architectures supported by
+		debos through QEMU, see the UAPI spec for an exhaustive list:
+		https://uapi-group.org/specifications/specs/discoverable_partitions_specification/
+	*/
+	var GPTPartitionTypeByArch = map[string]map[string]string{
+		"arm": {
+			"root":            "69dad710-2ce4-4e3c-b16c-21a1d49abed3",
+			"root-verity":     "7386cdf2-203c-47a9-a498-f2ecce45a2d6",
+			"root-verity-sig": "42b0455f-eb11-491d-98d3-56145ba9d037",
```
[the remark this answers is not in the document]
tobsan: Would've expected a compiler to optimize that, but fixed nonetheless.

actions/image_partition_action.go
```diff
@@ -879,16 +997,22 @@ func (i *ImagePartitionAction) Verify(_ *debos.Context) error {
...
+		if len(p.PartType) > 0 {
+			if i.PartitionType == "gpt" {
+				_, err := uuid.Parse(p.PartType)
+				if err != nil { // Not a valid UUID, but might be a valid short name
+					uuid, err := findPartTypeUUID(context, p.PartType)
+					if err != nil {
+						return fmt.Errorf("GPT partition type \"%s\" for partition \"%s\" is invalid UUID and not a recognized identifier", p.PartType, p.Name)
+					}
+
+					p.PartType = uuid
```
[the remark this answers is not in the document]
tobsan: Fixed

actions/image_partition_action.go
tobsan: @obbardc Sorry, somehow your comment wasn't shown until I refreshed the page. Feel free to break out the UUID utility stuff I did in image-partition in a separate PR, and I'll rebase this one again once that's finished.

obbardc: > Maybe a reasonable way would be to call Validate first. If that succeeds, run Parse and then string on the UUID object, which according to its docs always outputs in the format we (and sfdisk) expects?

sounds like we need a `parseUUID` helper function to do that :-). Happy if you wanna do it in separate PR? Or I can, up to you.

obbardc: OK, I have started a PR for this rework, expect it soonish :-).

obbardc: Had a first stab: https://github.com/go-debos/debos/pull/652

obbardc: Happy to leave this as-is using `uuid.Parse` for this PR; and cleanup in #652 before/after this PR :-)

tobsan: Alright, backpedaled my uuid changes here.

PR #646, opened 2026-01-22
tests: fix minor typo in partition tests

tobsan:


tests/msdos/test.yaml
```diff
@@ -8,7 +8,7 @@ actions:
     partitiontype: msdos
     diskid: deadbeef
     mountpoints:
-      - mounpoint: /
+      - mountpoint: /
```
sjoerdsimons, line 11: woops! Given this; would be great to also have debos verify each mountpoint in the list has well a mountpoint; Tbh i'm surprised this test didn't fail

obbardc: yep, sounds useful

tobsan: Updated PR with validation and tests for this

tobsan: Wouldn't this be something that we could validate (at least the existence of the key) with a schema rather than checking manually?

obbardc: yeah, existence of the key would be best. Not sure how you do that in the yaml library, though. How you've done it here matches the rest of the code :-).

sjoerdsimons: A schema check woudl be great but too much for this PR i think :) 

In this case i'm more wondering how this worked in the first place; as mountpoint is meant to be a mandatory option so the code is probably doing something weird if it's missing at a quick look i see we check for multiple mountpoints, but i guess not if it's nil or "" or somesuch

PR #648, opened 2026-01-23
Migrate to qemu-user-binfmt instead of qemu-user-static 

obbardc:
qemu-user-static has been removed from Debian. Install the replacement
package qemu-user-binfmt instead (which ships qemu-* instead of
qemu-*-static) since we no longer use the static binaries anyway.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -114,7 +114,7 @@ RUN apt-get update && \
         openssh-client \
         parted \
         pkg-config \
-        qemu-user-static \
+        qemu-user-binfmt \
         qemu-utils \
         rsync \
         systemd \


sjoerdsimons:
Just for reference the reason this works is because the F flag for binfmt is now a thing; so it's actually running the qemu binary from the environment even though it's *not* in the mount namespace of the thing it's emulating. It might be nice to mention this in the commit messages to explain why this change is ok now (but we needed the copy dance in the past)

obbardc:
> Just for reference the reason this works is because the F flag for binfmt is now a thing; so it's actually running the qemu binary from the environment even though it's _not_ in the mount namespace of the thing it's emulating. It might be nice to mention this in the commit messages to explain why this change is ok now (but we needed the copy dance in the past)

thanks for the additional context (rather than `we dont need that anymore`), updated commit message to be more clear i hope

PR #649, opened 2026-01-23
Allow specifying subdirs in both pack/unpack actions

### LINKED ISSUE #249: actions: pack: allow specifying path to pack
Sometimes it is useful to generate a tarball of a part of the filesystem as opposed to the whole of it; it’s not currently possible to do from inside debos.

tobsan:
Fixes #249


actions/pack_action.go
```diff
@@ -17,12 +18,16 @@ Optional properties:
 - compression -- compression type to use. Currently 'bzip2', 'gz', 'lzip', lzma', 'lzop',
 'xz' and 'zstd' compression types are supported. Use 'none' for uncompressed tarball.
 Use 'auto' to pick via file extension. The 'gz' compression type will be used by default.
+
+- subdir -- a subdirectory relative to / to compress, instead of the full rootfs
```
obbardc:
```suggestion
- subdir -- a subdirectory relative to '/' to compress, instead of the full rootfs.
```

actions/pack_action.go
```diff
@@ -90,7 +96,21 @@ func (pf *PackAction) Run(context *debos.Context) error {
...
+	var sourceDir = context.Rootdir
+	if len(pf.Subdir) > 0 {
+		var err error
+		sourceDir, err = debos.RestrictedPath(context.Rootdir, pf.Subdir)
+		if err != nil {
+			return err
+		}
+
+		if _, err := os.Stat(sourceDir); errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("subdirectory %s does not exist in rootfs", pf.Subdir)
```
obbardc:
```suggestion
			return fmt.Errorf("subdir '%s' does not exist in rootfs", pf.Subdir)
```

actions/pack_action.go
obbardc: also wondering if it makes sense to call this `source` - since the opposing name in unpack is `destination` ?

actions/unpack_action.go
```diff
@@ -94,5 +101,20 @@ func (pf *UnpackAction) Run(context *debos.Context) error {
...
+	if len(pf.Destination) > 0 {
+		destDir, err = debos.RestrictedPath(context.Rootdir, pf.Destination)
+		if err != nil {
+			return err
+		}
+
+		if _, err := os.Stat(destDir); errors.Is(err, os.ErrNotExist) {
+			err := os.MkdirAll(destDir, 0755)
+			if err != nil {
+				return fmt.Errorf("could not create %s to unpack into", destDir)
```
obbardc:
```suggestion
				return fmt.Errorf("could not create '%s' to unpack into", destDir)
```

actions/unpack_action.go
```diff
@@ -25,6 +26,9 @@ Optional properties:
 - origin -- reference to a named file or directory.
 The default value is 'artifacts' directory in case if this property is omitted.
 
+- destination -- path below / to unpack to. If this property is omitted, / is used.
```
obbardc: this and pack should have similar prose. i think this is a better description than pack somehow.

obbardc (changes requested):
minor comments

actions/pack_action.go
tobsan: imo, `source` is too vague here. To me it would make more sense to change `destination` to `targetdir` in the unpack action. EDIT: Or maybe `destdir`?

actions/unpack_action.go
tobsan: Updated pack to match the style used in unpack.

actions/pack_action.go
```diff
@@ -7,6 +7,7 @@ Create tarball with filesystem.
 	- action: pack
```
obbardc, line 7: nitpick: commit name could be `actions: pack: allow packing a subdirectory` and adding some prose into the commit message

actions/pack_action.go
```diff
@@ -7,6 +7,7 @@ Create tarball with filesystem.
...
 	  file: filename.ext
 	  compression: gz
+	  subdir: usr/
```
obbardc, line 10: should this have a `/` prefix ? also, does it work properly if it does have that prefix ?

with a `/` prefix set, does the resultant tar file still pack from the rootfs or possibly the fakemachine ? I doubt it, but good to check.

actions/pack_action.go
```diff
@@ -17,12 +18,17 @@ Optional properties:
 - compression -- compression type to use. Currently 'bzip2', 'gz', 'lzip', lzma', 'lzop',
 'xz' and 'zstd' compression types are supported. Use 'none' for uncompressed tarball.
 Use 'auto' to pick via file extension. The 'gz' compression type will be used by default.
+
+- subdir -- path below '/' to compress. If this property is omitted, '/' is used. The path
```
obbardc: how about:

`path inside the target rootfs to include in the tarball. The path must already exist inside the target rootfs. If this property is omitted, the whole target rootfs is included in the tarball.`

actions/pack_action.go
```diff
@@ -90,7 +97,21 @@ func (pf *PackAction) Run(context *debos.Context) error {
...
+	var sourceDir = context.Rootdir
+	if len(pf.Subdir) > 0 {
+		var err error
+		sourceDir, err = debos.RestrictedPath(context.Rootdir, pf.Subdir)
+		if err != nil {
+			return err
+		}
+
+		if _, err := os.Stat(sourceDir); errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("subdirectory '%s' does not exist in rootfs", pf.Subdir)
```
obbardc: `subdir` instead of `subdirectory` to match config

actions/pack_action.go
```diff
@@ -90,7 +97,21 @@ func (pf *PackAction) Run(context *debos.Context) error {
...
+		}
+	}
+
+	command = append(command, "-C", sourceDir)
```
obbardc, line 115: should this not go inside the `if len(pf.Subdir) > 0 {` stanza, so we only add `-C subdir` to tar when the property is set?

actions/pack_action.go
```diff
@@ -90,7 +97,21 @@ func (pf *PackAction) Run(context *debos.Context) error {
 	} else if tarOpts[pf.Compression] != "" {
 		command = append(command, tarOpts[pf.Compression])
 	}
-	command = append(command, "-C", context.Rootdir)
+
+	var sourceDir = context.Rootdir
+	if len(pf.Subdir) > 0 {
+		var err error
+		sourceDir, err = debos.RestrictedPath(context.Rootdir, pf.Subdir)
```
obbardc, line 105: nitpick: `subdir or subDir` instead of `sourceDir` to match config ?

.github/workflows/ci.yaml
```diff
@@ -216,6 +216,7 @@ jobs:
           - { name: "templating", case: "templating", variables: " -t escaped:\\$ba\\'d\\$gers\\ snakes" }
```
obbardc, line 216: nitpick: commit name could be `actions: unpack: allow unpacking into a subdirectory` and adding some prose into the commit message

actions/unpack_action.go
```diff
@@ -25,6 +26,9 @@ Optional properties:
 - origin -- reference to a named file or directory.
 The default value is 'artifacts' directory in case if this property is omitted.
 
+- destination -- path below '/' to unpack to. If this property is omitted, '/' is used.
```
obbardc: how about:

path inside the target rootfs to extract the tarball to. The path will be created inside the target rootfs if it doesn't already exist. If this property is omitted, the tarball is extracted to the root directory of the target rootfs.

actions/unpack_action.go
```diff
@@ -94,5 +100,18 @@ func (pf *UnpackAction) Run(context *debos.Context) error {
...
+	var destDir = context.Rootdir
+	if len(pf.Destination) > 0 {
+		destDir, err = debos.RestrictedPath(context.Rootdir, pf.Destination)
+		if err != nil {
+			return err
+		}
+
+		err := os.MkdirAll(destDir, 0755)
+		if err != nil {
+			return fmt.Errorf("could not create '%s' to unpack into", destDir)
```
obbardc: `could not create destination '%s' inside target rootfs`

tests/pack-subdir/test.yaml
```diff
@@ -11,9 +11,23 @@ actions:
     subdir: debos-docker/
 
   - action: run
-    command: tar tf ${ARTIFACTDIR}/subdir.gz > ${RECIPEDIR}/actual.txt
+    chroot: false
```
obbardc: nit: should this now be named `tests/pack-unpack-subdir/test.yaml` ?

tests/pack-unpack-subdir/expected-unpack.txt
```diff
@@ -1,3 +1,2 @@
-/scratch/root/local/test:
-README.md
 hello.txt
```
obbardc, line 1: nit: does this commit need folding into the other two commits ?

also, I'd be happy if you kept the tests together in a separate commit from the logic, but whatever you like.

obbardc (changes requested):
small nitpicks, but nice !

actions/pack_action.go
tobsan: No, `sourceDir` defaults to `context.Rootdir`, and we need to move into the rootdir before packing, or we'll get a tar with a` <topdir>/{usr, bin, home}` structure, instead of having `{usr, bin, home}` at the root.

actions/pack_action.go
tobsan: Given that we want to default to `context.Rootdir` when subdir property isn't set, I disagree.

actions/unpack_action.go
tobsan: Yep, that is way better, thanks!

tests/pack-unpack-subdir/expected-unpack.txt
tobsan: Indeed I messed that up. I've split tests out to a separate commit.

obbardc (approved):

actions/unpack_action.go
```diff
@@ -10,6 +10,7 @@ Only (compressed) tar archives are supported currently.
 	- action: unpack
 	  origin: name
 	  file: file.ext
+	  destination: usr/
```
tobsan: Should we call this `destdir` instead, to keep it similar between pack/unpack?

obbardc: `destdir` or `destination` is good with me. Will again defer to @sjoerdsimons

obbardc:
lgtm; will request @sjoerdsimons review before merging

PR #656, opened 2026-01-25
recipe: yaml: Disallow unknown action properties

### LINKED ISSUE #295: Debos silently accepts invalid action attributes
Consider the following example:

```
architecture: amd64

actions:
  - action: run
    description: Name
    chrot: false
    command: echo test

```

In the above example we've made a typo in the optional argument `chroot` -> `chrot`. One would expect that debos will parse and warn (fail even) on the chrot line, yet that's not the case. Mind you, this can be kept as opt-in toggle (when running debos), although debos should really verify and catch trivial human errors. We all make typos, some of use are even dyslexic.

obbardc:
Currently the YAML parser allows any property to be set in an action which can cause confusion as Debos will happily execute a recipe with any property set in an action (e.g. generative AI will happily create Debos recipes with wild, hallucinated properties).

Enable strict YAML decoding so recipes fail early when unknown or misspelled properties are present in actions, preventing silent misconfiguration of actions.

Also apply the same validation to the recipe action so that included recipes are also checked for missing action properties.

The YAML parser even logs where a user went wrong:

    [7:5] unknown field "unknown-property"
       5 | actions:
       6 |   - action: run
    >  7 |     unknown-property: value
           ^

Closes: #295

### DIFF
--- a/actions/recipe.go
+++ b/actions/recipe.go
@@ -115,14 +115,18 @@ type Recipe struct {
 }
 
 func (y *YamlAction) UnmarshalYAML(unmarshal func(interface{}) error) error {
-	var aux debos.BaseAction
+	var raw map[string]interface{}
 
-	err := unmarshal(&aux)
-	if err != nil {
+	if err := unmarshal(&raw); err != nil {
 		return err
 	}
 
-	switch aux.Action {
+	action, ok := raw["action"].(string)
+	if !ok || action == "" {
+		return fmt.Errorf("missing or invalid 'action' field")
+	}
+
+	switch action {
 	case "debootstrap":
 		y.Action = NewDebootstrapAction()
 	case "mmdebstrap":
@@ -158,10 +162,10 @@ func (y *YamlAction) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	case "recipe":
 		y.Action = &RecipeAction{}
 	default:
-		return fmt.Errorf("unknown action: %v", aux.Action)
+		return fmt.Errorf("unknown action: %v", action)
 	}
 
-	err = unmarshal(y.Action)
+	err := unmarshal(y.Action)
 	if err != nil {
 		return err
 	}
@@ -306,7 +310,14 @@ func (r *Recipe) Parse(file string, printRecipe bool, dump bool, templateVars ..
 		log.Printf("%s", data)
 	}
 
-	if err := yaml.Unmarshal(data.Bytes(), r); err != nil {
+	// Disallow unknown fields (e.g. those which are not actually available in an action)
+	// but allow fields prefixed with "." which is useful for YAML anchors in the top-level.
+	if err := yaml.UnmarshalWithOptions(
+		data.Bytes(),
+		r,
+		yaml.DisallowUnknownField(),
+		yaml.AllowFieldPrefixes("."),
+	); err != nil {
 		return err
 	}
 
--- a/actions/recipe.go
+++ b/actions/recipe.go
@@ -127,14 +127,18 @@ type Recipe struct {
 }
 
 func (y *YamlAction) UnmarshalYAML(unmarshal func(interface{}) error) error {
-	var aux debos.BaseAction
+	var raw map[string]interface{}
 
-	err := unmarshal(&aux)
-	if err != nil {
+	if err := unmarshal(&raw); err != nil {
 		return err
 	}
 
-	switch aux.Action {
+	action, ok := raw["action"].(string)
+	if !ok || action == "" {
+		return fmt.Errorf("missing or invalid 'action' field")
+	}
+
+	switch action {
 	case "debootstrap":
 		y.Action = NewDebootstrapAction()
 	case "mmdebstrap":
@@ -170,10 +174,10 @@ func (y *YamlAction) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	case "recipe":
 		y.Action = &RecipeAction{}
 	default:
-		return fmt.Errorf("unknown action: %v", aux.Action)
+		return fmt.Errorf("unknown action: %v", action)
 	}
 
-	err = unmarshal(y.Action)
+	err := unmarshal(y.Action)
 	if err != nil {
 		return err
 	}
@@ -318,7 +322,14 @@ func (r *Recipe) Parse(file string, printRecipe bool, dump bool, templateVars ..
 		log.Printf("%s", data)
 	}
 
-	if err := yaml.Unmarshal(data.Bytes(), r); err != nil {
+	// Disallow unknown fields (e.g. those which are not actually available in an action)
+	// but allow fields prefixed with "." which is useful for YAML anchors in the top-level.
+	if err := yaml.UnmarshalWithOptions(
+		data.Bytes(),
+		r,
+		yaml.DisallowUnknownField(),
+		yaml.AllowFieldPrefixes("."),
+	); err != nil {
 		return err
 	}
 
--- a/actions/recipe_action.go
+++ b/actions/recipe_action.go
@@ -35,6 +35,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strings"
 )
 
 type RecipeAction struct {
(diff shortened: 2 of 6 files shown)


obbardc:
oh, this causes `tests/templating/test.yaml` to fail:

```
$ go run cmd/debos/debos.go tests/templating/test.yaml 
2026/01/25 01:14:32 Recipe parsing failed:
2026/01/25 01:14:32 [7:1] unknown field ".common-run-settings"
2026/01/25 01:14:32    4 | architecture: amd64
2026/01/25 01:14:32    5 | 
2026/01/25 01:14:32    6 | # Test YAML anchors - define reusable command settings
2026/01/25 01:14:32 >  7 | .common-run-settings: &common-run-settings
2026/01/25 01:14:32        ^
2026/01/25 01:14:32    8 |   chroot: false
2026/01/25 01:14:32    9 | 
2026/01/25 01:14:32   10 | actions:
exit status 1
```

This is due to the additional top-level bits in that recipe not in `Recipe` struct (specifically the yaml anchor `.common-run-settings`).

Not sure what the right approach here is...

obbardc:
> oh, this causes `tests/templating/test.yaml` to fail:
> 
> ```
> $ go run cmd/debos/debos.go tests/templating/test.yaml 
> 2026/01/25 01:14:32 Recipe parsing failed:
> 2026/01/25 01:14:32 [7:1] unknown field ".common-run-settings"
> 2026/01/25 01:14:32    4 | architecture: amd64
> 2026/01/25 01:14:32    5 | 
> 2026/01/25 01:14:32    6 | # Test YAML anchors - define reusable command settings
> 2026/01/25 01:14:32 >  7 | .common-run-settings: &common-run-settings
> 2026/01/25 01:14:32        ^
> 2026/01/25 01:14:32    8 |   chroot: false
> 2026/01/25 01:14:32    9 | 
> 2026/01/25 01:14:32   10 | actions:
> exit status 1
> ```
> 
> This is due to the additional top-level bits in that recipe not in `Recipe` struct (specifically the yaml anchor `.common-run-settings`).
> 
> Not sure what the right approach here is...

OK, I limited the validation to just parsing of actions. Good trade-off, I think.

obbardc:
> > Not sure what the right approach here is...
> 
> OK, I limited the validation to just parsing of actions. Good trade-off, I think.

Spoke too soon, latest change causes the parsing error message to have a prefix of `action "run"` & wrong line numbers:

```
2026/01/25 02:00:56 action "run": [2:1] unknown field "unknown-property"
2026/01/25 02:00:56    1 | action: run
2026/01/25 02:00:56 >  2 | unknown-property: value
2026/01/25 02:00:56        ^
```

sjoerdsimons:
> > oh, this causes `tests/templating/test.yaml` to fail:
> > ```
> > $ go run cmd/debos/debos.go tests/templating/test.yaml 
> > 2026/01/25 01:14:32 Recipe parsing failed:
> > 2026/01/25 01:14:32 [7:1] unknown field ".common-run-settings"
> > 2026/01/25 01:14:32    4 | architecture: amd64
> > 2026/01/25 01:14:32    5 | 
> > 2026/01/25 01:14:32    6 | # Test YAML anchors - define reusable command settings
> > 2026/01/25 01:14:32 >  7 | .common-run-settings: &common-run-settings
> > 2026/01/25 01:14:32        ^
> > 2026/01/25 01:14:32    8 |   chroot: false
> > 2026/01/25 01:14:32    9 | 
> > 2026/01/25 01:14:32   10 | actions:
> > exit status 1
> > ```
  
> >   
> > This is due to the additional top-level bits in that recipe not in `Recipe` struct (specifically the yaml anchor `.common-run-settings`).
> > Not sure what the right approach here is...
> 
> OK, I limited the validation to just parsing of actions. Good trade-off, I think.

Ooi did you try with AllowFieldPrefixes https://github.com/goccy/go-yaml/blob/v1.19.2/option.go#L74 -- allowing "." as a prefix might do the trick -- That said i'm not sure how common/conventional the . prefix is for these kind of setups

obbardc:
> > > oh, this causes `tests/templating/test.yaml` to fail:
> > > ```
> > > $ go run cmd/debos/debos.go tests/templating/test.yaml 
> > > 2026/01/25 01:14:32 Recipe parsing failed:
> > > 2026/01/25 01:14:32 [7:1] unknown field ".common-run-settings"
> > > 2026/01/25 01:14:32    4 | architecture: amd64
> > > 2026/01/25 01:14:32    5 | 
> > > 2026/01/25 01:14:32    6 | # Test YAML anchors - define reusable command settings
> > > 2026/01/25 01:14:32 >  7 | .common-run-settings: &common-run-settings
> > > 2026/01/25 01:14:32        ^
> > > 2026/01/25 01:14:32    8 |   chroot: false
> > > 2026/01/25 01:14:32    9 | 
> > > 2026/01/25 01:14:32   10 | actions:
> > > exit status 1
> > > ```
> 
> > > This is due to the additional top-level bits in that recipe not in `Recipe` struct (specifically the yaml anchor `.common-run-settings`).
> > > Not sure what the right approach here is...
> > 
> > 
> > OK, I limited the validation to just parsing of actions. Good trade-off, I think.
> 
> Ooi did you try with AllowFieldPrefixes https://github.com/goccy/go-yaml/blob/v1.19.2/option.go#L74 -- allowing "." as a prefix might do the trick -- That said i'm not sure how common/conventional the . prefix is for these kind of setups

of course reading the documentation would have been too easy. Thanks - that works and seems to fix the anchors used in that test.

sjoerdsimons:
> > Ooi did you try with AllowFieldPrefixes https://github.com/goccy/go-yaml/blob/v1.19.2/option.go#L74 -- allowing "." as a prefix might do the trick -- That said i'm not sure how common/conventional the . prefix is for these kind of setups
> 
> of course reading the documentation would have been too easy. Thanks - that works and seems to fix the anchors used in that test.

lol; That said i'm not sure it's the *right* think to do; Googling aroudn quickly there seems to be some suggestion for conventions to use "hidden" blocks (aka those prefixed with `.`). Which is also something gitlab suggests (which will be the reason i've always used that convention). But i've not seen an authoritative source say that's the expected best practise :/.

This is one we need to be careful with as it can break existing recipes. So it needs to be very clearly documented and maybe even have it hidden behind a `--strict` option initially or somesuch? Or can the yaml parser spit out warnings rather then failures?

obbardc:
> > > Ooi did you try with AllowFieldPrefixes https://github.com/goccy/go-yaml/blob/v1.19.2/option.go#L74 -- allowing "." as a prefix might do the trick -- That said i'm not sure how common/conventional the . prefix is for these kind of setups
> > 
> > 
> > of course reading the documentation would have been too easy. Thanks - that works and seems to fix the anchors used in that test.
> 
> lol; That said i'm not sure it's the _right_ think to do; Googling aroudn quickly there seems to be some suggestion for conventions to use "hidden" blocks (aka those prefixed with `.`). Which is also something gitlab suggests (which will be the reason i've always used that convention). But i've not seen an authoritative source say that's the expected best practise :/.

Yeah, but really how many times do you use anchors in debos recipes ?

 
> This is one we need to be careful with as it can break existing recipes. So it needs to be very clearly documented and maybe even have it hidden behind a `--strict` option initially or somesuch? Or can the yaml parser spit out warnings rather then failures?

I doubt it would break any existing good recipes, unless someone uses incorrect parameters in their recipe (which is a cardinal sin anyway) or some strange anchoring?


Agree r.e. documentation, this was merely a nice-to-have idea rather than essential in my mind :-)

obbardc:
heh:

```
2026/02/05 20:27:53 Recipe '/recipes/ospack-debian/ospack-debian.yaml':
2026/02/05 20:27:53 Recipe parsing failed:
2026/02/05 20:27:53 [30:5] unknown field "descripion"
2026/02/05 20:27:53   27 |     script: scripts/setup-user.sh
2026/02/05 20:27:53   28 | 
2026/02/05 20:27:53   29 |   - action: overlay
2026/02/05 20:27:53 > 30 |     descripion: Copy files inside the rootfs
```

obbardc:
The failing jobs will be fixed by https://github.com/go-debos/debos/pull/673

PR #658, opened 2026-01-27
actions: overlay: create parent dest dirs as needed

### LINKED ISSUE #217: Create overlay destination if not exist
If overlay destination does not exist, the overlay action should create the destination directory structure before copying files.

tobsan:
Fixes #217

### DIFF
--- a/actions/overlay_action.go
+++ b/actions/overlay_action.go
@@ -19,8 +19,8 @@ Optional properties:
 - origin -- reference to named file or directory.
 
 - destination -- absolute path in the target rootfs where 'source' will be copied.
-All existing files will be overwritten.
-If destination isn't set '/' of the rootfs will be used.
+Any missing parent directories will be created. All existing files will be overwritten.
+If destination isn't set the root of the target rootfs will be used.
 */
 package actions
 
@@ -29,6 +29,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"path"
 
 	"github.com/go-debos/debos"
 )
@@ -71,12 +72,20 @@ func (overlay *OverlayAction) Run(context *debos.Context) error {
 		}
 	}
 
-	sourcedir := debos.CleanPathAt(overlay.Source, origin)
+	source := debos.CleanPathAt(overlay.Source, origin)
 	destination, err := debos.RestrictedPath(context.Rootdir, overlay.Destination)
 	if err != nil {
 		return err
 	}
 
-	log.Printf("Overlaying %s on %s", sourcedir, destination)
-	return debos.CopyTree(sourcedir, destination)
+	// Make sure all parts of the destination except the last exists.
+	destinationParent := path.Dir(destination)
+	err = os.MkdirAll(destinationParent, 0755)
+	if err != nil {
+		return fmt.Errorf("could not create parent destination path for overlay '%s': %w", destination, err)
+	}
+
+	// Copy source into dest
+	log.Printf("Overlaying %s on %s", source, destination)
+	return debos.CopyTree(source, destination)
 }
--- a/tests/exit_test/exit_test.sh
+++ b/tests/exit_test/exit_test.sh
@@ -51,6 +51,9 @@ else
     SUDO=
 fi
 
+# Overlay test needs a directory to overlay
+mkdir -p A/B/C/D
+
 expect_success debos --help
 expect_failure debos --not-a-valid-option
 expect_failure debos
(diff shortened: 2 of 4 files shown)


actions/overlay_action.go
```diff
@@ -19,7 +19,7 @@ Optional properties:
 - origin -- reference to named file or directory.
 
 - destination -- absolute path in the target rootfs where 'source' will be copied.
-All existing files will be overwritten.
+Any missing parent directories will be created. All existing files will be overwritten.
 If destination isn't set '/' of the rootfs will be used.
```
obbardc:
```suggestion
If destination isn't set, the root of the target rootfs will be used.
```

I know this isn't modified by you, but I think this is better gramatically.

obbardc (commented):
Looks good, thanks !

actions/overlay_action.go
tobsan: Agree, updated accordingly.

actions/overlay_action.go
```diff
@@ -19,8 +20,11 @@ Optional properties:
 - origin -- reference to named file or directory.
 
 - destination -- absolute path in the target rootfs where 'source' will be copied.
-All existing files will be overwritten.
-If destination isn't set '/' of the rootfs will be used.
+Any missing parent directories will be created if the 'created' property is set, otherwise
+missing parent directories is an error. All existing files will be overwritten.
+If destination isn't set the root of the target rootfs will be used.
+
+- create -- create leading parent directories for 'destination' as needed when set.
```
obbardc: Not really a fan of this property and think we should create destination by default if it doesn't exist; will defer to @sjoerdsimons

obbardc (changes requested):

actions/overlay_action.go
tobsan: Well, @andhe pointed out that it might make sense to be backwards compatible here and not create anything by default, which makes sense to me. Are there good arguments to *always* create the destination?

obbardc:
sidenote: does this need the patch from https://github.com/go-debos/debos/pull/660 cherry-picking ?

tobsan:
> sidenote: does this need the patch from #660 cherry-picking ?

#660 is about the source rather than the destination, but why not merge that first and I'll just rebase.

obbardc:
> > sidenote: does this need the patch from #660 cherry-picking ?
> 
> #660 is about the source rather than the destination, but why not merge that first and I'll just rebase.

thanks, just wondering if it was useful here.

actions/overlay_action.go
sjoerdsimons: The backwards compatible argument doesn't really fly here. Creating by default doesn't break currently working recipes  nor do currently working recipes see a behaviour change. The behaviour change would only be for currently failing recipes.

Note we already see directory behaviour if there is only one level of leading directory (snakes is a directory containing badgers.txt):
```
$ cat test.yaml 
architecture: arm64
actions:
   - action: overlay
     destination: /badgers
     source: snakes
   - action: run
     chroot: false
     command: find ${ROOTDIR}

$ debos --debug-shell test.yaml
2026/02/03 21:04:17 ==== overlay ====
2026/02/03 21:04:17 Overlaying /home/sjoerd/tmp/test/snakes on /scratch/root/badgers
2026/02/03 21:04:17 ==== run ====
2026/02/03 21:04:17 find ${ROOTDIR} | /scratch/root
2026/02/03 21:04:17 find ${ROOTDIR} | /scratch/root/badgers
2026/02/03 21:04:17 find ${ROOTDIR} | /scratch/root/badgers/badgers.txt
2026/02/03 22:04:18 ==== Recipe done ====
```

This explains why i was somewhat surprised by this PR; But it turns out everywhere i was already relying on the leading directory creation behavior, it was always just one level, hence masking the issue.  That also means it might actually be your new create option that causes backwards compat issues (though i didn't verify)

From a bigger picture though; I agree with chris on the default behaviour. Yes people can mistake, but that's no different from having typos in the directories inside an overlay. I could potentially be convinced to have it the other way around (by default create, allow configuration to deny)

sjoerdsimons (dismissed):

actions/overlay_action.go
tobsan: These are all valid points, and I'm not keen on accidentally creating a compatibility issue, so I'll just update to always create. If somebody needs the ability to deny in the future, we could add it then.

obbardc: sounds great to me, I can review & merge once done now we have direction.

obbardc (approved):

tobsan:
Just rebased to integrate on top of #660

obbardc (approved):

PR #660, opened 2026-02-03
actions: overlay: Allow absolute paths in overlay

vigneshraman:
path.Join() always prepends origin to source, breaking absolute paths. Use CleanPathAt() which preserves absolute paths and joins relative ones.

### DIFF
--- a/tests/exit_test/exit_test.sh
+++ b/tests/exit_test/exit_test.sh
@@ -67,6 +67,7 @@ expect_success debos good.yaml
 expect_failure debos bad.yaml
 expect_failure debos pre-machine-failure.yaml
 expect_failure debos post-machine-failure.yaml
+expect_success debos overlay-absolute-path.yaml
 expect_failure debos overlay-missing-destination.yaml
 expect_failure debos overlay-missing-source.yaml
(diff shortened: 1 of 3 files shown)


vigneshraman:
> Would be good to understand what usecase this fixes as its not clear from the commit message, or even better have a unit test for the case where it breaks.

Sure, will add a test case and update the commit message.

tests/overlay-absolute-path/test.yaml
```diff
@@ -0,0 +1,27 @@
...
+    apt-opts:
+      - Apt::Install-Recommends "true"
+{{- end }}
+
+  - action: apt
+    description: Install some base packages
+    packages:
+      - procps
+
+  - action: overlay
```
obbardc: would be better to add this into the exit_codes test, i guess, since that doesn't create a fakemachine for each unit test

vigneshraman: Sure, will test and update

vigneshraman: @obbardc please could you review the updated changes. Thanks.

obbardc (approved):

PR #674, opened 2026-02-06
actions: overlay: allow using wildcards in source

### LINKED ISSUE #638: actions: overlay: allow wildcards for source path
overlay action source should accept shell bits / wildcards (in destination would be great)

This was from my old Debos development notes; not sure why I say `in destination would be great` as I don't think that it would be useful now I've had some time to think about it?

tobsan:
Fixes #638

### DIFF
--- a/actions/overlay_action.go
+++ b/actions/overlay_action.go
@@ -13,6 +13,7 @@ Mandatory properties:
 
 - source -- relative path to the directory or file located in path referenced by `origin`.
 In case if this property is absent then pure path referenced by 'origin' will be used.
+The source property may contain wildcards.
 
 Optional properties:
 
@@ -28,7 +29,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
-	"os"
+	"path/filepath"
 
 	"github.com/go-debos/debos"
 )
@@ -51,9 +52,14 @@ func (overlay *OverlayAction) Verify(context *debos.Context) error {
 
 	/* if origin is the recipe, check the path exists on disk */
 	if len(overlay.Origin) == 0 || overlay.Origin == "recipe" {
-		sourceDir := debos.CleanPathAt(overlay.Source, context.RecipeDir)
-		if _, err := os.Stat(sourceDir); err != nil {
-			return err
+		pattern := debos.CleanPathAt(overlay.Source, context.RecipeDir)
+		matches, err := filepath.Glob(pattern)
+		if err != nil {
+			return fmt.Errorf("failed to glob for '%s': %w", pattern, err)
+		}
+
+		if len(matches) == 0 {
+			return fmt.Errorf("no matches for '%s'", pattern)
 		}
 	}
 
@@ -71,12 +77,20 @@ func (overlay *OverlayAction) Run(context *debos.Context) error {
 		}
 	}
 
-	sourcedir := debos.CleanPathAt(overlay.Source, origin)
 	destination, err := debos.RestrictedPath(context.Rootdir, overlay.Destination)
 	if err != nil {
 		return err
 	}
 
-	log.Printf("Overlaying %s on %s", sourcedir, destination)
-	return debos.CopyTree(sourcedir, destination)
+	pattern := debos.CleanPathAt(overlay.Source, origin)
+	matches, _ := filepath.Glob(pattern)
+	for _, source := range matches {
+		log.Printf("Overlaying %s on %s", source, destination)
+		err := debos.CopyTree(source, destination)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
--- a/actions/overlay_action.go
+++ b/actions/overlay_action.go
@@ -13,6 +13,7 @@ Mandatory properties:
 
 - source -- relative path to the directory or file located in path referenced by `origin`.
 In case if this property is absent then pure path referenced by 'origin' will be used.
+The source property may contain wildcards.
 
 Optional properties:
 
@@ -30,6 +31,7 @@ import (
 	"log"
 	"os"
 	"path"
+	"path/filepath"
 
 	"github.com/go-debos/debos"
 )
@@ -52,9 +54,14 @@ func (overlay *OverlayAction) Verify(context *debos.Context) error {
 
 	/* if origin is the recipe, check the path exists on disk */
 	if len(overlay.Origin) == 0 || overlay.Origin == "recipe" {
-		sourceDir := debos.CleanPathAt(overlay.Source, context.RecipeDir)
-		if _, err := os.Stat(sourceDir); err != nil {
-			return err
+		pattern := debos.CleanPathAt(overlay.Source, context.RecipeDir)
+		matches, err := filepath.Glob(pattern)
+		if err != nil {
+			return fmt.Errorf("failed to glob for '%s': %w", pattern, err)
+		}
+
+		if len(matches) == 0 {
+			return fmt.Errorf("no matches for '%s'", pattern)
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


actions/overlay_action.go
```diff
@@ -52,9 +53,14 @@ func (overlay *OverlayAction) Verify(context *debos.Context) error {
...
-		if _, err := os.Stat(sourceDir); err != nil {
-			return err
+		pattern := path.Join(context.RecipeDir, overlay.Source)
+		matches, err := filepath.Glob(pattern)
+		if err != nil {
+			return fmt.Errorf("failed to glob for '%s': %w", pattern, err)
+		}
+
+		if len(matches) == 0 {
+			return fmt.Errorf("no matches for '%s'", pattern)
```
tobsan, line 64: Is there a design decision ever made on what to check for in Verify? One could want to generate the source on the fly in the recipe, which would not work with this approach. I think most other actions check file existence etc in Verify, so I stuck with it.

obbardc: What about if the file comes from an origin, e.g. download action before the overlay?

about design decisions, that would be good in CONTRIBUTING. I did have an open issue about that, be good to store the idea in there.

obbardc: ah this is overlaying from the `recipe` origin, which should probably always exist at build-time.

tests/recipes/subdirs/overlay-b/overlay-B.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+   - action: overlay
+     source: B*
+     destination: /usr/local
+
+   # Check that both directories were properly overlayed
+
+   - action: run
+     chroot: false
+     command: find ${ROOTDIR}/usr/local/BA
```
obbardc: worth checking content list actually matches what is in the recipedir to properly check the overlay worked ?

tobsan: Alright, I added a check for that.

obbardc:
Can you split the test into separate commit?

tests/recipes/test.yaml
```diff
@@ -3,7 +3,7 @@ architecture: amd64
 
 actions:
   - action: recipe
-    description: Simple  test
+    description: Simple test
```
obbardc, line 6: nit: can you put this in a separate commit/PR which can go in immediately?

obbardc (dismissed):
lgtm; will request @sjoerdsimons review before merging

tests/recipes/test.yaml
tobsan: Sure thing: #679

obbardc:
@tobsan can you please resolve the conflicts ? ready to merge once done ;-)

tobsan:
> @tobsan can you please resolve the conflicts ? ready to merge once done ;-)

Done!

actions/overlay_action.go
```diff
@@ -52,9 +54,14 @@ func (overlay *OverlayAction) Verify(context *debos.Context) error {
 
 	/* if origin is the recipe, check the path exists on disk */
 	if len(overlay.Origin) == 0 || overlay.Origin == "recipe" {
-		sourceDir := debos.CleanPathAt(overlay.Source, context.RecipeDir)
-		if _, err := os.Stat(sourceDir); err != nil {
-			return err
+		pattern := debos.CleanPathAt(overlay.Source, context.RecipeDir)
```
obbardc, line 57: nit: call this `sourceDir`?

also, why do you remove the `os.Stat` call?

I wonder if it makes sense to just do the globbing && these checks in `Run` rather than globbing twice (once in Verify and again in Run?) OR possibly cache matches somewhere ?

Does it also make sense to call `os.Stat` on each matched file ? not sure.

actions/overlay_action.go
```diff
@@ -85,7 +91,15 @@ func (overlay *OverlayAction) Run(context *debos.Context) error {
...
 	}
 
-	// Copy source into dest
-	log.Printf("Overlaying %s on %s", source, destination)
-	return debos.CopyTree(source, destination)
+	pattern := debos.CleanPathAt(overlay.Source, origin)
+	matches, _ := filepath.Glob(pattern)
+	for _, source := range matches {
+		log.Printf("Overlaying %s on %s", source, destination)
+		err := debos.CopyTree(source, destination)
```
obbardc, line 98: nitpick:

```
if err := debos.CopyTree(source, destination); err != nil {
    return err
}
```

Also, does this work when copying single files matched by the glob? Maybe my question makes no sense and the globbing doesn't work this way :-).

obbardc (changes requested):
small comments

actions/overlay_action.go
tobsan: `sourcePattern` is perhaps a better name? My reasoning for not using `os.Stat` was that `filepath.Glob` gives everything that matches. But now that I read the docs for Glob in more detail I see that we probably should stat each file anyway :).

And yes, if we iterate over all the matches in `Verify`, we could just cache that result. Since we don't depend on something happening inside the sysroot for this, it makes more sense to do the checks in `Verify` rather than `Run` imo.

actions/overlay_action.go
tobsan: iirc, it does, but I will make sure and add a test for it.

actions/overlay_action.go
tobsan: `origin` made the caching impossible, since that's populated during runtime.

obbardc: > `origin` made the caching impossible, since that's populated during runtime.

yep, think moving this into Run is best because `origin` being populated in Run.

obbardc: also - not sure why you want to append "Pattern" to the variable name - I still think leaving it as `sourceDir` is better (unless you can convince me otherwise) :-)

actions/overlay_action.go
```diff
@@ -13,6 +13,7 @@ Mandatory properties:
 
 - source -- relative path to the directory or file located in path referenced by `origin`.
 In case if this property is absent then pure path referenced by 'origin' will be used.
+The source property may contain wildcards.
```
obbardc, line 16:
``` 
- source -- relative path to the directory or file located in the path referenced by `origin`.
  If this property is absent, the path referenced by `origin` is used directly.
  `source` may contain wildcards, which allow one pattern to match multiple files or directories.
```

The second line SHOULD be a separate commit to improve the doc's grammar.

actions/overlay_action.go
tobsan: Hah, it did NOT, and it revealed a bigger issue with globbing semantics; depending on how we want things to work, the following can be interpreted in two different ways depending on the number of matches:
```

- action: overlay
  source: a*.txt
  destination: /usr/local/something
```

Should a single match be replacing the `something` file/dir? That's how overlays work currently. But if there are multiple matches, it would make more sense to put them *inside* `something`, although that would make the semantics context-dependent. That can be fine, but we could also say that if you glob, the results will *always* end up *inside* the `destination`. That makes globbing semantics clear, although they would differ from the non-globbing cases, which would be identical to matching a single file/dir.

actions/overlay_action.go
tobsan: No, you are right. <strike>But `sourcePath` then, since one may very well want to overlay files and not only directories.</strike>

actions/overlay_action.go
obbardc: Could make sense to add a separate commit (or even PR) & test to check:

- when source is a file, check if `/usr/local/something` is already on the rootfs. If it's a directory then copy it into the dir, otherwise copy it as a file to the path?
- if there's a slash at the end of destination, copy into the directory (or create it)

I am probably missing some edge cases. It's a bit convoluted...

actions/overlay_action.go
obbardc: Ah - possibly just `source` would be better then. I know you reuse that variable later.

Issue #578, opened 2025-08-11
Run action with script property should truncate the label


obbardc:
Just a reminder from https://github.com/go-debos/debos/pull/302/commits/3a3a1f655153068660902f86f7a902316fdd34b9#r1451551605

PR #675, opened 2026-02-07
actions: run: fix label for script args with / in

### LINKED ISSUE #346: Wrong label for run action, when argument contains a slash
```
$ cat << EOF > recipe.yaml 
architecture: amd64

actions:
  - action: run
    postprocess: true
    script: scripts/process-template foobar.tpl

  - action: run
    postprocess: true
    script: scripts/process-template some-dir/foobar.tpl 
EOF

$ mkdir scripts
$ echo -e '#!/bin/sh\necho $0 $@' > scripts/process-template
$ chmod +x scripts/process-template

$ debos recipe.yaml 
Running /debos --artifactdir /home/me/recipes /home/me/recipes/recipe.yaml using kvm backend
Powering off.
2022/05/30 13:02:56 ==== run ====
2022/05/30 13:02:56 process-template foobar.tpl | /home/me/recipes/scripts/process-template template.txt
2022/05/30 13:02:56 ==== run ====
2022/05/30 13:02:56 foobar.tpl | /home/me/recipes/scripts/process-template scripts/template.txt
2022/05/30 13:02:56 ==== Recipe done ====
```

As we can see above:
1. First action, the label for the action is `process-template foobar.tpl`, aka. the basename of the script, rather than the complete path, is used for display.
2. Second action, the label is simply `foobar.tpl`, this is not correct.

I've looked at the code and it seems that this line is wrong:

https://github.com/go-debos/debos/blob/baf95f25772bad2d96df8da7356009b47b824081/actions/run_action.go#L117

`path.Base` should be applied on the name of the script, rather than on the whole command line.

tobsan:
Fixes #346

### DIFF
--- a/actions/run_action.go
+++ b/actions/run_action.go
@@ -112,7 +112,10 @@ func (run *RunAction) doRun(context debos.Context) error {
 			script[0] = strings.Replace(script[0], scriptpath, "/tmp/script", 1)
 		}
 		cmdline = []string{strings.Join(script, " ")}
-		label = path.Base(run.Script)
+		label = path.Base(script[0])
+		if len(script) > 1 {
+			label += " " + strings.Join(script[1:], " ")
+		}
 	} else {
 		cmdline = []string{run.Command}
 


tobsan:
Should I update this to fix #578 while I'm at it?

obbardc:
> Should I update this to fix #578 while I'm at it?

Can do, or a separate pr. I've already set this to automerge :p

PR #676, opened 2026-02-07
actions: run: truncate label for scripts too

### LINKED ISSUE #578: Run action with script property should truncate the label

tobsan:
Closes #578

### DIFF
--- a/actions/run_action.go
+++ b/actions/run_action.go
@@ -116,6 +116,12 @@ func (run *RunAction) doRun(context debos.Context) error {
 		if len(script) > 1 {
 			label += " " + strings.Join(script[1:], " ")
 		}
+
+		if len(label) > maxLabelLength {
+			label = label[:maxLabelLength]
+			label = strings.TrimSpace(label)
+			label += "..."
+		}
 	} else {
 		cmdline = []string{run.Command}
 
@@ -128,9 +134,7 @@ func (run *RunAction) doRun(context debos.Context) error {
 		// Make it clear a long or a multi-line command is being run
 		if len(label) > maxLabelLength {
 			label = label[:maxLabelLength]
-
 			label = strings.TrimSpace(label)
-
 			label += "..."
 		} else if len(commands) > 1 {
 			label += "..."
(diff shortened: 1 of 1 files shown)


actions/run_action.go
```diff
@@ -113,6 +113,12 @@ func (run *RunAction) doRun(context debos.Context) error {
 		}
 		cmdline = []string{strings.Join(script, " ")}
 		label = path.Base(run.Script)
```
obbardc: it'd be quite nice to move the truncate bits outside of the if statements, so we only truncate once.

What do you think ? Happy to do that in a cleanup PR if you'd rather,

obbardc (changes requested):

actions/run_action.go
tobsan: You read my mind, I also found the label handling code to be a bit repetetive. I'll do it in here.

actions/run_action.go
```diff
@@ -116,23 +126,19 @@ func (run *RunAction) doRun(context debos.Context) error {
...
-
-		// Make it clear a long or a multi-line command is being run
-		if len(label) > maxLabelLength {
-			label = label[:maxLabelLength]
 
-			label = strings.TrimSpace(label)
-
-			label += "..."
-		} else if len(commands) > 1 {
+		// Multiline commands get labeled based on the first command
```
obbardc, line 151: would it not just make sense to have the truncateLabel code inline after the if statement and only called once ? e.g. inside the if statement generate the long label, then truncate it after ?
maybe i am missing something though and this is just a nitpick :-)

obbardc (changes requested):

actions/run_action.go
tobsan: Updated but still with its own function. Could be inlined instead if we want to, but felt a bit long for that.

obbardc (approved):

PR #677, opened 2026-02-08
recipe: Ensure postprocess run actions are at the end of the recipe

### LINKED ISSUE #493: postprocess commands should be expected to be at the end of the recipe
Error out in `Verify` if `postprocess: true` run actions are not at the end of a recipe (or sub-recipe!). This will help people shooting themselves in the foot.

obbardc:
Validate that all run actions with `postprocess: true` appear at the
end of the recipe. Once the first postprocess run action is encountered
all remaining actions must also be postprocess run actions otherwise
Parse() returns an error.

Note: postprocess run actions in child recipes execute at the end of the
child recipe, not at the end of the overall recipe.

Closes: https://github.com/go-debos/debos/issues/493

### DIFF
--- a/actions/recipe.go
+++ b/actions/recipe.go
@@ -322,6 +322,26 @@ func (r *Recipe) Parse(file string, printRecipe bool, dump bool, templateVars ..
 		return fmt.Errorf("Recipe file must have at least one action")
 	}
 
+	/* ensure all run actions with postprocess set are at the end of the recipe.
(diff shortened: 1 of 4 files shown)


obbardc:
Need to consider:

1. what we do with child recipes (e.g move any run postprocess actions to end of execution; that could be a follow-up PR)
2. if we can have other non-run actions running after run actions with postprocess: true (I doubt it, but need to point it out in case I have missed something silly)

PR #682, opened 2026-02-14
actions: add image-extract action

### LINKED ISSUE #495: Create dissect-image action to extract separate filesystem images
We need a solution to extract sparse images from a created disk image.

- input is a list of partition name and filename
- output is a number of sparse images in `$ARTIFACTDIR`
- action could be defined anywhere in the recipe, but run after the fakemachine has finished (e.g. after the image has been unmounted) and before the `postprocess: true` commands.

tobsan:
Doing the opposite of the raw action, in the simplest possible way.

Fixes #495

### DIFF
--- a/actions/dissect_image_action.go
+++ b/actions/dissect_image_action.go
@@ -1,0 +1,69 @@
+/*
+DissectImage Action
+
+Dissect allows to pick out parts of the created image from the ImagePartition
+and FilesystemDeploy actions, and save those parts as raw image files. In a way,
+this is the Raw action in reverse.
+
+	# Yaml syntax:
+	- action: image-dissect
+	  parts:
+	  	<list of parts>
+
+Mandatory properties:
+- parts -- list of parts of the image to dissect. These are processed one by one
+and may overlap.
+
+Optional properties:
+
+	# Yaml syntax for parts:
+	  parts:
+	    - partname: partition name
+		  start: offset
+		  end: offset
+		  destination: filename
+
+Mandatory properties:
+- destination -- where to store the extracted part
+
+Optional properties:
+- partname -- to extract a full partition, matching on partname is the easiest
+
+- start -- offset in bytes or in sector numbers, e.g. 256s. This becomes the starting
+point from where to extract when partname is not given.
+
+- end -- offset in bytes or in sector numbers, e.g. 512s. This becomes the end point
+at which the extraction stops, when partname is not given.
+
+The sector size is either the recipe header 'sectorsize' or the default 512 sector
+size.
+*/
+package actions
+
+import (
+	"github.com/go-debos/debos"
+)
+
+type ImagePart struct {
+	PartName	string
+	StartOffset string
+	EndOffset	string
+	// TBD
+}
+
+type ImageDissectAction struct {
+	debos.BaseAction    `yaml:",inline"`
+	Parts			 []ImagePart
+	// TBD
+}
+
+func (dis *ImageDissectAction) Verify(ctx *debos.Context) error {
+	// TODO
+	return nil
+}
+
+func (dis *ImageDissectAction) PostMachine(ctx *debos.Context) error {
+	// Has to run after the images have been unmounted.
+	// TODO
+	return nil
+}
--- a/actions/image_extract_action.go
+++ b/actions/image_extract_action.go
@@ -1,0 +1,118 @@
+/*
+ImageExtract Action
+
+Extract allows to pick out parts of the created image from the ImagePartition
+and FilesystemDeploy actions, and save those parts as raw image files. In a way,
+this is the Raw action in reverse.
+
+	# Yaml syntax:
+	- action: image-extract
+	  parts:
+	  	<list of parts>
+
+Mandatory properties:
+- parts -- list of parts of the image to extract.
+
+Optional properties:
+
+	# Yaml syntax for parts:
+	  parts:
+	    - partname: partition name
+		  destination: filename
+
+Mandatory properties:
+- destination -- where to store the extracted part
+
+- partname -- to extract a full partition, matching on partname is the easiest
+*/
+package actions
+
+import (
+	"errors"
+	"fmt"
+	"github.com/go-debos/debos"
+	"io"
+	"os"
+	"path"
+)
+
+type ImagePart struct {
+	PartName    string
+	Destination string
+}
+
+type ImageExtractAction struct {
+	debos.BaseAction `yaml:",inline"`
+	Parts            []ImagePart
+}
+
+func (ext *ImageExtractAction) Verify(_ *debos.Context) error {
+	for _, part := range ext.Parts {
+		if len(part.PartName) == 0 {
+			return fmt.Errorf("partname is a mandatory property")
+		}
+
+		if len(part.Destination) == 0 {
+			return fmt.Errorf("destination is a mandatory property")
+		}
+	}
+
+	// Can't verify partname existence at this point as ImagePartition metadata
+	// isn't added to the Context until its Run part
+	return nil
+}
+
+func (ext *ImageExtractAction) Run(context *debos.Context) error {
+	if context.ImageFSTab.Len() == 0 {
+		return errors.New("fstab not generated, missing image-partition action?")
+	}
+
+	origin := context.RecipeDir
+	for _, part := range ext.Parts {
+		// TODO: This lookup logic is copied from raw action, put in a shared
+		// 		 helper util instead?
+		var devicePath string
+		for _, existingPart := range context.ImagePartitions {
+			if existingPart.Name == part.PartName {
+				devicePath = existingPart.DevicePath
+				break
+			}
+		}
+
+		if devicePath == "" {
+			return fmt.Errorf("failed to find partition named %s", part.PartName)
+		}
+
+		// TODO: This copy logic is ALSO copied from raw action. Should do a
+		// 		 generic copyFromTo(source, offset, dest, offset)
+		source, err := os.Open(devicePath)
+		if err != nil {
+			return fmt.Errorf("failed to open %s: %w", devicePath, err)
+		}
+		defer source.Close()
+
+		fi, err := source.Stat()
+		if err != nil {
+			return fmt.Errorf("failed to stat %s: %w", devicePath, err)
+		}
+
+		destPath := path.Join(origin, part.Destination)
+		dest, err := os.OpenFile(destPath, os.O_WRONLY, 0)
+		if err != nil {
+			return fmt.Errorf("failed to open %s: %w", part.Destination, err)
+		}
+		defer dest.Close()
+
+		bytesCopied, err := io.Copy(dest, source)
+		if err != nil || bytesCopied < fi.Size() {
+			return fmt.Errorf("couldn't write complete data: %w", err)
+		}
+
+		err = dest.Sync()
+		if err != nil {
+			return fmt.Errorf("couldn't sync content: %w", err)
+		}
+	}
+
+	return nil
+}
--- a/actions/recipe.go
+++ b/actions/recipe.go
@@ -62,6 +62,8 @@ other value than '512' is not supported by the 'uml' fakemachine backend.
 
 - image-partition -- https://godoc.org/github.com/go-debos/debos/actions#hdr-ImagePartition_Action
 
+- image-dissect -- https://godoc.org/github.com/go-debos/debos/actions#hdr-ImageDissect_Action
+
 - install-deb -- https://godoc.org/github.com/go-debos/debos/actions#hdr-InstallDeb_Action
 
 - ostree-commit -- https://godoc.org/github.com/go-debos/debos/actions#hdr-OstreeCommit_Action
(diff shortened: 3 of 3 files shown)


actions/dissect_image_action.go
```diff
@@ -0,0 +1,69 @@
+/*
+DissectImage Action
+
+Dissect allows to pick out parts of the created image from the ImagePartition
+and FilesystemDeploy actions, and save those parts as raw image files. In a way,
+this is the Raw action in reverse.
+
+	# Yaml syntax:
+	- action: image-dissect
```
tobsan: Should we call it image-extract instead?

actions/dissect_image_action.go
```diff
@@ -0,0 +1,69 @@
...
+		  end: offset
+		  destination: filename
+
+Mandatory properties:
+- destination -- where to store the extracted part
+
+Optional properties:
+- partname -- to extract a full partition, matching on partname is the easiest
+
+- start -- offset in bytes or in sector numbers, e.g. 256s. This becomes the starting
```
obbardc: personally I'd rather just rely on partname, can't see where start/end would be useful. But maybe i am missing some usecase.

actions/dissect_image_action.go
```diff
@@ -0,0 +1,69 @@
...
+
+	# Yaml syntax for parts:
+	  parts:
+	    - partname: partition name
+		  start: offset
+		  end: offset
+		  destination: filename
+
+Mandatory properties:
+- destination -- where to store the extracted part
```
obbardc: would it be interesting to also add compression? but perhaps compression would be better ran manually in a postprocess run action

actions/dissect_image_action.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"github.com/go-debos/debos"
+)
+
+type ImagePart struct {
+	PartName	string
+	StartOffset string
+	EndOffset	string
+	// TBD
```
obbardc: remove tdb, placeholder doesn't add much value I don't think

actions/dissect_image_action.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+type ImageDissectAction struct {
+	debos.BaseAction    `yaml:",inline"`
+	Parts			 []ImagePart
+	// TBD
```
obbardc: remove tdb, placeholder doesn't add much value I don't think

actions/dissect_image_action.go
```diff
@@ -0,0 +1,69 @@
...
+type ImagePart struct {
+	PartName	string
+	StartOffset string
+	EndOffset	string
+	// TBD
+}
+
+type ImageDissectAction struct {
+	debos.BaseAction    `yaml:",inline"`
+	Parts			 []ImagePart
```
obbardc: nit: spacing

actions/dissect_image_action.go
```diff
@@ -0,0 +1,69 @@
...
+	// TBD
+}
+
+func (dis *ImageDissectAction) Verify(ctx *debos.Context) error {
+	// TODO
```
obbardc: you could verify:
- partitions exist in the image
- an `image-partition` action has already been ran before this one (there is already some code which does this in debos, I believe)

actions/dissect_image_action.go
```diff
@@ -0,0 +1,69 @@
...
+	return nil
+}
+
+func (dis *ImageDissectAction) PostMachine(ctx *debos.Context) error {
```
obbardc: you need a run function too, i think ?

obbardc (commented):
small thoughts/comments

actions/dissect_image_action.go
tobsan: If we want to be able to do things like extracting u-boot for some reason, which one typically writes into unpartitioned space, this would be useful. But perhaps that could be saved for a future improvements to keep this version simple.

actions/dissect_image_action.go
tobsan: Compression would be good to add, sure.

actions/dissect_image_action.go
tobsan: Yeah, the idea with this initial push was just to get feedback on the yaml/structure.

actions/dissect_image_action.go
obbardc: I prefer the name `image-extract`! Let;s do it.

obbardc:
@tobsan do you plan to push this forward ? I'd quite like this feature in debos :-). Let me know if I can do anything to help.

tobsan:
> @tobsan do you plan to push this forward ? I'd quite like this feature in debos :-). Let me know if I can do anything to help.

Yes, I've just been too busy with work and foss-north. But since foss-north is next week, I hope to have some more time after that to get this in!

actions/dissect_image_action.go
tobsan: Looks like none of those can be checked at Verify time, but I do check both in the Run function.

actions/dissect_image_action.go
tobsan: With streaming out from the image, adding compression requires some more work. I'm leaving it for now but if we give the archiver module some love it should be doable.

actions/dissect_image_action.go
tobsan: Kept it simple with only partname. We can add offsets later if needed.

actions/image_extract_action.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+	}
+
+	// Can't verify partname existence at this point as ImagePartition metadata
+	// isn't added to the Context until its Run part
+	return nil
+}
+
+func (ext *ImageExtractAction) Run(context *debos.Context) error {
+	if context.ImageFSTab.Len() == 0 {
```
obbardc, line 66: The exact check is used in `actions/filesystem_deploy_action.go` - I wonder if it makes sense to extract this into a helper && make the error more descriptive?

Also, it may be better rework it by setting some boolean in `context` (e.g. `context.ImageCreated`) once the `image-partition` action has ran? ALSO, while thinking out loud, should we also use that boolean to prevent multiple `image-partition` actions to be ran?

actions/image_extract_action.go
```diff
@@ -0,0 +1,118 @@
+/*
+ImageExtract Action
+
+Extract allows to pick out parts of the created image from the ImagePartition
```
obbardc, line 4:
```suggestion
This action extracts raw partitions from the image created using the ImagePartition action and saves those partitions as raw image files in the artefact directory. In a way, this is the Raw action in reverse.
```

actions/image_extract_action.go
```diff
@@ -0,0 +1,118 @@
...
+and FilesystemDeploy actions, and save those parts as raw image files. In a way,
+this is the Raw action in reverse.
+
+	# Yaml syntax:
+	- action: image-extract
+	  parts:
```
obbardc, line 10:
```suggestion
	  partitions:
```

actions/image_extract_action.go
```diff
@@ -0,0 +1,118 @@
...
+	  	<list of parts>
```
obbardc, line 11:
```suggestion
	  	<list of partitions>
```

actions/image_extract_action.go
```diff
@@ -0,0 +1,118 @@
...
+
+Mandatory properties:
+- parts -- list of parts of the image to extract.
```
obbardc, line 14:
```suggestion

- partitions -- list of partitions to extract from the image.
```

actions/image_extract_action.go
```diff
@@ -0,0 +1,118 @@
...
+
+Optional properties:
```
obbardc, line 16: remove this, no optional properties

actions/image_extract_action.go
```diff
@@ -0,0 +1,118 @@
...
+
+	# Yaml syntax for parts:
```
obbardc, line 18:
```
	   # Yaml syntax for partitions:
	   partitions:
	     - name: partition name
		   destination: filename
```

actions/image_extract_action.go
```diff
@@ -0,0 +1,118 @@
...
+Extract allows to pick out parts of the created image from the ImagePartition
+and FilesystemDeploy actions, and save those parts as raw image files. In a way,
+this is the Raw action in reverse.
+
+	# Yaml syntax:
+	- action: image-extract
+	  parts:
+	  	<list of parts>
+
+Mandatory properties:
```
obbardc, line 13: nit: newline after

actions/image_extract_action.go
```diff
@@ -0,0 +1,118 @@
...
+- parts -- list of parts of the image to extract.
+
+Optional properties:
+
+	# Yaml syntax for parts:
+	  parts:
+	    - partname: partition name
+		  destination: filename
+
+Mandatory properties:
```
obbardc, line 23: nit: newline after

actions/image_extract_action.go
```diff
@@ -0,0 +1,118 @@
...
+- destination -- where to store the extracted part
```
obbardc, line 24:
```suggestion
- destination -- path to store the extracted partition image, relative to the artifact directory (with `.img` suffix)
```

Should we check there is a `.img` suffix and error out if not ?

I also wonder whether this should be optional, and if not set, set to `name` && `.img` ?

actions/image_extract_action.go
```diff
@@ -0,0 +1,118 @@
...
+
+- partname -- to extract a full partition, matching on partname is the easiest
```
obbardc, line 26: should go before `destination`.


```suggestion
- partname -- the unique partition name identifier from the `image-partition` action.
```

actions/image_extract_action.go
```diff
@@ -0,0 +1,118 @@
...
+		if len(part.PartName) == 0 {
+			return fmt.Errorf("partname is a mandatory property")
+		}
+
+		if len(part.Destination) == 0 {
+			return fmt.Errorf("destination is a mandatory property")
+		}
+	}
+
+	// Can't verify partname existence at this point as ImagePartition metadata
```
obbardc, line 60:
```suggestion
	// Can't verify partitions exist at this point as ImagePartition metadata isn't set until it has been ran.
```

Also move up to the beginning of the function.

actions/image_extract_action.go
```diff
@@ -0,0 +1,118 @@
...
+	Destination string
+}
+
+type ImageExtractAction struct {
+	debos.BaseAction `yaml:",inline"`
+	Parts            []ImagePart
+}
+
+func (ext *ImageExtractAction) Verify(_ *debos.Context) error {
+	for _, part := range ext.Parts {
```
obbardc, line 50: should we error out if `ext.Parts` is empty ?

actions/image_extract_action.go
```diff
@@ -0,0 +1,118 @@
...
+			return fmt.Errorf("failed to open %s: %w", devicePath, err)
+		}
+		defer source.Close()
+
+		fi, err := source.Stat()
+		if err != nil {
+			return fmt.Errorf("failed to stat %s: %w", devicePath, err)
+		}
+
+		destPath := path.Join(origin, part.Destination)
```
obbardc, line 99: Think we need to unmount the partition && sync everything before copying ?

OR do this after the unmount has been done ? iirc the image is unmounted in the `image-partition` action

actions/image_extract_action.go
```diff
@@ -0,0 +1,118 @@
...
+		dest, err := os.OpenFile(destPath, os.O_WRONLY, 0)
+		if err != nil {
+			return fmt.Errorf("failed to open %s: %w", part.Destination, err)
+		}
+		defer dest.Close()
+
+		bytesCopied, err := io.Copy(dest, source)
```
obbardc, line 106: iirc `io.Copy` uses a small buffer - is it worth increasing it ? will need to do some testing I think

actions/recipe.go
```diff
@@ -73,6 +73,8 @@ Optional properties for recipe:
 
 - image-partition -- https://godoc.org/github.com/go-debos/debos/actions#hdr-ImagePartition_Action
 
+- image-extract -- https://godoc.org/github.com/go-debos/debos/actions#hdr-ImageExtract_Action
```
obbardc, line 76: need to add a test recipe, possibly create some images with known data && check the output ?

actions/image_extract_action.go
```diff
@@ -0,0 +1,118 @@
+/*
+ImageExtract Action
```
obbardc, line 2: I also wonder whether it'd make sense to be able to run this action without an image-partition action, not sure how that would work though?

I just think it's a bit of a hack creating a disk image, to create partition images in cases where you don't need a complete disk image.

See for instance this previous implementation for background: https://github.com/go-debos/debos/pull/94

obbardc (changes requested):
some small changes!

actions/image_extract_action.go
tobsan: Yep, that would make sense to me. Especially since one can do a partitioning **without** setting up an fstab on purpose, so the check feels semantically wrong. I'll see what I can do.

actions/image_extract_action.go
tobsan: I'd rather make it optional with the default you suggested, than force a `.img` suffix.

actions/image_extract_action.go
tobsan: We should indeed, it should be easy to do things correctly, and extracting **no** partitions with this action is incorrect.

PR #695, opened 2026-04-18
Convert go unit tests to recipes

### LINKED ISSUE #672: Convert action go tests to recipes
We have a few actions which have go unit tests:

- https://github.com/go-debos/debos/blob/main/actions/download_action_test.go
- https://github.com/go-debos/debos/blob/main/actions/raw_action_test.go
- https://github.com/go-debos/debos/blob/main/actions/recipe_test.go

I wonder if it could be possible/worth it to convert these into actual recipes and test the actions that way?

obbardc:
Convert some go unit tests to recipe tests.

Closes: #672


obbardc:
The recipe tests aren't nearly similar enough coverage to the unit tests. Will try to fix them or close the bug/issue if I can't fix them.

obbardc:
The list of tests to cover are:



download tests:
- [x] Correct sha256sum
- [x] Incorrect sha256sum
- [x] Invalid sha256sum length in Verify
- [x] Invalid hex characters in Verify
- [x] No sha256sum provided

raw tests:
- [x] Raw action without origin (should default to recipe directory)
- [x] Raw action with explicit 'recipe' origin
- [x] Raw action without source property should fail
- [x] Raw action with missing origin (e.g. nonexistent origin rather than missing property) should fail

PR #696, opened 2026-04-19
image-partition: add mountidentifier for fstab and kernel root= identification

### LINKED ISSUE #409: It should be possible to use a label or device name for a mount in the image partition action
Currently debos uses UUIDs for the `/etc/fstab` entries. However, in some cases when using an A/B system upgrade machanism this can lead to the problem that partitions can't be mounted on boot. See related issue for RAUC: https://github.com/rauc/rauc/issues/1149

The image partition action should have an option to use either the label (`LABEL=...`) or the device name for the specific entry in the `/etc/fstab`.

obbardc:
Currently `/etc/fstab` entries and the kernel `root=` argument are generated using the partition's `UUID=<fsuuid>` which can be quite limiting.

Add a new per-partition property `mountidentifier` which controls how a partition is referenced when generating `/etc/fstab` and the kernel `root=` cmdline argument.

Accepted values are:

 - `uuid`:      `UUID=<fsuuid>`  (default if unset; original behaviour)
 - `partuuid`:  `PARTUUID=<partuuid>`
 - `label`:     `LABEL=<fslabel>`
 - `partlabel`: `PARTLABEL=<partlabel>`

Fixes: #409

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -60,6 +60,7 @@ a 32 bits hexadecimal number (e.g. '1234ABCD' without any dash separator).
 		   fsuuid: string
 		   partuuid: string
 		   partattrs: list of partition attribute bits to set
+		   mountidentifier: how the partition is identified in `/etc/fstab`
 
 Mandatory properties:
 
@@ -124,6 +125,12 @@ A version 5 UUID can be easily generated using the uuid5 template function
 {{ uuid5 $namespace $data }} $namespace should be a valid UUID and $data can be
 any string, to generate reproducible UUID value pass a fixed value of namespace
 and data.
+
+- mountidentifier -- how to identify this partition in `/etc/fstab` and the kernel
+`root=` argument. Valid values: `uuid` (default, uses fsuuid), `partuuid` (uses
+partuuid, GPT only), `label` (uses fslabel), `partlabel` (uses partlabel, GPT only).
+For `label` and `partlabel`, if the `fslabel` or `partlabel` properties are not
+explicitly set, the partition name is used.
 
 - extendedoptions -- list of additional filesystem extended options which need
 to be enabled for the partition.
@@ -217,6 +224,7 @@ type Partition struct {
 	ExtendedOptions []string
 	Fsck            bool `yaml:"fsck"`
 	FSUUID          string
+	MountIdentifier string `yaml:"mountidentifier"`
 }
 
 type Mountpoint struct {
@@ -271,6 +279,33 @@ func (p *Partition) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return nil
 }
 
+func partitionIdentifier(p *Partition) (string, error) {
+	switch p.MountIdentifier {
+	case "", "uuid":
+		if p.FSUUID == "" {
+			return "", fmt.Errorf("partition '%s' has no fsuuid (required for mountidentifier 'uuid')", p.Name)
+		}
+		return "UUID=" + p.FSUUID, nil
+	case "partuuid":
+		if p.PartUUID == "" {
+			return "", fmt.Errorf("partition '%s' has no partuuid (required for mountidentifier 'partuuid')", p.Name)
+		}
+		return "PARTUUID=" + p.PartUUID, nil
+	case "label":
+		if p.FSLabel == "" {
+			return "", fmt.Errorf("partition '%s' has no fslabel (required for mountidentifier 'label')", p.Name)
+		}
+		return "LABEL=" + p.FSLabel, nil
+	case "partlabel":
+		if p.PartLabel == "" {
+			return "", fmt.Errorf("partition '%s' has no partlabel (required for mountidentifier 'partlabel')", p.Name)
+		}
+		return "PARTLABEL=" + p.PartLabel, nil
+	default:
+		return "", fmt.Errorf("partition '%s' has unknown mountidentifier '%s' (valid: uuid, partuuid, label, partlabel)", p.Name, p.MountIdentifier)
+	}
+}
+
 func (i *ImagePartitionAction) generateFSTab(context *debos.Context) error {
 	context.ImageFSTab.Reset()
 
@@ -281,8 +316,10 @@ func (i *ImagePartitionAction) generateFSTab(context *debos.Context) error {
 			/* Do not need to add mount point into fstab */
 			continue
 		}
-		if m.part.FSUUID == "" {
-			return fmt.Errorf("missing fs UUID for partition %s", m.part.Name)
+
+		identifier, err := partitionIdentifier(m.part)
+		if err != nil {
+			return err
 		}
 
 		fsPassno := 0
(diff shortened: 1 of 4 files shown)


actions/image_partition_action.go
```diff
@@ -301,21 +338,23 @@ func (i *ImagePartitionAction) generateFSTab(context *debos.Context) error {
...
-			m.part.FSUUID, m.Mountpoint, fsType,
+		context.ImageFSTab.WriteString(fmt.Sprintf("%s\t%s\t%s\t%s\t0\t%d\n",
+			identifier, m.Mountpoint, fsType,
 			strings.Join(options, ","), fsPassno))
 	}
 
 	return nil
 }
 
+/* TODO: Is it worth allowing a different identifier for `root=` and `/etc/fstab` ?! */
```
obbardc: @sjoerdsimons any idea if it makes sense to allow different identifiers for /etc/fstab and cmdline ? I guess that would be another property, possibly in the `filesystem-deploy` action.

sjoerdsimons: in theory yes; in practise i don't think i've ever needed it so feel free to drop that TODO

obbardc: done. i will create an issue about it once this is merged.

actions/image_partition_action.go
```diff
@@ -217,6 +224,7 @@ type Partition struct {
 	ExtendedOptions []string
 	Fsck            bool `yaml:"fsck"`
 	FSUUID          string
+	MountIdentifier string `yaml:"mountidentifier"`
```
sjoerdsimons, line 227: I guess go doesn't support enum types to be deserialized :/

sjoerdsimons (commented):
Would be good to have tests for this as well

actions/image_partition_action.go
obbardc: It could be done with a custom `UnmarshalYAML` BUT it's probably not worth it vs the current implementation.

Here's a git diff:

```
diff --git a/actions/image_partition_action.go b/actions/image_partition_action.go
index 5c51cf4..6de2d6c 100644
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -208,6 +208,29 @@ import (
        "github.com/go-debos/debos"
 )
 
+type MountIdentifier string
+
+const (
+       MountIdentifierUUID      MountIdentifier = "uuid"
+       MountIdentifierPartUUID  MountIdentifier = "partuuid"
+       MountIdentifierLabel     MountIdentifier = "label"
+       MountIdentifierPartLabel MountIdentifier = "partlabel"
+)
+
+func (m *MountIdentifier) UnmarshalYAML(unmarshal func(interface{}) error) error {
+       var s string
+       if err := unmarshal(&s); err != nil {
+               return err
+       }
+       switch MountIdentifier(s) {
+       case MountIdentifierUUID, MountIdentifierPartUUID, MountIdentifierLabel, MountIdentifierPartLabel:
+               *m = MountIdentifier(s)
+       default:
+               return fmt.Errorf("unknown mountidentifier '%s' (valid: uuid, partuuid, label, partlabel)", s)
+       }
+       return nil
+}
+
 type Partition struct {
        number          int
        Name            string
@@ -224,7 +247,7 @@ type Partition struct {
        ExtendedOptions []string
        Fsck            bool `yaml:"fsck"`
        FSUUID          string
-       MountIdentifier string `yaml:"mountidentifier"`
+       MountIdentifier MountIdentifier `yaml:"mountidentifier"`
 }
 
 type Mountpoint struct {
@@ -281,28 +304,28 @@ func (p *Partition) UnmarshalYAML(unmarshal func(interface{}) error) error {
 
 func partitionIdentifier(p *Partition) (string, error) {
        switch p.MountIdentifier {
-       case "", "uuid":
+       case "", MountIdentifierUUID:
                if p.FSUUID == "" {
                        return "", fmt.Errorf("partition '%s' has no fsuuid (required for mountidentifier 'uuid')", p.Name)
                }
                return "UUID=" + p.FSUUID, nil
-       case "partuuid":
+       case MountIdentifierPartUUID:
                if p.PartUUID == "" {
                        return "", fmt.Errorf("partition '%s' has no partuuid (required for mountidentifier 'partuuid')", p.Name)
                }
                return "PARTUUID=" + p.PartUUID, nil
-       case "label":
+       case MountIdentifierLabel:
                if p.FSLabel == "" {
                        return "", fmt.Errorf("partition '%s' has no fslabel (required for mountidentifier 'label')", p.Name)
                }
                return "LABEL=" + p.FSLabel, nil
-       case "partlabel":
+       case MountIdentifierPartLabel:
                if p.PartLabel == "" {
                        return "", fmt.Errorf("partition '%s' has no partlabel (required for mountidentifier 'partlabel')", p.Name)
                }
                return "PARTLABEL=" + p.PartLabel, nil
        default:
-               return "", fmt.Errorf("partition '%s' has unknown mountidentifier '%s' (valid: uuid, partuuid, label, partlabel)", p.Name, p.MountIdentifier)
+               return "", fmt.Errorf("partition '%s' has unknown mountidentifier '%s'", p.Name, p.MountIdentifier)
        }
 }
 
@@ -901,12 +924,6 @@ func (i *ImagePartitionAction) Verify(_ *debos.Context) error {
                        }
                }
 
-               switch p.MountIdentifier {
-               case "", "uuid", "partuuid", "label", "partlabel":
-               default:
-                       return fmt.Errorf("partition '%s' has unknown mountidentifier '%s' (valid: uuid, partuuid, label, partlabel)", p.Name, p.MountIdentifier)
-               }
-
                if i.PartitionType != "gpt" && p.PartLabel != "" {
                        return fmt.Errorf("can only set partition partlabel on GPT filesystem")
                }
```

sjoerdsimons: Yeah.. seems a bit too much; One more point for Rust i guess ;)

obbardc:
> Would be good to have tests for this as well

@sjoerdsimons I added a quite basic test for this to the partitioning recipe. Please check.

At the same time I found that the `partitioning` and `msdos` tests fail but the error is silently ignored && also failure on non-deterministic partition UUIDs.

I fixed the tests as part of this PR and added the commits at the beginning of the PR - but happy to move that to a different PR if you like?

tests/msdos/test.yaml
```diff
@@ -53,4 +53,7 @@ actions:
   - action: run
     description: Compare expected and actual
     chroot: false
-    command: bash -c 'diff -u <(jq . ${RECIPEDIR}/expected.json) <(jq . ${RECIPEDIR}/actual.json)'
+    command: |
+      jq . ${RECIPEDIR}/expected.json > ${SCRATCHDIR}/expected.json
```
obbardc: the other pattern here could be to not check the uuids of those partitions without any set, but that will need additional code:
```
+        <(jq "del(.partitiontable.partitions[].uuid)" "${RECIPEDIR}/expected.json") \
+        <(jq "del(.partitiontable.partitions[].uuid)" "${RECIPEDIR}/actual.json")'
```

PR #697, opened 2026-04-19
Clean up Docker file layout and documentation

obbardc:
This PR simplifies the Docker-related structure and documentation:

- Move the Dockerfile from docker/ to the repository root to align with
  other open-source projects.
- Consolidate the Docker README into the main README so all
  documentation is in a single, easy-to-find location.

No functional changes intended. This is purely a documentation cleanup.


README.md
```diff
@@ -82,7 +82,28 @@ An official debos container is available:
 docker pull godebos/debos
 ```
 
-See [docker/README.md](docker/README.md) for usage.
```
sjoerdsimons, line 85: Might be better to refer to doc/man/debos.md to avoid du[plication? I'm sure it'll go out of sync if not

obbardc: Not sure what you mean here? There is nothing to go out of sync?

The man page is check that it is updated as part of CI && as a side note I think we should also remove the intermediate md files, but that's another point.

sjoerdsimons: Ah right; it's just a copy; them being checked in i just confusing :)

PR #705, opened 2026-04-21
docs: clarify -t flag variables use dot notation not $var syntax

### LINKED ISSUE #356: yaml template syntax could use more idiot-proofing by raising errors
(Running debos 1.0.0+git20201203.e939090-4+b3, from Debian 11's repos)

...or if you want to be nice to me, normal person spoofing.

I spent an evening struggling with debos. I feel like I lost 75% of my time because mistakes I made would result in lines simply being ignored by the tool instead of causing an error. All my problems were caused by template variables.

**Hard to understand variable assignment syntax** 

Here I messed up, I started from the recipe example and renamed image to image_name_ext4, but forgot to change the .image before the printf. End-result? The variable is set to WARNING_UNSET.

```
$ cat test.yaml
{{- $architecture := or .architecture "amd64" -}}
{{- $suite := or .suite "buster" -}}
{{- $type := or .type "min" -}}
{{- $image_name_ext4 := or .image (printf "WARNING_UNSET_%s_%s.ext4" $suite $architecture) -}}
architecture: {{ $architecture }}
actions:
  - action: run
    chroot: false
    description: Checking image name
    command: echo {{ $image_name_ext4 }}

$ debos --print-recipe --dry-run -t "image_name_ext4":"image_1.1.ext4" test.yaml
2022/07/23 14:34:30 Recipe '/home/user/dev/test.yaml':
2022/07/23 14:34:30 architecture: amd64
actions:
  - action: run
    chroot: false
    description: Checking image name
    command: echo WARNING_UNSET_buster_amd64.ext4
2022/07/23 14:34:30 ==== Recipe done (Dry run) ====
```

This syntax is confusing to me. I feel like image_name_ext4 should either be empty/unset because I set a different hidden variable called image to WARNING_UNSET, or throw an error.

**silently dropping the whole line if missing a hyphen in that template syntax**
When I was editing, I left out the hyphen before the closing brackets on a few variables and wasted a lot of time figuring out why. Is there a way to detect this sort of thing? Why ignore the whole line instead of showing an error, or at least a warning? 

**Nice to have 1**: it would be nice if --verbose would print the value of all template variables at the start. I'm write bogus run actions so I can write my variables in an echo. Personally I think even in non-verbose mode it should print the template variables at the start, so they can be logged.

**Nice to have 2**: please add a --version flag to see what version we're running

obbardc:
Variables injected to recipes via `-t NAME:VALUE` are accessed as `.NAME` in recipe templates rather than `$NAME`. Add an explicit example to the recipe package doc to distinguish `-t` flag variables (`.NAME`) from recipe-local variables (`$NAME`).

Update the help text for the `-t` flag and also update the README and man pages to match the new option.

Fixes: #356

### DIFF
--- a/actions/recipe.go
+++ b/actions/recipe.go
@@ -30,6 +30,18 @@ Comments are allowed and should be prefixed with '#' symbol.
 	    # Use value of variable 'Var' defined above
 	    property2: {{$Var}}
 
+Variables passed via the `-t NAME:VALUE` command line argument are available as
+a field on the template data object, not as local variables:
+
+	# Use a variable passed with `-t Var:Value`
+	property: {{ .Var }}
+
+This is distinct from local variables declared inside the recipe itself,
+which use the `$Var` syntax:
+
+	{{- $Var := "Value" -}}
+	property: {{ $Var }}
+
 The following custom template functions are available:
 
 - sector: Returns the argument with 's' suffix for raw action` (Deprecated)


actions/recipe.go
```diff
@@ -30,6 +30,18 @@ Comments are allowed and should be prefixed with '#' symbol.
 	    # Use value of variable 'Var' defined above
 	    property2: {{$Var}}
 
+Variables passed via the `-t NAME:VALUE` command line argument are available as
+dot-notation fields on the template data object, not as local variables:
```
sjoerdsimons:
```suggestion
a field on the template data object, not as local variables:
```
dot-notation field seems weird? it's just a vfield in the end ;)

sjoerdsimons (commented):
just one small niggle ;)

actions/recipe.go
obbardc: done.

cmd/debos/debos.go
```diff
@@ -107,7 +107,7 @@ func main() {
 		Backend            string            `short:"b" long:"fakemachine-backend" description:"Fakemachine backend to use" default:"auto"`
 		ArtifactDir        string            `long:"artifactdir" description:"Directory for packed archives and ostree repositories (default: current directory)"`
 		InternalImage      string            `long:"internal-image" hidden:"true"`
-		TemplateVars       map[string]string `short:"t" long:"template-var" description:"Template variables (use -t VARIABLE:VALUE syntax)"`
+		TemplateVars       map[string]string `short:"t" long:"template-var" description:"Template variables (use -t NAME:VALUE syntax); accessed in recipes as .NAME (e.g. {{ .NAME }})"`
```
obbardc: @sjoerdsimons I was thinking this extra text seems excessive; what do you think ?

sjoerdsimons: yeah seems a bit over the top

obbardc: removed

PR #715, opened 2026-05-19
Assorted Fixes

richardweinberger:
A few bugs I found while browsing the code.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -165,11 +165,13 @@ func (cmd *Command) restoreResolvConf(sum *[sha256.Size]byte) error {
 	defer os.Remove(savedconf)
 
 	fi, err := os.Lstat(chrootedconf)
-
-	// resolv.conf was removed during the command call
-	// Nothing to do with it -- file has been changed anyway
-	if os.IsNotExist(err) {
-		return nil
+	if err != nil {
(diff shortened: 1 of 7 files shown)


commands.go
```diff
@@ -172,6 +172,10 @@ func (cmd *Command) restoreResolvConf(sum *[sha256.Size]byte) error {
 		return nil
 	}
 
+	if err != nil {
```
obbardc, line 168: I could nitpick this a bit and say that the whole section including the `os.Lstat` call could be reworked a bit to be more readable - if you wanted to fix it while you're here it'd be great.

obbardc (dismissed):
Small nitpick, but you didn't introduce it so happy if you would rather leave it ;-).

commands.go
richardweinberger: Okay, let me look into that. :-)

obbardc:
@richardweinberger any update? This also needs to be rebased onto main.

richardweinberger:
> @richardweinberger any update? This also needs to be rebased onto main.

Just rebased and simplified the code around lstat.

obbardc (approved):
LGTM, thanks!

PR #716, opened 2026-05-20
actions/image-partition: validate parted units for start/end

richardweinberger:
The 'start' and 'end' YAML properties were forwarded to parted(8) verbatim, so typos or unsupported strings only surfaced as parted errors at run time. Validate them in Verify() against parted's documented unit set (s, B, kB, MB, GB, TB, KiB, MiB, GiB, TiB, cyl, chs, %) and reject anything else with an error that lists the canonical forms.

Decimal units (kB, MB, GB, TB) remain accepted for backward compatibility, but now emit a warning recommending the binary equivalents (KiB, MiB, GiB, TiB) so partitions align on power-of-two boundaries.

Match parted's documented casing exactly rather than folding case, so common typos like 'KB' or 'MIB' are reported instead of silently mapped.

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -23,7 +23,7 @@ Mandatory properties:
 
 - imagename -- the name of the image file, relative to the artifact directory.
 
-- imagesize -- generated image size in human-readable form, examples: 100MB, 1GB, etc.
+- imagesize -- generated image size in human-readable form, examples: 100MiB, 1GiB, etc.
 
 - partitiontype -- partition table type. Currently only 'gpt' and 'msdos'
 partition tables are supported.
@@ -75,8 +75,13 @@ unique.
 
 - end -- offset from beginning of the disk there the partition ends.
 
-For 'start' and 'end' properties offset can be written in human readable
-form -- '32MB', '1GB' or as disk percentage -- '100%'.
+For 'start' and 'end' properties the offset must use a unit understood by
+parted(8): 's' (sectors), 'B' (bytes), 'kB', 'MB', 'GB', 'TB' (decimal,
+powers of 1000), 'KiB', 'MiB', 'GiB', 'TiB' (binary, powers of 1024),
+'cyl', 'chs' or '%' (percentage of the device). The value is passed to
+parted as-is; debos only validates the unit and prints a warning when a
+decimal (non-power-of-two) unit is used, since the binary equivalents
+align more naturally on disk sector boundaries.
 
 Optional properties:
 
@@ -157,7 +162,7 @@ Defaults to false.
 	# Layout example for Raspberry PI 3:
 	- action: image-partition
 	  imagename: "debian-rpi3.img"
-	  imagesize: 1GB
+	  imagesize: 1GiB
 	  partitiontype: msdos
 	  mountpoints:
 	    - mountpoint: /
@@ -169,10 +174,10 @@ Defaults to false.
 	    - name: firmware
 	      fs: vfat
 	      start: 0%
-	      end: 64MB
+	      end: 64MiB
 	    - name: root
 	      fs: ext4
-	      start: 64MB
+	      start: 64MiB
 	      end: 100%
 	      flags: [ boot ]
 */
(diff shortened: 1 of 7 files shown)


actions/image_partition_action.go
```diff
@@ -769,6 +774,52 @@ func (i ImagePartitionAction) PostMachineCleanup(context *debos.Context) error {
 	return nil
 }
 
+// validatePartedOffset checks that a partition start/end offset uses a unit
```
obbardc: it'd be great to add a unit test for this, if you can think of any useful tests?

actions/image_partition_action.go
```diff
@@ -769,6 +774,52 @@ func (i ImagePartitionAction) PostMachineCleanup(context *debos.Context) error {
...
+	case "", "s", "B", "%", "cyl", "chs", "KiB", "MiB", "GiB", "TiB":
+		return nil
+	case "kB", "MB", "GB", "TB":
+		log.Printf("Warning: partition %s: %s value %q uses non-power-of-two unit %q; "+
+			"prefer KiB, MiB, GiB or TiB to align on power-of-two boundaries",
+			partName, field, value, unit)
+		return nil
+	default:
+		return fmt.Errorf("partition %s: %s value %q uses unsupported unit %q "+
+			"(parted accepts: s, B, kB, MB, GB, TB, KiB, MiB, GiB, TiB, cyl, chs, %%)",
```
obbardc: nit: error is quite terse; possibly leave this part as a comment && also in the docs at top of the file ?

actions/image_partition_action.go
```diff
@@ -769,6 +774,52 @@ func (i ImagePartitionAction) PostMachineCleanup(context *debos.Context) error {
...
+	if _, err := strconv.ParseFloat(number, 64); err != nil {
+		return fmt.Errorf("partition %s: %s value %q is not a number", partName, field, value)
+	}
+
+	switch unit {
+	case "", "s", "B", "%", "cyl", "chs", "KiB", "MiB", "GiB", "TiB":
+		return nil
+	case "kB", "MB", "GB", "TB":
+		log.Printf("Warning: partition %s: %s value %q uses non-power-of-two unit %q; "+
+			"prefer KiB, MiB, GiB or TiB to align on power-of-two boundaries",
```
obbardc, line 829: nit: error is quite terse; possibly leave this part as a comment && also in the docs at top of the file ?

actions/image_partition_action.go
richardweinberger: Good point. I'll give it a try.

actions/image_partition_action.go
richardweinberger: I'd love to keep the error, or at least as warning. The reason behind this change is real,  I had a stupid typo in my recipe and parted guessed wrong what I meant. As a consequence my partition layout ended up as garbage...

obbardc: Sorry, I meant keep the errors, just not include the `parted accepts: long string` part in the error text, instead keep that as a comment / in the docs :-)

actions/image_partition_action.go
```diff
@@ -906,6 +957,13 @@ func (i *ImagePartitionAction) Verify(_ *debos.Context) error {
 			return fmt.Errorf("partition %s missing end", p.Name)
 		}
 
+		if err := validatePartedOffset(p.Name, "start", p.Start); err != nil {
+			return err
+		}
+		if err := validatePartedOffset(p.Name, "end", p.End); err != nil {
```
obbardc: nitpick: newline above possibly

actions/image_partition_action.go
richardweinberger: Ok!

obbardc:
Looks like the existing unit tests need fixing as part of this too :-).
I already have open https://github.com/go-debos/debos/pull/708 for other issues with those tests, you may want to rebase your work on top of that?

obbardc:
@richardweinberger please fix test fails &rebase :-). Small comments to be fixed too.

richardweinberger:
> @richardweinberger please fix test fails &rebase :-). Small comments to be fixed too.

I'm on it. What shall we do about `gpt_gap` and `imagesize`? I kind of forgot about them.
Currently they get parsed by `units.FromHumanSize` and `units.FromHumanSize`.

Maybe it's better to sanitize all size inputs though the go-units package and pass to parted only sizes with sector units?

richardweinberger:
@obbardc It took me some time to fully understand these failures.
Before my changes, the test created a `8G` sized image and specified the last partition with `start` and `end` being `7G` and `8G`.
But the final size of the partition was never `1G` because parted treats these units just as hint, while IEC binary units (GiB, etc...) are treated as exact values. So when using IEC binary units the very last partition should be configured like:
```
      - name: data5
        fs: ext4
        start: 7GiB
        end: 100%
```

I'd update all tests accordingly. Or do you want me to keep non IEC binary units in the tests?

actions/image_partition_action.go
```diff
@@ -23,7 +23,7 @@ Mandatory properties:
 
 - imagename -- the name of the image file, relative to the artifact directory.
 
-- imagesize -- generated image size in human-readable form, examples: 100MB, 1GB, etc.
+- imagesize -- generated image size in human-readable form, examples: 100MiB, 1GiB, etc.
```
obbardc, line 26: I am not sure - this (same for start/end props) looks like a change that will break existing recipes to me :/

obbardc (changes requested):

actions/image_partition_action.go
richardweinberger: No worries. Existing recipes won't break. They are already broken because units are passed as-is no parted.

obbardc: Just curious what will happen if you pass e.g. `100MB, 1GB` to this property ? debos will error out ? That sounds like a breaking change to me, even if it does use the wrong units :/

PR #719, opened 2026-05-24
actions: add template action

### LINKED ISSUE #718: Add support for templating files
It would be useful for debos to support templating files (e.g. copying a file from the recipe and replacing some variables).

This could be done either by adding a `template` action or extending the `overlay` action.

At the moment debos can copy files into the target filesystem using the `overlay` action and can execute commands inside using the `run` action. Currently there is no built-in way to generate a file from a template using variables from the recipe.

This would be useful for files that are mostly static, but need a small amount of build-specific or image-specific substitution. Examples include:

- `/etc/os-release`
- `/etc/hostname`
- `/etc/apt/sources.list`
- systemd unit files
- bootloader configuration
- etc.

Currently this usually requires using a `run` action with shell scripting (e.g. `sed`, heredocs or custom helper scripts). It works but makes recipes harder to read and maintain -- especially when generating multi-line configuration files.

## Suggested behaviour

debos could support a templating mechanism similar to the [Ansible `template` module](https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/template_module.html) where a source template file is rendered into the target filesystem.

For example, a new action:
```yaml
- action: template
  source: files/os-release.in
  destination: /etc/os-release
  variables:
    image_name: "Example"
    image_version: "1.0"
    variant: "${variant}"
````

Or by extending the existing `overlay` action:
```yaml
- action: overlay
  source: overlays/os-release
  destination: /
  template: true
  variables:
    image_name: "Example"
    image_version: "1.0"
    variant: "${variant}"
```

The input file might look like:

```text
NAME="{{ $image_name }}"
VERSION="{{ $image_version }}"
VARIANT="{{ $variant }}"
```

and debos would render it into the destination path during image creation using the [golang template library](https://pkg.go.dev/text/template). I think jinja2 would be too complex, since it'd require another library. We use the Golang template library elsewhere, it makes sense to use it here.

## Implementation options

There are a few possible approaches:

1. Add a dedicated `template` action for rendering a single file.
2. Extend `overlay` so selected files can be rendered as templates while being copied.
3. Support templating only for files with a special suffix, such as `.in`.
4. Allow variables to come from an explicit `variables:` map.

A dedicated action could be much clearer, because `overlay` currently implies a direct copy. However, extending `overlay` could be convenient when installing a tree of configuration files where only some files need substitution.

I think a separate action would be best.

cc @lool

obbardc:
Add a new template action to render a template file from the recipe
directory into the target filesystem and allow variables to be set
in the template from the recipe. Add some test cases for the newly
introduced template action.

Resolves: https://github.com/go-debos/debos/issues/718


actions/template_action.go
```diff
@@ -0,0 +1,122 @@
+/*
+Template Action
+
+Render a text template from the recipe directory into the target filesystem.
+The template is processed using Go's standard text/template package.
+
+	# Yaml syntax:
+	- action: template
+	  source: path/to/template.in
```
obbardc, line 9: Should we allow directories or even wildcards to allow multiple files to be installed ?

actions/template_action.go
```diff
@@ -0,0 +1,122 @@
...
+	  destination: /absolute/path/inside/rootfs
```
obbardc, line 10: Should this be a dir or exact file ?

PR #727, opened 2026-06-22
doc: don't store the generated man page in git

### LINKED ISSUE #395: Don't store man page in git

obbardc:
The man page doc/man/debos.1 is generated from README.md by create_manpage.sh so there is no need to track it in git. The CI job already discards any regenerated debos.1 and only verifies the intermediate debos.md, which remains tracked.

Stop tracking debos.1, add it to .gitignore and drop the now-redundant 'git checkout *.1' from the CI man-page check.

Closes: #395


.github/workflows/ci.yaml
```diff
@@ -47,7 +47,7 @@ jobs:
         version: v2.3.1
 
   man-page:
-    name: Check if man page has been regenerated
+    name: Generate man page
```
obbardc, line 50: need to modify the manpage bits under `CONTRIBUTING.md` and `.github/copilot-instructions.md` for this change too

PR #729, opened 2026-06-22
Small test improvements

obbardc:


sjoerdsimons:
@obbardc ooi did you churn through this by hand or LLM-assisted? If the latter please mention that in the commit message e.g. following https://docs.kernel.org/process/coding-assistants.html#attribution

obbardc:
> @obbardc ooi did you churn through this by hand or LLM-assisted? If the latter please mention that in the commit message e.g. following https://docs.kernel.org/process/coding-assistants.html#attribution

This so far has been a manual labour of love while understanding loads and loads more about golang testing etc.

I did open #730 for adding AI contribution docs, though (and same for the other debos repos).

Issue #734, opened 2026-07-05
[Question] Root on ZFS and ZFSBootMenu


adionditsak:
debos maintainers and contributors,

I am to an increasing degree using root filesystems on ZFS, and have recently experimented with ZFSBootMenu instead of GRUB to make seamless rollback of an entire OS possible.

I use shell scripts to build my own images, but would like to integrate into debos in my own fork. Before sending future PRs, as I do not consider this a "small" feature, I am interested in your, the maintainers, stance on this trajectory?

obbardc:
Hi, thanks for opening an issue to discuss this.

Can you use a `run` action with a shell script to integrate it?  I guess it'd be good to see what the recipe/debos patches look like first, I can then help guide you to integrate the generic parts into Debos :-)

adionditsak:
@obbardc Thanks for the quick response, and good idea.

I think it could make sense to add it as a `fs: zfs` under `partitions`, but please have a look at the below recipe (here set as `none` and demo with debian trixie). Eventually we can also start with zfs as a filesystem, and then look at ZFSBootMenu later on, to do first things first, but below is for both. 

ZFS does have lots of properties can be set, both for zpools, datasets, and kernel parameters. The ones in the shell scripts below are retrieved from the OpenZFS reference (in the bottom of this comment), but could potentially be set as YAML.

Nevertheless here is the requested:

**root-on-zfs.yaml:**

```
architecture: amd64

actions:
  - action: debootstrap
    suite: trixie
    components: [ main, contrib ]
    mirror: https://deb.debian.org/debian
    variant: minbase

  - action: apt
    recommends: false
    packages:
      - systemd
      - systemd-sysv
      - udev
      - linux-image-amd64
      - linux-headers-amd64
      - zfs-dkms
      - zfsutils-linux
      - zfs-initramfs
      - dosfstools

  - action: run
    description: prepare root on ZFS
    chroot: true
    script: scripts/setup-zfs.sh

  - action: image-partition
    imagename: debos-zfs.img
    imagesize: 8GB
    partitiontype: gpt
    partitions:
      - name: ESP
        fs: vfat
        start: 1MB
        end: 512MB
        flags: [ boot, esp ]
      - name: root
        fs: none
        start: 512MB
        end: 100%
        parttype: 6a898cc3-1dd2-11b2-99a6-080020736631 # solaris

  - action: run
    description: create zpool, deploy rootfs, install ZFSBootMenu
    chroot: false
[...]

PR #735, opened 2026-07-06
Add ZFS support

adionditsak:
Based on conversation in issue #734 this is the first PR for supporting ZFS filesystems. This is as minimal as possible I could leave it as a first PR, with the support of both root and non-root ZFS datasets/filesystems, while still making it flexible in terms of setting necessary and default properties to zpools and datasets, while still have the option to set more properties.

This is tested both with GRUB and with ZBM (ZFSBootMenu), and started with QEMU. See below recipe for an example.

**Recipe example** (updated 04/08/2026)

```
# Build: sudo ./debos -m 4G --scratchsize 10G zfs-root.yaml
# Test:  qemu-system-x86_64 -machine q35,accel=kvm -cpu host -m 2G \
#          -drive if=pflash,format=raw,readonly=on,file=/usr/share/OVMF/OVMF_CODE_4M.fd \
#          -drive file=ubuntu-zfs-on-root.img,format=raw,if=virtio
#

architecture: amd64

actions:
  - action: debootstrap
    suite: noble
    components: [ main, universe ]
    mirror: http://archive.ubuntu.com/ubuntu
    variant: minbase
    keyring-file: /usr/share/keyrings/ubuntu-archive-keyring.gpg
    # ZFS overlay for hostid and resume config
 
  - action: overlay
    source: overlays/zfs

  # ubuntu has zfs kernel module included with linux-image-generic other debian-based distros is better of with zfs-dkms package
  - action: apt
    recommends: false 
    packages:
      - systemd
      - systemd-sysv
      - systemd-resolved
      - udev
      - linux-image-generic
      - zfsutils-linux
      - zfs-initramfs
      - grub-efi-amd64
      - dosfstools
      - iputils-ping
      - iproute2

  - action: run
    description: enable networking
    chroot: true
    command: |
      set -eu
      systemctl enable systemd-networkd systemd-resolved

  - action: image-partition
    imagename: ubuntu-zfs-on-root.img
    imagesize: 8GB
    partitiontype: gpt
    mountpoints:
      - mountpoint: /boot/efi
        partition: ESP
    partitions:
      - name: ESP
        fs: vfat
        start: 1MB
        end: 512MB
        flags: [ boot, esp ]
      - name: root
        fs: zfs
        start: 512MB
        end: 100%
        # properites below follow the OpenZFS "Root on ZFS" recommendations
        zfs:
          pool: zpool
          bootfs: zpool/ROOT/os
          pool-properties:
            ashift: 12
            autotrim: "on"
          dataset-properties:
            acltype: posixacl
            xattr: sa
            dnodesize: auto
            compression: lz4
            normalization: formD
            atime: "off"
            relatime: "on"
          # boot environment layout
          datasets:
            - name: ROOT
              properties: { canmount: "off", mountpoint: none }
            - name: ROOT/os
              properties: { canmount: noauto, mountpoint: / }
            - name: home
            - name: home/root
              properties: { mountpoint: /root }
            - name: var
              properties: { canmount: "off" }
            - name: var/lib
              properties: { canmount: "off" }
            - name: var/log
            - name: var/spool
            - name: var/cache
              properties: { "com.sun:auto-snapshot": "false" }
          swap:
            name: "swap"
            size: 2G
            properties:
              volblocksize: "4096"
              compression: zle
              logbias: throughput
              sync: always
              primarycache: metadata
              secondarycache: none
              "com.sun:auto-snapshot": "false"
          encryption:
            enabled: true
            keyfile: zfs-keyfile

  - action: filesystem-deploy

  # ZFSBootMenu
  - action: run
    description: install ZFSBootMenu
    chroot: false
    command: |
      set -eu
      ESP="$IMAGEMNTDIR/boot/efi"
      mkdir -p "$ESP/EFI/BOOT" "$ESP/EFI/ZBM"
      if [ -f "$RECIPEDIR/zfsbootmenu.EFI" ]; then
          cp "$RECIPEDIR/zfsbootmenu.EFI" "$ESP/EFI/BOOT/BOOTX64.EFI"
      else
          curl -fL -o "$ESP/EFI/BOOT/BOOTX64.EFI" https://get.zfsbootmenu.org/efi
      fi
      cp "$ESP/EFI/BOOT/BOOTX64.EFI" "$ESP/EFI/ZBM/VMLINUZ-BACKUP.EFI"
      zfs set org.zfsbootmenu:commandline="quiet rw" zpool/ROOT
      echo "root:debos" | chroot "$IMAGEMNTDIR" chpasswd

  # GRUB alternative -- NOTE: incompatible with the encryption block
  # above (GRUB cannot read encrypted datasets), and requires the pool
  # to be restricted to GRUB-readable features:
  #   compatibility: grub2
  #   dataset-properties: dnodesize: legacy (instead of auto)
  # - action: run
  #   description: install GRUB
  #   chroot: false
  #   command: |
  #     set -eu
  #     for d in dev dev/pts proc sys; do
  #         mount --bind "/$d" "$IMAGEMNTDIR/$d"
  #     done
  #     chroot "$IMAGEMNTDIR" grub-install --target=x86_64-efi \
  #         --efi-directory=/boot/efi --removable
  #     chroot "$IMAGEMNTDIR" update-grub
  #     chroot "$IMAGEMNTDIR" sh -c 'echo root:debos | chpasswd'
  #     for d in sys proc dev/pts dev; do
  #         umount "$IMAGEMNTDIR/$d"
  #     done
```

**Overlay**

Overlay files for overlay action can be created with shell commands:

```
# overlay
mkdir -p overlays/zfs/etc/initramfs-tools/conf.d
printf '\x0c\xb1\xba\x00' > overlays/zfs/etc/hostid
echo RESUME=none > overlays/zfs/etc/initramfs-tools/conf.d/resume

mkdir -p overlays/zfs/etc/systemd/network
cat > overlays/zfs/etc/systemd/network/10-dhcp.network <<'EOF'
[Match]
Name=en* eth*

[Network]
DHCP=yes
EOF

# zfs-keyfile
echo "testpassword123" > zfs-keyfile
```


adionditsak:
Here is another working recipe example, but where the root is on ext4, and another filesystem is on ZFS. This one uses GRUB instead of ZBM. This is to give two different ways to build images with ZFS that both works from the additions in this PR.

**Recipe example** (updated 04/08/2026)

```
# Build: sudo ./debos -m 4G --scratchsize 10G zfs-non-root.yaml
# Test:  qemu-system-x86_64 -machine q35,accel=kvm -cpu host -m 2G \
#          -drive if=pflash,format=raw,readonly=on,file=/usr/share/OVMF/OVMF_CODE_4M.fd \
#          -drive file=ubuntu-zfs-zbm.img,format=raw,if=virtio

architecture: amd64

actions:
  - action: debootstrap
    suite: noble
    components: [ main, universe ]
    mirror: http://archive.ubuntu.com/ubuntu
    variant: minbase
    keyring-file: /usr/share/keyrings/ubuntu-archive-keyring.gpg

  # ZFS overlay for hostid and resume config
  - action: overlay
    source: overlays/zfs

  # ubuntu has zfs kernel module included with linux-image-generic other debian-based distros is better of with zfs-dkms package
  - action: apt
    recommends: false
    packages:
      - systemd
      - systemd-sysv
      - systemd-resolved
      - udev
      - linux-image-generic
      - zfsutils-linux
      - zfs-initramfs
      - grub-efi-amd64
      - dosfstools
      - iputils-ping
      - iproute2

  - action: run
    description: enable networking
    chroot: true
    command: |
      set -eu
      systemctl enable systemd-networkd systemd-resolved

  - action: run
    description: enable ZFS import at boot
    chroot: true
    command: systemctl enable zfs-import-scan.service zfs-import.target zfs-mount.service zfs.target

  - action: image-partition
    imagename: ubuntu-zfs.img
    imagesize: 8GB
    partitiontype: gpt
    mountpoints:
      - mountpoint: /boot/efi
        partition: ESP
      - mountpoint: /
        partition: root
    partitions:
      - name: ESP
        fs: vfat
        start: 1MB
        end: 512MB
        flags: [ boot, esp ]
      - name: root
        fs: ext4
        start: 512MB
        end: 4GB
      - name: tank
        fs: zfs
        start: 4GB
        end: 100%
        zfs:
          datasets:
            - name: media
              properties:
                mountpoint: /srv/media

  - action: filesystem-deploy

  - action: run
    description: install GRUB
    chroot: true
    command: grub-install --target=x86_64-efi --efi-directory=/boot/efi --removable && update-grub && echo root:debos | chpasswd

```

adionditsak:
I made a commit on the branch since last commit, but it is purely to keep consistent terminology for ZFS. Using dataset instead of filesystem where sensible, and properties instead of options. The above recipes still work as intended

actions/image_partition_action.go
```diff
@@ -71,6 +71,9 @@ unique.
 
 'none' fs type should be used for partition without filesystem.
 
+'zfs' fs type creates a ZFS pool on the partition instead of formatting a
```
obbardc, line 74: May be good to add a unit test suite for this; I guess that may be not very possible with current infra - we can add a note that it's untested if you like :-).

actions/image_partition_action.go
```diff
@@ -455,7 +596,7 @@ func (i ImagePartitionAction) formatPartition(p *Partition, context debos.Contex
 		}
 	}
 
-	if p.FS != "none" && p.FSUUID == "" {
+	if p.FS != "none" && p.FS != "zfs" && p.FSUUID == "" {
```
obbardc, line 681: this check seems a bit long to me, is there a better way to do it ?

perhaps (don't know if break works here):
OR export into another function maybe ? Just htinking out loud

```
if p.FSUUID == "" {
    if p.FS == "none" {
        break;
    }

    if p.FS == "zfs" {
        break;
    }

/* existing logic */
```

actions/image_partition_action.go
```diff
@@ -708,6 +870,36 @@ func (i ImagePartitionAction) Run(context *debos.Context) error {
...
+		return err
+
+	}
+	defer lock.unlock()
+
+	device := i.getPartitionDevice(p.number, *context)
+	log.Printf("Creating zpool %s on %s", p.ZFS.Pool, device)
+
+	/* create ZFS pool */
+	if err := p.ZFS.createPool(device, context.ImageMntDir, context.RecipeDir); err != nil {
```
obbardc: i guess in `Verify` we should check if ZFS is supported on this build host.

actions/image_partition_action.go
adionditsak: Perhaps we can make another issue/PR for this as I think this could indeed delay the PR for quite some time. I agree tho that it should definitely be done.

Where should we write that it is untested - here in the header?

actions/image_partition_action.go
adionditsak: Great stuff! `break` is only for switch and loops AFAI remember.

If you want this more structured and clear we could create a function for this to determine whether the partition is a "conventional" filesystem ie. one made with mkfs with a uuid, whereas ZFS has its own guid.

such as 

```
func (p Partition) hasConventionalFilesystem() bool {
	switch p.FS {
	case "none", "zfs":
		return false
	}
	return true
}
```

But it feels a bit like filler, maybe nice to keep it as is? If the check becomes more frequent the function would make sense but not sure right now. Also thinking out loud

actions/image_partition_action.go
adionditsak: Good call @obbardc

Will add with:
```
...
		if p.FS == "zfs" {
			if _, err := exec.LookPath("zpool"); err != nil {
				return fmt.Errorf("zfs partitions require zfs on the build host")
			}
...
```

adionditsak: seems like it is working:

```bash
$ sudo ./debos -m 4G --scratchsize 10G ../debos_with_zfs_changes/zfs-ro
ot.yaml 
2026/07/08 23:01:21 Action image-partition failed at stage Verify, error: zfs is required on the build
 host
```

adionditsak: Pushed this to the branch, so I think this comment can be resolved

actions/image_partition_action.go
obbardc: I am happy to write a test for this; should be trivial like the btrfs test I did in #736

I guess we can leave this thread open and create an issue once this has been merged?

actions/image_partition_action.go
obbardc: sure, happy to leave this as-is; it's trivial cleanup work

actions/image_partition_action.go
```diff
@@ -175,17 +178,130 @@ Defaults to false.
...
 	      end: 100%
 	      flags: [ boot ]
+
+ZFS partitions:
+
+A partition with 'fs: zfs' hosts a ZFS pool instead of a plain filesystem.
+
+A pool which sets 'bootfs' provides the root filesystem of the image.
+
+Importing pools at boot is the OS' responsibility (e.g. zfs-import-scan.service).
```
obbardc: Think this comment may be better changed to something like:

debos does not configure the target system to import ZFS pools at boot; this is the responsibility of the target operating system, for example via `zfs-import-scan.service`.

actions/image_partition_action.go
```diff
@@ -175,17 +178,130 @@ Defaults to false.
...
+	        keylocation: prompt
+
+All zfs properties are optional:
+
+- pool -- name of the zpool, defaults to the partition name. Pool names
+must be unique within the image.
+
+- ashift -- pool sector size exponent, defaults to 12 (4 KiB), which is
+recommended as many drives use 4 KiB physical sectors even when reporting
+512 B logical sectors.
```
obbardc:
```suggestion
512 byte logical sectors.
```

actions/image_partition_action.go
```diff
@@ -175,17 +178,130 @@ Defaults to false.
...
+
+- compatibility -- restrict the pool to a zpool-features compatibility
+feature set (e.g. 'openzfs-2.1-linux') for portability to older ZFS
+implementations. Unrestricted by default.
+
+- bootfs -- boot environment dataset providing /, set as the pool's
+'bootfs' property. Setting it makes this the root pool
```
obbardc:
```suggestion
'bootfs' property. Setting it makes this pool the root pool.
```

also what kind of value should this property be? bool/string? worth explaining it

actions/image_partition_action.go
```diff
@@ -175,17 +178,130 @@ Defaults to false.
...
+	partitions:
+	  - name: label
+	    fs: zfs
+	    start: offset
+	    end: offset
+	    zfs:
+	      pool: name
+	      ashift: 12
+	      compatibility: feature-set
+	      bootfs: pool/ROOT/dataset
```
obbardc: is `pool/ROOT/dataset` right here ?

also a lot of these mixes real values (e.g. `12`) and descriptions (e.g. `offset`). I think all of the properties should be changed to the latter.

actions/image_partition_action.go
```diff
@@ -175,17 +178,130 @@ Defaults to false.
...
+- pool -- name of the zpool, defaults to the partition name. Pool names
+must be unique within the image.
+
+- ashift -- pool sector size exponent, defaults to 12 (4 KiB), which is
+recommended as many drives use 4 KiB physical sectors even when reporting
+512 B logical sectors.
+
+- compatibility -- restrict the pool to a zpool-features compatibility
+feature set (e.g. 'openzfs-2.1-linux') for portability to older ZFS
+implementations. Unrestricted by default.
```
obbardc: be worth saying that e.g. an empty value or not including the properties ends up being unrestricted ?

actions/image_partition_action.go
```diff
@@ -175,17 +178,130 @@ Defaults to false.
...
+
+- bootfs -- boot environment dataset providing /, set as the pool's
+'bootfs' property. Setting it makes this the root pool
+
+- pool-properties / dataset-properties -- additional or overriding '-o' /
```
obbardc:
```suggestion
- pool-properties / dataset-properties -- additional properties passed to 'zpool create' with '-o' / '-O'. Also allows overwriting existing properties.
```

actions/image_partition_action.go
```diff
@@ -175,17 +178,130 @@ Defaults to false.
...
+'-O' properties for 'zpool create'
+
+- datasets -- datasets to create. Each entry has a 'name' (relative to
+the pool) and an 'properties' map passed as '-o' properties to 'zfs create'.
```
obbardc:
```suggestion
the pool) and a 'properties' map which are passed to 'zfs create' with '-o'.
```

actions/image_partition_action.go
```diff
@@ -175,17 +178,130 @@ Defaults to false.
...
+For a root pool this replaces the default layout and must contain the
+bootfs dataset.
+
+- swap -- create a swap zvol of 'size' (dataset 'name' defaults to
```
obbardc: explain what size is; human-readable ?

obbardc: also `creates a swap...`

actions/image_partition_action.go
```diff
@@ -175,17 +178,130 @@ Defaults to false.
...
+'swap') with the recommended properties (page-size blocks,
+compression=zle, logbias=throughput, sync=always, primarycache=metadata,
+secondarycache=none, auto-snapshot disabled) and add it to fstab.
```
obbardc:
```suggestion
secondarycache=none, auto-snapshot disabled) and adds it to the target's '/etc/fstab'.
```

actions/image_partition_action.go
```diff
@@ -175,17 +178,130 @@ Defaults to false.
...
+
+- encryption -- enable ZFS native encryption on the pool root. The key
```
obbardc:
```suggestion
- encryption -- enables native ZFS encryption on the root pool. The key
```

actions/image_partition_action.go
```diff
@@ -175,17 +178,130 @@ Defaults to false.
...
+material is read from 'keyfile' (relative to the recipe directory) at
+build time. Afterwards 'keylocation' (default 'prompt') is set so the
```
obbardc: this last sentence doesn't make much sense to me ?

actions/image_partition_action.go
```diff
@@ -198,6 +314,11 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/docker/go-units"
```
obbardc: nitpick: this change isn't related and should be removed from this PR

actions/image_partition_action.go
```diff
@@ -306,10 +428,28 @@ func (i *ImagePartitionAction) generateFSTab(context *debos.Context) error {
 			strings.Join(options, ","), fsPassno))
 	}
 
+	/* add ZFS swap to FSTab if existing */
```
obbardc:
```suggestion
	/* add ZFS swap to fstab */
```

actions/image_partition_action.go
```diff
@@ -649,6 +791,16 @@ func (i ImagePartitionAction) Run(context *debos.Context) error {
 		return fmt.Errorf("failed to create mount directory: %w", err)
 	}
 
+	/* ZFS root pool provides / of the image and must be set up before any other mountpoint is mounted below the image mount directory */
```
obbardc:
```suggestion
	/* if the target rootfs is backed by a ZFS root pool, set it up early */
```

actions/image_partition_action.go
```diff
@@ -689,6 +841,16 @@ func (i ImagePartitionAction) Run(context *debos.Context) error {
 	}
 	lock.unlock()
 
+	/* ZFS pools mount their datasets inside the now-existing ZFS tree */
```
obbardc:
```suggestion
	/* setup ZFS pools */
```

actions/image_partition_action.go
```diff
@@ -708,6 +870,36 @@ func (i ImagePartitionAction) Run(context *debos.Context) error {
 	return nil
 }
 
+/* setupZFSPool creates the zpool on a 'fs: zfs' partition, populates it with datasets and an optional swap zvol and mounts them */
```
obbardc:
```suggestion
/* create a zpool, populate it with datasets, an optional swap zvol and mount it */
```

maybe `it` can be replaced with some better term

actions/image_partition_action.go
```diff
@@ -935,6 +1145,45 @@ func (i *ImagePartitionAction) Verify(_ *debos.Context) error {
...
+			if p.ZFS == nil {
+				p.ZFS = &ZFSConfig{}
+			}
+
+			p.ZFS.setDefaults(p.Name)
+			if err := p.ZFS.verify(); err != nil {
+				return err
+			}
+
+			if ZFSPools[p.ZFS.Pool] {
```
obbardc: this check is a bit odd, how about something like this ?

```
// check for duplicate partitions
		for j := idx + 1; j < len(i.Partitions); j++ {
			if i.Partitions[j].Name == p.Name {
				return fmt.Errorf("partition %s already exists", p.Name)
			}
		}
```

we can then remove the `ZFSPools` var

actions/image_partition_action.go
```diff
@@ -935,6 +1145,45 @@ func (i *ImagePartitionAction) Verify(_ *debos.Context) error {
...
+				return fmt.Errorf("duplicate zfs pool name '%s'", p.ZFS.Pool)
+			}
+			ZFSPools[p.ZFS.Pool] = true
+
+			if p.ZFS.isRootPool() {
```
obbardc: move declaration `ZFSBootFS := ""` before the if statement

actions/image_partition_action.go
```diff
@@ -935,6 +1145,45 @@ func (i *ImagePartitionAction) Verify(_ *debos.Context) error {
...
+				}
+				ZFSBootFS = p.ZFS.BootFS
+			}
+
+			/* Default to the Solaris (ZFS) partition type */
+			if p.PartType == "" {
+				p.PartType = "6a898cc3-1dd2-11b2-99a6-080020736631"
+			}
+
+		} else if p.ZFS != nil {
```
obbardc: possibly move this check before `if p.FS == "zfs" {` and do:

```
if p.ZFS != nil && p.FS != "zfs" {
```

actions/image_partition_action.go
```diff
@@ -969,6 +1218,10 @@ func (i *ImagePartitionAction) Verify(_ *debos.Context) error {
 		if strings.ToLower(m.part.FS) == "none" {
 			return fmt.Errorf("cannot mount %s: filesystem not present", m.Mountpoint)
 		}
+
+		if m.Mountpoint == "/" && ZFSBootFS != "" {
+			return fmt.Errorf("cannot mount: the root filesystem is provided by ZFS bootfs '%s'", ZFSBootFS)
```
obbardc:
```suggestion
			return fmt.Errorf("cannot mount %s: the root filesystem is provided by ZFS bootfs '%s'", m.Mountpoint, ZFSBootFS)
```
keep style consistent

actions/zfs.go
```diff
@@ -0,0 +1,270 @@
+/*
+ZFS filesystem support for the image-partition action
```
obbardc:
```suggestion
ZFS filesystem helper
```

actions/zfs.go
```diff
@@ -0,0 +1,270 @@
+/*
```
obbardc: move this file to `wrapper/zfs_wrapper.go`

actions/zfs.go
```diff
@@ -0,0 +1,270 @@
...
+ZFS filesystem support for the image-partition action
+
+A partition with 'fs: zfs' hosts a ZFS pool instead of a conventional filesystem. The defaults implemented here follow the OpenZFS "Root on ZFS" documentation (https://openzfs.github.io/openzfs-docs/Getting%20Started/Debian/Debian%20Trixie%20Root%20on%20ZFS.html).
```
obbardc: nitpick: these two comments need to not refer to the action yaml syntax and instead use the property names from the structs.

actions/zfs.go
```diff
@@ -0,0 +1,270 @@
...
+			z.Encryption.Keylocation = "prompt"
+		}
+	}
+	if z.Swap != nil && z.Swap.Size != "" && z.Swap.Name == "" {
+		z.Swap.Name = "swap"
+	}
+
+	if z.isRootPool() && len(z.Datasets) == 0 {
+		bootEnv := strings.TrimPrefix(z.BootFS, z.Pool+"/")
+		z.Datasets = []ZFSDataset{
```
obbardc: hmm - is this right to hardcode these datasets ?? Also these defaults seem quite specific ?

I don't know zfs so maybe I am misunderstanding something.

Maybe it is better to force the user to set them themselves?

actions/zfs.go
```diff
@@ -0,0 +1,270 @@
...
+			{Name: "var/cache", Properties: map[string]string{"com.sun:auto-snapshot": "false"}},
+		}
+	}
+}
+
+func (z *ZFSConfig) verify() error {
+	if !zfsNameRegexp.MatchString(z.Pool) {
+		return fmt.Errorf("invalid zfs pool name '%s'", z.Pool)
+	}
+	if z.isRootPool() {
```
obbardc: for all of this file, please add a newline before `if` statements. linebreaks are cheap :-)

wrapper/zfs_wrapper.go
```diff
@@ -0,0 +1,270 @@
...
+				found = true
+			}
+		}
+		if !found {
+			return fmt.Errorf("zfs bootfs '%s' is not part of the dataset list", z.BootFS)
+		}
+	}
+	if z.Encryption != nil && z.Encryption.Enabled {
+		switch z.Encryption.Keyformat {
+		case "passphrase", "hex", "raw":
```
obbardc, line 120: add these values into the image-partition readme (if not there already).

actions/zfs.go
```diff
@@ -0,0 +1,270 @@
...
+			return fmt.Errorf("unsupported zfs encryption keyformat '%s'", z.Encryption.Keyformat)
+		}
+		if z.Encryption.Keyfile == "" {
+			return fmt.Errorf("zfs encryption requires a 'keyfile' at image build time")
+		}
+	}
+	return nil
+}
+
+func sortedProperties(flag string, properties map[string]string) []string {
```
obbardc:
```suggestion
func sortProperties(flag string, properties map[string]string) []string {
```

Maybe worth adding a note as to why we need to sort the properties ?

actions/zfs.go
```diff
@@ -0,0 +1,270 @@
...
+
+	args := []string{}
+	for _, k := range keys {
+		args = append(args, flag, fmt.Sprintf("%s=%s", k, properties[k]))
+	}
+	return args
+}
+
+func (z *ZFSConfig) createPool(device string, altroot string, recipeDir string) error {
+	_ = debos.Command{}.Run("udevadm", "udevadm", "settle")
```
obbardc: you're hiding an error here; check it and return it if there is one :-)

actions/zfs.go
```diff
@@ -0,0 +1,270 @@
...
+		"-o", "secondarycache=none",
+		"-o", "com.sun:auto-snapshot=false",
+		full,
+	}
+	err := debos.Command{}.Run("zfs", cmdline...)
+	if err != nil {
+		return fmt.Errorf("failed to create swap zvol '%s': %w", full, err)
+	}
+
+	_ = debos.Command{}.Run("udevadm", "udevadm", "settle")
```
obbardc: hiding an error again :-)

actions/zfs.go
```diff
@@ -0,0 +1,270 @@
...
+		args = append(args, flag, fmt.Sprintf("%s=%s", k, properties[k]))
+	}
+	return args
+}
+
+func (z *ZFSConfig) createPool(device string, altroot string, recipeDir string) error {
+	_ = debos.Command{}.Run("udevadm", "udevadm", "settle")
+
+	poolProperties := map[string]string{
+		"ashift":    fmt.Sprintf("%d", z.Ashift),
```
obbardc: these default properties seem quite specific ? are they reasonable for all users ? we may want to add them into the image-partition docs.

actions/zfs.go
```diff
@@ -0,0 +1,270 @@
...
+		"cachefile": "none",
+	}
+	if z.Compatibility != "" {
+		poolProperties["compatibility"] = z.Compatibility
+	}
+	for k, v := range z.PoolProperties {
+		poolProperties[k] = v
+	}
+
+	datasetProperties := map[string]string{
```
obbardc: these default properties seem quite specific ? are they reasonable for all users ? we may want to add them into the image-partition docs.

actions/zfs.go
```diff
@@ -0,0 +1,270 @@
...
+
+func (z *ZFSConfig) createSwap() error {
+	if z.Swap == nil || z.Swap.Size == "" {
+		return nil
+	}
+
+	full := z.Pool + "/" + z.Swap.Name
+	blocksize := fmt.Sprintf("%d", os.Getpagesize())
+
+	cmdline := []string{"zfs", "create",
```
obbardc: these default properties seem quite specific ? are they reasonable for all users ? we may want to add them into the image-partition docs.

actions/image_partition_action.go
```diff
@@ -935,6 +1145,45 @@ func (i *ImagePartitionAction) Verify(_ *debos.Context) error {
 		if maxLength > 0 && len(p.FSLabel) > maxLength {
 			return fmt.Errorf("fs label for %s '%s' is too long", p.Name, p.FSLabel)
 		}
+
+		if p.FS == "zfs" {
+			if _, err := exec.LookPath("zpool"); err != nil {
```
obbardc: also need to check for zpool tool ?

obbardc (changes requested):
small nitpicks

obbardc:
There are also lint errors:
```
  actions/image_partition_action.go:879:2: unnecessary trailing newline (whitespace)
  	}
  	^
  actions/image_partition_action.go:1184:3: unnecessary trailing newline (whitespace)
  		} else if p.ZFS != nil {
  		^
  2 issues:
  * whitespace: 2
```

obbardc:
@adionditsak I think it may be worth commiting your recipes from this PR under `tests/partitioning-zfs` ? If not I can do that when I add unit tests.

actions/image_partition_action.go
adionditsak: I agree, and well written replacement. Changed this will push soon

actions/zfs.go
adionditsak: Are you referring to the below timeout? I will have a deeper look now.

`udevadm | Timed out while waiting for udev queue to empty.`

actions/image_partition_action.go
adionditsak: Yes it is right: you can see above it that the dataset define has `name: dataset`. I will change this naming to be more clear so you should not be in doubt.

Actually yes, but 12 is default as per best practices from OpenZFS documentation, but can be, like any other property, set in the recipe if you want another value. Therefore I will remove this from the example as it is not useful. I will look for the others with hard/real values and change them into description.

actions/zfs.go
adionditsak: Will add comment above function:
`sorting ZFS pool and dataset properties to keep a consistent order and thereby comparability and testability throughout builds`

actions/zfs.go
adionditsak: These are not fatal, so is it ok to use `log.Printf()` instead? 

```
	if err := (debos.Command{}).Run("udevadm", "udevadm", "settle"); err != nil {
		log.Printf("Warning: udevadm settle returned: %s", err)
	}
```

actions/zfs.go
adionditsak: These are not fatal, so is it ok to use `log.Printf()` instead? 

```
	if err := (debos.Command{}).Run("udevadm", "udevadm", "settle"); err != nil {
		log.Printf("Warning: udevadm settle returned: %s", err)
	}
```

actions/zfs.go
adionditsak: From OpenZFS documentation (https://openzfs.github.io/openzfs-docs/Getting%20Started/Debian/Debian%20Trixie%20Root%20on%20ZFS.html), `ashift: 12` as default is good to have because:

`The use of ashift=12 is recommended here because many drives today have 4 KiB (or larger) physical sectors, even though they present 512 B logical sectors. Also, a future replacement drive may have 4 KiB physical sectors (in which case ashift=12 is desirable) or 4 KiB logical sectors (in which case ashift=12 is required).`

These can be changed in recipes as both pool an dataset properties as I think is quite clear.

For the image-partition docs, can you more specific what is needed there and where it is needed?

actions/zfs.go
adionditsak: From OpenZFS documentation (https://openzfs.github.io/openzfs-docs/Getting%20Started/Debian/Debian%20Trixie%20Root%20on%20ZFS.html). There are properties that is important to set to default as many users otherwise would go wrong such as `atime:off` (just set this) that gives a performance boost and `acltype:posixacl` that makes Posix ACL's possible on a ZFS dataset. It basically just helps users with a good start with root on ZFS that is usually not a straightfoward task. These can easily be changed in recipes as both pool an dataset properties as I think is quite clear.

For the image-partition docs, can you more specific what is needed there and where it is needed?

actions/zfs.go
adionditsak: Again, refer to the OpenZFS documentation. These properties are especially important to make the zvol fit as a swap device.

Whereas the dataset and pool properties can easily be replaced these can not in the current code. I think it is more a question if the user wants to build with or without this zvol, and not so much whether this should be configured diferently, but I can add the option to configure this in more detail in recipe if you think it is a good idea.

For the image-partition docs, can you more specific what is needed there and where it is needed?

actions/image_partition_action.go
adionditsak: you mean `zfs` tool? We can do that!

actions/image_partition_action.go
adionditsak: `/* create a zpool, populate it with datasets, an optional zvol mounted as a swap device */`

Maybe like this?

actions/image_partition_action.go
adionditsak: This is for the ZFS pools themselves, that contain the ZFS datasets (filesystems). I dont think this is connected or will work. Then we have to do something like:

```
		for j := 0; j < idx; j++ {
				prev := &i.Partitions[j]
				if prev.FS == "zfs" && prev.Zfs.Pool == p.Zfs.Pool {
					return fmt.Errorf("duplicate zfs pool name '%s'", p.Zfs.Pool)
				}
			}
```

?

actions/image_partition_action.go
adionditsak: This is also used in the mountpoints on line 1224-1225 so it has to exist outside the i.Partitions for loop

actions/zfs.go
adionditsak: This is best practice datasets when using root on ZFS that I have read of the OpenZFS documentation, but we can make it less specific if you want. The specifics such as `canmount`, `mountpoint`, etc. is how ZFS datasets are mounted as filesystems, and also canmount makes it not automatically mount as this happens otherwise for the root dataset. Therefore these defaults are strictly necessary in this case. What we can make less specific is the root filesystem structure, but I can not see why we should not do this. Perhaps if future Debian root filesystems look very different

actions/image_partition_action.go
adionditsak: Pushed this

actions/image_partition_action.go
adionditsak: Pushed this

actions/image_partition_action.go
adionditsak: Pushed this

actions/zfs.go
adionditsak: Pushed this

actions/image_partition_action.go
adionditsak: Pushed this

actions/image_partition_action.go
adionditsak:
```
encryption -- enables native ZFS encryption on the pool. The encryption key cannot be prompted for at
build time: instead the key material is read from 'keyfile' (relative
to the recipe directory). The keyfile is only used while building and
is not shipped in the image; once the pool has been created, the
pool's 'keylocation' property is changed from the build keyfile to
'keylocation' (default 'prompt'), so the deployed system asks for the
passphrase at boot instead of looking for the key in a file.
```

Does this make more sense? If yes, then I can push it.

The key and keylocation terms are from ZFS terminology. 

Alternatively it can also be set to a keyfile on the server or a remote http URL. See more at: https://kldload.com/zfs-wiki/encryption

actions/zfs.go
obbardc: Some thoughts here...

1) can you not just call `ImagePartitionAction.triggerDeviceNodes` ? I wonder if that times out too ?

2) Is it normal for `udevadm settle` to timeout with zfs ?

3) To make the error check here really robust; we could check the error code from `udevadm settle` and only mark the timeout as a warning.

`udevadm settle` SHOULD exit with `-ETIMEDOUT` when it fails: (https://github.com/systemd/systemd/blob/1fcd92a21464d5d8466c1990490015793ec5231e/src/udev/udevadm-settle.c#L243):
```
        r = sd_event_loop(event);
        if (r == -ETIMEDOUT)
                return log_error_errno(r, "Timed out while waiting for udev queue to empty.");
        if (r < 0)
                return log_error_errno(r, "Event loop failed: %m");
```

so we could detect that like this:
```
if err := (debos.Command{}).Run("udevadm", "udevadm", "settle"); err != nil {
	var exitErr *exec.ExitError

	if errors.As(err, &exitErr) &&
		exitErr.ExitCode() == (-int(syscall.ETIMEDOUT)&0xff) {
		log.Printf("Warning: timed out while waiting for the udev queue to empty")
	} else {
		return err
	}
}
```

We may also want to do the same check in `ImagePartitionAction.triggerDeviceNodes` ? Or maybe not ?

actions/image_partition_action.go
adionditsak: Pushed this

actions/zfs.go
adionditsak: @obbardc Thanks for your thoughts and pointing out the existing method. 

To respond to your points to keep an overview on this thread.

1. `udevadm settle` waits passively for all events in the udevd device queue to be finished while the trigger command only settles the one triggered from when it is executed. It is a much more ideal command to run for this. I reused the triggerDeviceNodes method, and moved the settle command out of zfs.go createZpool method and into image_partition_action.go setupZFSPool method also keeps a more consistent structure throughout so zfs only contains the wrapper logic and not the partitioning logic. 
2. No this is not related to ZFS, but this can hang for many reasons especially when building on different building hosts. As mentioned in 1) this command is passive and waits for all events in the queue and these can be anything coming from the host building.
3. `udevadm-settle.c` returns `-ETIMEDOUT`, but udevadm's entry point goes through `DEFINE_MAIN_FUNCTION_WITH_POSITIVE_FAILURE`, which maps all negative returns to `EXIT_FAILURE`, so timeout and any other failure both exit 1 so the `-ETIMEDOUT & 0xff` would not match. Correct me if I am wrong: 

udevadm entry point:
[https://github.com/systemd/systemd/blob/main/src/udev/udevadm.c#L124](https://github.com/systemd/systemd/blob/main/src/udev/udevadm.c#L124)

systemd shared method for handling exit:
[https://github.com/systemd/systemd/blob/main/src/shared/main-func.h#L35](https://github.com/systemd/systemd/blob/main/src/shared/main-func.h#L35)

systemd shared method for converting non-zero exit codes:
[https://github.com/systemd/systemd/blob/4e67935b77a0eb8e7d954ba37510355691ed1585/src/shared/main-func.c#L34](https://github.com/systemd/systemd/blob/4e67935b77a0eb8e7d954ba37510355691ed1585/src/shared/main-func.c#L34)

I am not sure it is worth it to make this more detailed as it would require more maintenance if logic in udevadm/systemd changes going forward - more dependence.

4. An extra thing: For the swap device I replaced `udevadm settle` with `zvol_wait` (https://openzfs.github.io/openzfs-docs/man/v2.2/1/zvol_wait.1.html), as this is the exact directory (`/dev/zvol`) we look for the symlink in, and therefore a more ideal way to check this for a zvol.

I have tested the above with existing recipes, timeouts are not occurring anymore, `zvol_wait` returns with: 

```
2026/07/14 07:11:17 zvol_wait | Testing 1 zvol links
2026/07/14 07:11:17 zvol_wait | All zvol links are now present
```

Images build well.

I pushed these changes.

actions/zfs.go
adionditsak: Did this. Except when using `if err != nil`

actions/image_partition_action.go
adionditsak: Which change do you refer to here? I can not see any line additions at line 317 (go-units module)

actions/zfs.go
adionditsak: Lets move the file when we have resolved the other comments?

actions/zfs.go
adionditsak: Pushed this

actions/zfs.go
adionditsak: After having slept on it, I can see the point about code being a bit too detailed in terms of pool and dataset properties as well predefined root tree setup as datasets. I will do a commit now that keeps the absolutely necessary ones such as `mountpoint` and `canmount`, and `cachefile`, and then we can provide the user with reference to the OpenZFS documentation as well as recipe examples to ZFS partitioning for best practices they can implement themselves, so we debos stays generic and we do not have to change source when best practices changes for OpenZFS on Linux

actions/zfs.go
adionditsak: After having slept on it, I can see the point about code being a bit too detailed in terms of pool and dataset properties as well predefined root tree setup as datasets. I will do a commit now that keeps the absolutely necessary ones such as `mountpoint` and `canmount`, and `cachefile`, and then we can provide the user with reference to the OpenZFS documentation as well as recipe examples to ZFS partitioning for best practices they can implement themselves, so we debos stays generic and we do not have to change source when best practices changes for OpenZFS on Linux

actions/zfs.go
adionditsak: After having slept on it, I can see the point about code being a bit too detailed in terms of pool and dataset properties as well predefined root tree setup as datasets. I will do a commit now that keeps the absolutely necessary ones such as `mountpoint` and `canmount`, and `cachefile`, and then we can provide the user with reference to the OpenZFS documentation as well as recipe examples to ZFS partitioning for best practices they can implement themselves, so we debos stays generic and we do not have to change source when best practices changes for OpenZFS on Linux

actions/zfs.go
adionditsak: After having slept on it, I can see the point about code being a bit too detailed in terms of pool and dataset properties as well predefined root tree setup as datasets. I will do a commit now that keeps the absolutely necessary ones such as `mountpoint` and `canmount`, and `cachefile`, and then we can provide the user with reference to the OpenZFS documentation as well as recipe examples to ZFS partitioning for best practices they can implement themselves, so we debos stays generic and we do not have to change source when best practices changes for OpenZFS on Linux

adionditsak:
@obbardc I have made a commit (https://github.com/go-debos/debos/pull/735/commits/7f83bded644be1c219db303482b7ded760df44ff) where I remove the default properties and datasets and then it is up to the users to set these themselves. I have of course kept the strictly necessary properties.

**prerequisities to recipe**

```
# overlay
mkdir -p overlays/zfs/etc/initramfs-tools/conf.d
printf '\x0c\xb1\xba\x00' > overlays/zfs/etc/hostid
echo RESUME=none > overlays/zfs/etc/initramfs-tools/conf.d/resume

mkdir -p overlays/zfs/etc/systemd/network
cat > overlays/zfs/etc/systemd/network/10-dhcp.network <<'EOF'
[Match]
Name=en* eth*

[Network]
DHCP=yes
EOF

# zfs-keyfile
echo "testpassword123" > zfs-keyfile
```

**recipe** 

This is the recipe I test with. We can add some version of this to `/tests/zfs-partitioning` when we start the work with the test suite, as you already pointed out.

Provided an example of encryption as well so it is clear to the user, then they can choose to omit it if they don't need it.

```
# Build: sudo ./debos -m 4G --scratchsize 10G zfs-root.yaml
# Test:  qemu-system-x86_64 -machine q35,accel=kvm -cpu host -m 2G \
#          -drive if=pflash,format=raw,readonly=on,file=/usr/share/OVMF/OVMF_CODE_4M.fd \
#          -drive file=ubuntu-zfs-on-root.img,format=raw,if=virtio
#

architecture: amd64

actions:
  - action: debootstrap
    suite: noble
    components: [ main, universe ]
    mirror: http://archive.ubuntu.com/ubuntu
    variant: minbase
    keyring-file: /usr/share/keyrings/ubuntu-archive-keyring.gpg
    # ZFS overlay for hostid and resume config
 
  - action: overlay
    source: overlays/zfs

  # ubuntu has zfs kernel module included with linux-image-generic other debian-based distros is better of with zfs-dkms package
  - action: apt
    recommends: false 
    packages:
      - systemd
      - systemd-sysv
      - systemd-resolved
      - udev
      - linux-image-generic
      - zfsutils-linux
      - zfs-initramfs
      - grub-efi-amd64
      - dosfstools
      - iputils-ping
      - iproute2

  - action: run
    description: enable networking
    chroot: true
    command: |
      set -eu
      systemctl enable systemd-networkd systemd-resolved

  - action: image-partition
    imagename: ubuntu-zfs-on-root.img
    imagesize: 8GB
    partitiontype: gpt
    mountpoints:
      - mountpoint: /boot/efi
        partition: ESP
    partitions:
      - name: ESP
        fs: vfat
        start: 1MB
        end: 512MB
        flags: [ boot, esp ]
      - name: root
        fs: zfs
        start: 512MB
        end: 100%
        # properites below follow the OpenZFS "Root on ZFS" recommendations
        zfs:
          pool: zpool
          bootfs: zpool/ROOT/os
          pool-properties:
            ashift: 12
            autotrim: "on"
          dataset-properties:
            acltype: posixacl
            xattr: sa
            dnodesize: auto
            compression: lz4
            normalization: formD
            atime: "off"
            relatime: "on"
          # boot environment layout
          datasets:
            - name: ROOT
              properties: { canmount: "off", mountpoint: none }
            - name: ROOT/os
              properties: { canmount: noauto, mountpoint: / }
            - name: home
            - name: home/root
              properties: { mountpoint: /root }
            - name: var
              properties: { canmount: "off" }
            - name: var/lib
              properties: { canmount: "off" }
            - name: var/log
            - name: var/spool
            - name: var/cache
              properties: { "com.sun:auto-snapshot": "false" }
          swap:
            name: "swap"
            size: 2G
            properties:
              volblocksize: "4096"
              compression: zle
              logbias: throughput
              sync: always
              primarycache: metadata
              secondarycache: none
              "com.sun:auto-snapshot": "false"
          encryption:
            enabled: true
            keyfile: zfs-keyfile

  - action: filesystem-deploy

  # ZFSBootMenu
  - action: run
    description: install ZFSBootMenu
    chroot: false
    command: |
      set -eu
      ESP="$IMAGEMNTDIR/boot/efi"
      mkdir -p "$ESP/EFI/BOOT" "$ESP/EFI/ZBM"
      if [ -f "$RECIPEDIR/zfsbootmenu.EFI" ]; then
          cp "$RECIPEDIR/zfsbootmenu.EFI" "$ESP/EFI/BOOT/BOOTX64.EFI"
      else
          curl -fL -o "$ESP/EFI/BOOT/BOOTX64.EFI" https://get.zfsbootmenu.org/efi
      fi
      cp "$ESP/EFI/BOOT/BOOTX64.EFI" "$ESP/EFI/ZBM/VMLINUZ-BACKUP.EFI"
      zfs set org.zfsbootmenu:commandline="quiet rw" zpool/ROOT
      echo "root:debos" | chroot "$IMAGEMNTDIR" chpasswd

  # GRUB alternative -- NOTE: incompatible with the encryption block
  # above (GRUB cannot read encrypted datasets), and requires the pool
  # to be restricted to GRUB-readable features:
  #   compatibility: grub2
  #   dataset-properties: dnodesize: legacy (instead of auto)
  # - action: run
  #   description: install GRUB
  #   chroot: false
  #   command: |
  #     set -eu
  #     for d in dev dev/pts proc sys; do
  #         mount --bind "/$d" "$IMAGEMNTDIR/$d"
  #     done
  #     chroot "$IMAGEMNTDIR" grub-install --target=x86_64-efi \
  #         --efi-directory=/boot/efi --removable
  #     chroot "$IMAGEMNTDIR" update-grub
  #     chroot "$IMAGEMNTDIR" sh -c 'echo root:debos | chpasswd'
  #     for d in sys proc dev/pts dev; do
  #         umount "$IMAGEMNTDIR/$d"
  #     done
```

actions/image_partition_action.go
adionditsak: Done this

actions/zfs.go
adionditsak: Did this

actions/zfs.go
adionditsak: I acted on the other thread on the same topic. Moved this into `image_partitioning_action.go` and reused method `triggerDeviceNodes`

actions/image_partition_action.go
obbardc: this small diff just moves the imports around (possibly automatically done by your IDE?); which is unrelated to your change. It belongs in its own commit or not in this PR.

actions/image_partition_action.go
obbardc: yeah; your change is better than what currently exists; but do replace `prev` with `part` or similar better name.

actions/image_partition_action.go
```diff
@@ -969,6 +1277,10 @@ func (i *ImagePartitionAction) Verify(_ *debos.Context) error {
 		if strings.ToLower(m.part.FS) == "none" {
 			return fmt.Errorf("cannot mount %s: filesystem not present", m.Mountpoint)
 		}
+
+		if m.Mountpoint == "/" && ZFSBootFS != "" {
```
obbardc: does it make more sense to move the `ZFSBootFS != ""` check first ?

actions/image_partition_action.go
obbardc: ah; makes sense. Not really a fan of these kind of very specific "global-ish" variables; can you add a comment to when you declare it to explain what it does?

actions/zfs.go
obbardc: sure; i think it makes sense to move the file out of `actions` though :-)

.gitignore
```diff
@@ -86,3 +86,4 @@ tags
 # Test artifacts
 *.res
 *.img
+*.img.*
```
obbardc: This change belongs in its own commit; possibly in another PR as it's unrelated to this PR.

actions/image_partition_action.go
obbardc: How about this - also includes an example of how to set the properties.

I think we need to make sure all sub-properties for all other properties are documented in a similar way

```
encryption -- native ZFS encryption configuration for the pool.

The encryption key is read from 'keyfile', relative to the recipe directory. This file is used only while creating the pool and is not included in the target filesystem.

After the pool has been created, its ZFS 'keylocation' property is changed from the temporary build-time key file to the configured 'keylocation'. The default value is 'prompt', which causes the target system to request the passphrase when the encrypted dataset is loaded.

Alternatively, 'keylocation' may refer to a key file available on the target system or to a remote HTTP or HTTPS URL. For more information about ZFS encryption keys and the 'keylocation' property, see https://kldload.com/zfs-wiki/encryption.

For example:

          encryption:
            enabled: true
            keyfile: zfs-keyfile
            keylocation: prompt

In this example, 'zfs-keyfile' is read from the recipe directory while the image is being built. The file is not copied into the target filesystem and the deployed system prompts for the encryption key when loading the encrypted dataset.
```

actions/zfs.go
obbardc: thanks; looks good :-)

actions/image_partition_action.go
```diff
@@ -71,6 +71,9 @@ unique.
 
 'none' fs type should be used for partition without filesystem.
 
+'zfs' fs type creates a ZFS pool on the partition instead of formatting a
+conventional filesystem. See "ZFS partitions" section below.
```
obbardc: nitpick:

```suggestion
conventional filesystem. See the 'ZFS partitions' section below.
```

actions/image_partition_action.go
```diff
@@ -175,17 +178,182 @@ Defaults to false.
...
+	        - name: dataset
+	          properties:
+	            property: value
+	      swap:
+	        size: 2G
+	        name: swap
+	        properties:
+	          property: value
+	      encryption:
+	        enabled: true
```
obbardc:
```suggestion
	        enabled: bool
```

actions/image_partition_action.go
```diff
@@ -175,17 +178,182 @@ Defaults to false.
...
+
+- ashift -- pool sector size exponent. If unset, no ashift is passed and
+zfs detects the sector size; the OpenZFS documentation recommends 12
+(4 KiB) as many drives use 4 KiB physical sectors even when reporting
+512 byte logical sectors.
+
+- compatibility -- restrict the pool to a zpool-features compatibility
+feature set (e.g. 'openzfs-2.1-linux') for portability to older ZFS
+implementations. If not including the property or leaving the field empty
+the default is Unrestricted.
```
obbardc:
```suggestion
the default value is 'Unrestricted'.
```

actions/image_partition_action.go
```diff
@@ -175,17 +178,182 @@ Defaults to false.
...
+must be unique within the image.
+
+- ashift -- pool sector size exponent. If unset, no ashift is passed and
+zfs detects the sector size; the OpenZFS documentation recommends 12
+(4 KiB) as many drives use 4 KiB physical sectors even when reporting
+512 byte logical sectors.
+
+- compatibility -- restrict the pool to a zpool-features compatibility
+feature set (e.g. 'openzfs-2.1-linux') for portability to older ZFS
+implementations. If not including the property or leaving the field empty
```
obbardc:
```suggestion
implementations. If field is not included or empty,
```

actions/image_partition_action.go
```diff
@@ -175,17 +178,182 @@ Defaults to false.
...
+the default is Unrestricted.
+
+- bootfs -- boot environment dataset providing /, set as the pool's
```
obbardc:
```suggestion
- bootfs -- boot environment dataset providing '/', set as the pool's
```

actions/image_partition_action.go
```diff
@@ -175,17 +178,182 @@ Defaults to false.
...
+'bootfs' property. Setting it makes this pool the root pool.
+
+- pool-properties / dataset-properties -- additional properties passed
+to 'zpool create' with '-o' / '-O'. Also allows overwriting existing
+properties.
+'-O' properties for 'zpool create'
```
obbardc: this line seems out of place ?

actions/image_partition_action.go
```diff
@@ -175,17 +178,182 @@ Defaults to false.
...
+Required for root pools: the list must contain the bootfs dataset, which
+must set the property mountpoint '/' and may not set canmount 'off' (with
+canmount 'noauto', as recommended for boot environments, the action mounts
+it explicitly). Optional for data pools, where the pool root itself is a
+filesystem.
+
+- swap -- create a zvol for swap memory of 'size' in human-readable
+form, examples: 500MB, 2GB, etc. (dataset 'name' defaults to 'swap') and
+add it to fstab via its /dev/zvol alias. Zvol properties are passed via
+the 'properties' map; the OpenZFS documentation recommends page-size
```
obbardc, line 260: maybe link to documentation here too ?

actions/image_partition_action.go
adionditsak: seems like a copy paste error from my side. Will remove

actions/image_partition_action.go
adionditsak: Now i understand. I use Gofmt (https://pkg.go.dev/cmd/gofmt). Will move it out

actions/image_partition_action.go
adionditsak: I can do that. If it is a no-go, we can make a helper method to avoid declaring this var. I will push the comment and then you can have a look at that first

actions/zfs.go
adionditsak: I agree

actions/image_partition_action.go
adionditsak: This seems very clear. Will add this in the next commit. In terms of the others, I agree that an example of encryption was missing so far, but I think the others are mostly present in the two "layout examples" included further below in this file and that is what they are for? Please feel free to add concrete suggestions like these ones :-)

actions/image_partition_action.go
adionditsak: Pushed this

actions/image_partition_action.go
adionditsak: Pushed this

actions/image_partition_action.go
adionditsak: Pushed this

actions/image_partition_action.go
adionditsak: Pushed this

actions/image_partition_action.go
adionditsak: Pushed this

actions/zfs.go
adionditsak: Pushed this in yesterdays commit

actions/zfs.go
adionditsak: Pushed this in yesterdays commit

actions/zfs.go
adionditsak: Pushed this in yestedays commit

actions/image_partition_action.go
adionditsak: Pushed this

.gitignore
adionditsak: Pushed this

actions/zfs.go
adionditsak: Pushed this in yesterdays commit

actions/zfs.go
adionditsak: Pushed this - see the other conversation on the same topic

actions/image_partition_action.go
```diff
@@ -306,10 +497,28 @@ func (i *ImagePartitionAction) generateFSTab(context *debos.Context) error {
 			strings.Join(options, ","), fsPassno))
 	}
 
+	/* add ZFS swap to fstab */
+	for idx := range i.Partitions {
+		p := &i.Partitions[idx]
+		if p.FS == "zfs" && p.ZFS.Swap != nil && p.ZFS.Swap.Size != "" {
```
obbardc: Should we the check of `p.ZFS.Swap != nil && p.ZFS.Swap.Size != ""` to `Verify` ? Also should we check that `Name` is not empty in `Verify` too ?

That way here we can just do:

```
if p.FS == "zfs" && p.ZFS.Swap != nil {
...
```

actions/image_partition_action.go
adionditsak: @obbardc This is pushed (please resolve these if it looks ok)

actions/image_partition_action.go
adionditsak: @obbardc I pushed this (https://github.com/go-debos/debos/pull/735/changes/10b4cc360c171dc830b47cc300f9ebd1db3a5c09). I moved the checks into the verify() method in zfs.go, that is for checks related to ZFS that does not need FS, partitions or mountpoints.

adionditsak: @obbardc can this be resolved?

actions/zfs.go
adionditsak: @obbardc Since I believe most of the other comments are resolved I will move this file to wrapper now

actions/image_partition_action.go
adionditsak: @obbardc I think this is a good idea. Lets wait to write the test until we have merged if possible, since we then know the form of its first iteration

obbardc: @adionditsak Once you've cleaned the commits up a bit I will do another review this week and I will also create and push some unit tests to your branch (if you dont mind) :-)

adionditsak: @obbardc That sounds perfect. Should I resolve the conversations open that I believe are done, or do you want to do that?

obbardc: Please resolve them if you think they are complete. I will do another review round tomorrow, hopefully nothing major now :-).

actions/zfs.go
adionditsak: @obbardc I moved it to `wrapper/zfs_wrapper.go` and adapted other files accordingly:
https://github.com/go-debos/debos/pull/735/changes/10eccd9943776f6041e1cfe1da2e2f937eb6317e

wrapper/zfs_wrapper.go
adionditsak: I made this clearer now:
https://github.com/go-debos/debos/pull/735/changes/6f9301ed20587d5f372d1e97c9219ae44bf1620f

actions/image_partition_action.go
adionditsak: @obbardc This has been reverted now

actions/image_partition_action.go
adionditsak: @obbardc Resolved the others except this conversation we can keep open until tests are merged

actions/image_partition_action.go
adionditsak: I think this is quite clear, but I added further details for clarity to this one

adionditsak:
Below is a recipe with root on ZFS on Debian Trixie. I only provided examples on Ubuntu this far. Since the ZFS kernel module is a part of the linux-image-generic on Ubuntu, but other Debian distros usually is better of with zfs-dkms, here is an example of that.

So now in this PR there should be recipe examples of:

- root on ZFS on Ubuntu 24
- root on ZFS on Debian 13
- root on ext with ZFS on another filesystem on Ubuntu 24

@obbardc Beside the test suites that you are working on, is there a place for recipe examples? 

root on ZFS on Debian 13:

```
# Build: sudo ./debos -m 4G --scratchsize 10G zfs-root.yaml
# Test:  qemu-system-x86_64 -machine q35,accel=kvm -cpu host -m 2G \
#          -drive if=pflash,format=raw,readonly=on,file=/usr/share/OVMF/OVMF_CODE_4M.fd \
#          -drive file=debian-zfs-on-root.img,format=raw,if=virtio
#

architecture: amd64

actions:
  - action: debootstrap
    suite: trixie
    components: [ main, non-free-firmware, contrib ]
    mirror: http://deb.debian.org/debian
    variant: minbase
    # ZFS overlay for hostid and resume config
 
  - action: overlay
    source: overlays/zfs

  # ubuntu has zfs kernel module included with linux-image-generic other debian-based distros is better of with zfs-dkms package
  - action: apt
    recommends: false 
    packages:
      - systemd
      - systemd-sysv
      - systemd-resolved
      - udev
      - firmware-linux
      - linux-image-amd64
      - linux-headers-amd64
      - openssl
      - zfs-dkms
      - zfsutils-linux
      - zfs-initramfs
      - dosfstools
      - iputils-ping
      - iproute2
      - adduser
      - sudo

  - action: run
    description: build and verify ZFS module
    chroot: true
    command: |
      set -eu
      KVER=$(ls /lib/modules)
      ZVER=$(ls -d /usr/src/zfs-*); ZVER=${ZVER##*-}
      dkms add "zfs/$ZVER"
      dkms install -k "$KVER" "zfs/$ZVER"
      update-initramfs -u -k "$KVER"

  - action: run
    description: enable networking
    chroot: true
    command: |
      set -eu
      systemctl enable systemd-networkd systemd-resolved

  - action: image-partition
    imagename: debian-zfs-on-root.img
    imagesize: 8GB
    partitiontype: gpt
    mountpoints:
      - mountpoint: /boot/efi
        partition: ESP
    partitions:
      - name: ESP
        fs: vfat
        start: 1MB
        end: 512MB
        flags: [ boot, esp ]
      - name: root
        fs: zfs
        start: 512MB
        end: 100%
        # properites below follow the OpenZFS "Root on ZFS" recommendations
        zfs:
          pool: zpool
          bootfs: zpool/ROOT/os
          pool-properties:
            ashift: 12
            autotrim: "on"
          dataset-properties:
            acltype: posixacl
            xattr: sa
            dnodesize: auto
            compression: lz4
            normalization: formD
            atime: "off"
            relatime: "on"
          # boot environment layout
          datasets:
            - name: ROOT
              properties: { canmount: "off", mountpoint: none }
            - name: ROOT/os
              properties: { canmount: noauto, mountpoint: / }
            - name: home
            - name: home/root
              properties: { mountpoint: /root }
            - name: var
              properties: { canmount: "off" }
            - name: var/lib
              properties: { canmount: "off" }
            - name: var/log
            - name: var/spool
            - name: var/cache
              properties: { "com.sun:auto-snapshot": "false" }
          swap:
            name: "swap"
            size: 2G
            properties:
              volblocksize: "4096"
              compression: zle
              logbias: throughput
              sync: always
              primarycache: metadata
              secondarycache: none
              "com.sun:auto-snapshot": "false"
          encryption:
            enabled: true
            keyfile: zfs-keyfile

  - action: filesystem-deploy

  # ZFSBootMenu
  - action: run
    description: install ZFSBootMenu
    chroot: false
    command: |
      set -eu
      ESP="$IMAGEMNTDIR/boot/efi"
      mkdir -p "$ESP/EFI/BOOT" "$ESP/EFI/ZBM"
      if [ -f "$RECIPEDIR/zfsbootmenu.EFI" ]; then
          cp "$RECIPEDIR/zfsbootmenu.EFI" "$ESP/EFI/BOOT/BOOTX64.EFI"
      else
          curl -fL -o "$ESP/EFI/BOOT/BOOTX64.EFI" https://get.zfsbootmenu.org/efi
      fi
      cp "$ESP/EFI/BOOT/BOOTX64.EFI" "$ESP/EFI/ZBM/VMLINUZ-BACKUP.EFI"
      zfs set org.zfsbootmenu:commandline="quiet rw" zpool/ROOT
      echo "root:debos" | chroot "$IMAGEMNTDIR" chpasswd
```

I use the same overlay:

```
# overlay
mkdir -p overlays/zfs/etc/initramfs-tools/conf.d
printf '\x0c\xb1\xba\x00' > overlays/zfs/etc/hostid
echo RESUME=none > overlays/zfs/etc/initramfs-tools/conf.d/resume

mkdir -p overlays/zfs/etc/systemd/network
cat > overlays/zfs/etc/systemd/network/10-dhcp.network <<'EOF'
[Match]
Name=en* eth*

[Network]
DHCP=yes
EOF

# zfs-keyfile
echo "testpassword123" > zfs-keyfile
```

obbardc:
@adionditsak please put your example recipe under `doc/examples/<todo>` in a separate commit (you can also include overlays and other configuration etc in there, see the existing example) :-)

adionditsak:
> @adionditsak please put your example recipe under `doc/examples/<todo>` in a separate commit (you can also include overlays and other configuration etc in there, see the existing example) :-)

@obbardc I just had a look and I think they will fit in well there in the project structure. With "separate commit", do you mean to do this on a separate branch + separate PR?

obbardc:
> > @adionditsak please put your example recipe under `doc/examples/<todo>` in a separate commit (you can also include overlays and other configuration etc in there, see the existing example) :-)
> 
> @obbardc I just had a look and I think they will fit in well there in the project structure. With "separate commit", do you mean to do this on a separate branch + separate PR?

Please add them in this branch & PR as separate commits :-).

obbardc:
@adionditsak there are golangci-lint issues to fix: https://github.com/go-debos/debos/actions/runs/30197947331/job/92514789843?pr=735

```
  Running [/github/home/golangci-lint-2.3.1-linux-amd64/golangci-lint config path] in [/__w/debos/debos] ...
  Running [/github/home/golangci-lint-2.3.1-linux-amd64/golangci-lint config verify] in [/__w/debos/debos] ...
  Running [/github/home/golangci-lint-2.3.1-linux-amd64/golangci-lint run] in [/__w/debos/debos] ...
  actions/image_partition_action.go:242:1: File is not properly formatted (gofmt)
  'bootfs' property. Setting it to a specific dataset makes this pool 
  ^
  wrapper/zfs_wrapper.go:288:1: File is not properly formatted (gofmt)
  
  ^
  2 issues:
  * gofmt: 2
```

actions/image_partition_action.go
obbardc: FYI, once you add the example recipes to this PR, I will create some tests with them :-)

adionditsak:
@obbardc Fixed linting issues, squashed commits into one commit, and pushed example recipes in another commit 8109f8bf9fc4414076c1d46b36c79e0119349a2e

PR #736, opened 2026-07-08
image-partition: add btrfs subvolume support

### LINKED ISSUE #733: Native support for btrfs subvolumes and filesystem-specific mount options
The image-partition can format and mount a btrfs partition, but it has no concept of subvolumes, so  any subvolume layout has to be rebuilt by hand in run actions (create the subvolumes, remount each with subvol=, set properties, write the fstab lines). It also mounts with an empty data argument, so mountpoint options aren't applied at mount time:

https://github.com/go-debos/debos/blob/15493799c3ee10c22f40bc09c018b34b67a8b07e/actions/image_partition_action.go#L685

It would be great if image-partition could create subvolumes and mount them with options directly, something like:

```
partitions:
  - name: root
    fs: btrfs
    subvolumes:
      - name: "@"
      - name: boot
        properties: { compression: none }
      - name: "@home"
mountpoints:
  - { mountpoint: /,     partition: root, subvolume: "@",     options: [compress=zstd, noatime] }
  - { mountpoint: /boot, partition: root, subvolume: boot }
  - { mountpoint: /home, partition: root, subvolume: "@home", options: [compress=zstd, noatime] }
```

debos would create the subvolumes, apply properties, mount them in path order with subvol= plus the options (and reuse them in the fstab lines it already writes). Mountpoints with no subvolume behave as they do now.

obbardc:
Add support for declaring btrfs subvolumes when partitioning an image. When a btrfs partition declares subvolumes, create them when partitioning an image, apply subvolume properties and mount subvolume-backed mountpoints using the `subvol=` mount option.

Also pass filesystem-specific mount options to the mount syscall instead of mounting with an empty data argument. This allows options such as `compress=zstd` and `noatime` to be applied at mount time and ensures the same options are included in generated fstab entries.

Mountpoints without a subvolume continue to behave as before.

Add unit tests covering btrfs subvolume creation, subvolume mount option handling, property handling and the unchanged behaviour for regular mountpoints.

Closes: #733

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -274,6 +274,26 @@ type Partition struct {
 	End             string
 	FS              string
 	Flags           []string
+	// TODO: Features (mkfs -O) and ExtendedOptions (mkfs -E) are the only mkfs
+	// flags exposed, so anything else has to be worked around post-mkfs or not
+	// at all. For btrfs on single-device flash images we need -m/-d (allocation
+	// profiles; the mkfs default has flipped between single and dup across
+	// btrfs-progs versions, so it cannot be relied on) and -s (sectorsize,
+	// fixed at creation and immutable afterwards, so unlike the profiles it
+	// cannot be corrected later). Rather than a field per flag, add a
+	// free-form escape hatch appended verbatim to the mkfs command, which also
+	// covers ext4 -i/-b, f2fs, xfs and so on without modelling each:
+	//
+	//   partitions:
+	//     - name: root
+	//       fs: btrfs
+	//       mkfs-options: "-m dup -d single -s 4096"
+	//
+	// To be done as a separate commit/PR; the property name is not settled
+	// (fs-create-options vs mkfs-options).
+	// current workaround handles the profiles only (post-mkfs, on the empty fs,
+	// idempotent, but it can't touch sectorsize):
+	// btrfs balance start -mconvert=dup,soft -sconvert=dup,soft -dconvert=single,soft --force "$IMAGEMNTDIR"
 	Features        []string
 	ExtendedOptions []string
 	Fsck            bool `yaml:"fsck"`
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -558,13 +558,16 @@ func (i ImagePartitionAction) setupBtrfsSubvolumes(p *Partition, context debos.C
 
 	for _, sv := range p.Subvolumes {
 		svPath := path.Join(mntpath, sv.Name)
+		// TODO: is this label useful ?
 		label := fmt.Sprintf("Creating subvolume %s on partition %d", sv.Name, p.number)
 		err := debos.Command{}.Run(label, "btrfs", "subvolume", "create", svPath)
 		if err != nil {
 			return err
 		}
 		for key, value := range sv.Properties {
-			err := debos.Command{}.Run("btrfs", "btrfs", "property", "set", svPath, key, value)
+			// TODO: is this label useful ?
+			label := fmt.Sprintf("Setting properties for subvolue %s", sv.Name)
+			err := debos.Command{}.Run(label, "btrfs", "property", "set", svPath, key, value)
 			if err != nil {
 				return err
 			}
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -60,6 +60,7 @@ a 32 bits hexadecimal number (e.g. '1234ABCD' without any dash separator).
 		   fsuuid: string
 		   partuuid: string
 		   partattrs: list of partition attribute bits to set
+		   subvolumes: list of btrfs subvolumes
 
 Mandatory properties:
 
@@ -128,10 +129,31 @@ and data.
 - extendedoptions -- list of additional filesystem extended options which need
 to be enabled for the partition.
 
+- subvolumes -- list of btrfs subvolumes to create on the partition. Only
+supported for the 'btrfs' filesystem. Subvolumes are created in the order they
+are listed, so nested subvolumes can be created by listing their parent first.
+Properties for subvolumes are described below.
+
+	   # Yaml syntax for subvolumes:
+	   subvolumes:
+	     - name: subvolume name
+		   properties: map of btrfs properties
+
+Mandatory properties:
+
+- name -- name of the subvolume to create, relative to the top-level of the
+btrfs filesystem. Must be unique within the partition.
+
+Optional properties:
+
+- properties -- map of btrfs properties (key: value) to set on the subvolume
+using `btrfs property set`, for example '{ compression: none }'.
+
 	   # Yaml syntax for mount points:
 	   mountpoints:
 	     - mountpoint: path
 		   partition: partition label
+		   subvolume: btrfs subvolume name
 		   options: list of options
 		   buildtime: bool
 
@@ -144,6 +166,12 @@ should be mounted. Must be unique, only one partition can be mounted per
 mountpoint.
 
 Optional properties:
+
+- subvolume -- name of the btrfs subvolume to mount at this mountpoint. The
+subvolume must be declared in the `subvolumes` list of the referenced partition.
+The subvolume is selected at mount time with `subvol=` and the same is written
+to the corresponding `/etc/fstab` entry. Only supported for the 'btrfs'
+filesystem.
 
 - options -- list of options to be added to appropriate entry in fstab file.
 
@@ -175,6 +203,34 @@ Defaults to false.
 	      start: 64MB
 	      end: 100%
 	      flags: [ boot ]
+
+	# Layout example using btrfs subvolumes:
+	- action: image-partition
+	  imagename: "debian-btrfs.img"
+	  imagesize: 2GB
+	  partitiontype: gpt
+	  mountpoints:
+	    - mountpoint: /
+	      partition: root
+	      subvolume: "@"
+	      options: [ compress=zstd, noatime ]
+	    - mountpoint: /boot
+	      partition: root
+	      subvolume: boot
+	    - mountpoint: /home
+	      partition: root
+	      subvolume: "@home"
+	      options: [ compress=zstd, noatime ]
+	  partitions:
+	    - name: root
+	      fs: btrfs
+	      start: 0%
+	      end: 100%
+	      subvolumes:
+	        - name: "@"
+	        - name: boot
+	          properties: { compression: none }
+	        - name: "@home"
 */
 package actions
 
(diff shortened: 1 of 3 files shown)


actions/image_partition_action.go
```diff
@@ -145,7 +167,14 @@ mountpoint.
...
 Optional properties:
 
+- subvolume -- name of the btrfs subvolume to mount at this mountpoint. The
+subvolume must be declared in the `subvolumes` list of the referenced partition.
+The subvolume is selected at mount time with `subvol=` and the same is written
+to the corresponding `/etc/fstab` entry. Only supported for the 'btrfs'
+filesystem.
+
 - options -- list of options to be added to appropriate entry in fstab file.
+The options are also applied when the partition is mounted during the debos run.
```
obbardc: move this and its implementation into its own commit ?

actions/image_partition_action.go
```diff
@@ -128,10 +129,31 @@ and data.
...
+
+Mandatory properties:
+
+- name -- name of the subvolume to create, relative to the top-level of the
+btrfs filesystem. Must be unique within the partition.
+
+Optional properties:
+
+- properties -- map of btrfs properties (key: value) to set on the subvolume
+using `btrfs property set`, for example '{ compression: none }'.
```
obbardc: wrapping with `{ }` is wrong ? the excample recipe below also uses this syntax ?

actions/image_partition_action.go
```diff
@@ -201,6 +258,11 @@ import (
 	"github.com/go-debos/debos"
 )
 
+type Subvolume struct {
```
obbardc, line 260: Should we rename to `BtrfsSubvolume` ? Maybe in future other filesystems may use subvolumes ? Not sure it matters as it's an implementation detail.

(also possibly rename `Partition.Subvolumes`)

actions/image_partition_action.go
```diff
@@ -682,9 +798,27 @@ func (i ImagePartitionAction) Run(context *debos.Context) error {
...
+			/* Shell out to mount(8) so it can correctly split the options
+			 * into kernel VFS flags, filesystem-specific options and
+			 * userspace-only options (e.g. x-systemd.*), and so the btrfs
+			 * subvolume can be selected with subvol=. */
+			options := []string{}
+			if m.Subvolume != "" {
+				options = append(options, "subvol="+m.Subvolume)
+			}
+			options = append(options, m.Options...)
+			err = debos.Command{}.Run("mount", "mount", "-t", fsType,
```
obbardc: nitpick: only have one call to `mount`

actions/image_partition_action.go
obbardc: also in `tests/partitioning-btrfs-subvolumes/test.yaml`

actions/image_partition_action.go
Yury-MonZon: Good idea, agreed. The build-time options change stands on its own, so its own commit sounds clean.

actions/image_partition_action.go
Yury-MonZon: It's valid either way, but the rest of the recipes/tests all use block style, so maybe match that for consistency:

```yaml
properties:
  compression: none
```

actions/image_partition_action.go
Yury-MonZon: Both fine by me. Slight lean toward keeping `Subvolume`/`Partition.Subvolumes` so it matches the `subvolumes` key.

actions/image_partition_action.go
Yury-MonZon: +1. A small `Mountpoint.mountOptions()` helper could fold the branches into one `mount` call and be reused in `generateFSTab` too.

actions/image_partition_action.go
```diff
@@ -466,6 +533,48 @@ func (i ImagePartitionAction) formatPartition(p *Partition, context debos.Contex
...
+
+	for _, sv := range p.Subvolumes {
+		svPath := path.Join(mntpath, sv.Name)
+		label := fmt.Sprintf("Creating subvolume %s on partition %d", sv.Name, p.number)
+		err := debos.Command{}.Run(label, "btrfs", "subvolume", "create", svPath)
+		if err != nil {
+			return err
+		}
+		for key, value := range sv.Properties {
+			err := debos.Command{}.Run("btrfs", "btrfs", "property", "set", svPath, key, value)
```
Yury-MonZon: Tiny thing: this logs with the bare label `"btrfs"`, while `subvolume create` just above uses a descriptive one. Might be nice to match for consistency.

Yury-MonZon:
Tested the current branch against our example recipe (root `@` + `boot` + `@home`, `compress=zstd,noatime`, `compression=none` on `boot`) and it works nicely: subvolumes get created, mounted with the right `subvol=` and options, and properties applied. Thanks for turning this around so fast.

One thing to flag, possibly as a follow-up rather than something for this PR: `generateKernelRoot` produces `root=UUID=...` with no `rootflags=subvol=`, and that value feeds `filesystem-deploy`'s `/etc/kernel/cmdline`. So for a subvolume-backed root that isn't the filesystem default (this PR doesn't `set-default`), the initramfs would mount the top-level subvol (id 5) instead of `@`, and the image wouldn't boot into the right tree even though fstab is correct. Might be worth appending `rootflags=subvol=<subvolume>` for the root mountpoint when it has a subvolume, or just noting it as a known limitation for now. Not a blocker either way.

Yury-MonZon:
Related request in the same area: `image-partition` only exposes `features` (mkfs `-O`) and `extended-options` (mkfs `-E`), so other mkfs flags can't be set. Two we need for single-device flash images:

- `-m`/`-d` (allocation profiles, e.g. `-m dup -d single`): matters for resilience/endurance, and the mkfs default has flipped between `single` and `dup` across btrfs-progs versions, so it can't be relied on.
- `-s <sectorsize>` (e.g. `-s 4096`): no workaround at all, sectorsize is fixed at creation and immutable afterward, so unlike the profiles it can't be corrected post-mkfs. Right now we inherit the page-size default and hope it's 4096.

Rather than a field per flag, could you add one free-form escape hatch appended verbatim to the mkfs command?

```yaml
partitions:
  - name: root
    fs: btrfs
    fs-create-options: "-m dup -d single -s 4096"
```

One field covers `-m`/`-d`/`-s` and anything else per filesystem (ext4 `-i`/`-b`, f2fs, xfs...) without modeling each. Our current workaround handles the profiles only (post-mkfs, on the empty fs, idempotent, but it can't touch sectorsize):

```
btrfs balance start -mconvert=dup,soft -sconvert=dup,soft -dconvert=single,soft --force "$IMAGEMNTDIR"
```

actions/image_partition_action.go
obbardc: Think leaving it like this is fine to me; also we error out if `subvolumes` property is set when the fs is not `btrfs`.

actions/image_partition_action.go
obbardc: Agree. Will do.

obbardc:
> Related request in the same area: `image-partition` only exposes `features` (mkfs `-O`) and `extended-options` (mkfs `-E`), so other mkfs flags can't be set. Two we need for single-device flash images:
> 
> * `-m`/`-d` (allocation profiles, e.g. `-m dup -d single`): matters for resilience/endurance, and the mkfs default has flipped between `single` and `dup` across btrfs-progs versions, so it can't be relied on.
> * `-s <sectorsize>` (e.g. `-s 4096`): no workaround at all, sectorsize is fixed at creation and immutable afterward, so unlike the profiles it can't be corrected post-mkfs. Right now we inherit the page-size default and hope it's 4096.
> 
> Rather than a field per flag, could you add one free-form escape hatch appended verbatim to the mkfs command?
> 
> ```yaml
> partitions:
>   - name: root
>     fs: btrfs
>     fs-create-options: "-m dup -d single -s 4096"
> ```
> 
> One field covers `-m`/`-d`/`-s` and anything else per filesystem (ext4 `-i`/`-b`, f2fs, xfs...) without modeling each. Our current workaround handles the profiles only (post-mkfs, on the empty fs, idempotent, but it can't touch sectorsize):
> 
> ```
> btrfs balance start -mconvert=dup,soft -sconvert=dup,soft -dconvert=single,soft --force "$IMAGEMNTDIR"
> ```

Think adding this as a field `fs-create-options` (in a separate commit/PR) makes sense. Maybe the property name needs changing though; something like `mkfs-options` or similar ?

actions/image_partition_action.go
obbardc: OK; I looked at this a bit more. I created https://github.com/go-debos/debos/pull/738 to document the build-time options (haven't decided if it's good idea to change the behaviour yet).

and in this PR I think I will just add the needed mount option `"subvol="+m.Subvolume` for btrfs mountpoints

obbardc: Done. with some TODOs.

actions/image_partition_action.go
obbardc: Done in latest commit.

actions/image_partition_action.go
obbardc: Done. But I think it may be wrong to set a dynamic label here and pollute logs. Will look at the log output first.

obbardc:
@Yury-MonZon pushed some more changes, can you review again, please ?

actions/image_partition_action.go
Yury-MonZon: Moving the behaviour change to #738 makes sense, keeps this one focused. subvol= only here works for me. I'll take a look at #738.

actions/image_partition_action.go
Yury-MonZon: Great, thanks.

actions/image_partition_action.go
Yury-MonZon: Agreed, and the non-btrfs guard covers it anyway. Fine to keep Subvolume.

actions/image_partition_action.go
Yury-MonZon: Fair point. If the per-subvol label turns out noisy, a static one is fine. No strong preference either way.

Yury-MonZon (commented):
Done

actions/image_partition_action.go
obbardc: Think one problem here is; if you create a subvolume with e.g. `options: [ compress=zstd, noatime ]` it doesn't get applied in the build environment and the subvolume isn't compressed?

Yury-MonZon: Yeah, exactly. It only ends up compressed if the mountpoint is mounted with those options at build time, and right now only subvol= is passed, not m.Options. So applying m.Options on the build mount is the fix.

Yury-MonZon:
Hi! Just checking in to see if you've had a chance to look at this. Thanks!

PR #737, opened 2026-07-15
fix: truncate existing /etc/fstab to prevent leftover data corruption

ginta1337:
## Problem

`setupFSTab()` opens an existing `/etc/fstab` with `O_RDWR|O_CREATE`, but does not truncate it before writing.

After the `debootstrap` stage, `/etc/fstab` already contains the default template:

```text
# UNCONFIGURED FSTAB FOR BASE SYSTEM
#
# <file system> <mount point>   <type>  <options>       <dump>  <pass>
```

The generated root filesystem entry is shorter than this template. As a result, the previous file contents are only partially overwritten, leaving trailing data from the original template.

For example, the resulting `fstab` becomes:

```text
UUID=12345678-1234-1234-1234-123456789abc / ext4 defaults,errors=remount-ro 0 1
options>       <dump>  <pass>
/swap none swap defaults 0 0
```

The leftover line is interpreted by `systemd-fstab-generator` as an invalid mount entry, producing warnings during boot.

This issue affects the generated Kali VM images (both stable and weekly), which are built using debos.

## Fix

Open `/etc/fstab` with `os.O_TRUNC` so any existing contents are removed before writing the new configuration.

Also change the file mode from `0755` to `0644`, which is the conventional permission for `/etc/fstab`.

### DIFF
--- a/actions/filesystem_deploy_action.go
+++ b/actions/filesystem_deploy_action.go
@@ -18,7 +18,8 @@ image.
 Optional properties:
 
 - setup-fstab -- generate '/etc/fstab' file according to information provided
-by 'image-partition' action. By default is 'true'.
+by 'image-partition' action. If an existing '/etc/fstab' is present,
+it is overwritten. By default is 'true'.
 
 - setup-kernel-cmdline -- add location of root partition to '/etc/kernel/cmdline'
 file on target image. By default is 'true'.
@@ -66,7 +67,7 @@ func (fd *FilesystemDeployAction) setupFSTab(context *debos.Context) error {
 	}
 
 	fstab := path.Join(context.Rootdir, "etc/fstab")
-	f, err := os.OpenFile(fstab, os.O_RDWR|os.O_CREATE, 0755)
+	f, err := os.OpenFile(fstab, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
 	if err != nil {
 		return fmt.Errorf("couldn't open /etc/fstab: %w", err)
 	}


actions/filesystem_deploy_action.go
```diff
@@ -66,7 +66,7 @@ func (fd *FilesystemDeployAction) setupFSTab(context *debos.Context) error {
 	}
 
 	fstab := path.Join(context.Rootdir, "etc/fstab")
-	f, err := os.OpenFile(fstab, os.O_RDWR|os.O_CREATE, 0755)
+	f, err := os.OpenFile(fstab, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
```
obbardc, line 70: We should document this in the `setup-fstab` documentation at the top of the file.

Include what you have in the commit message; about overwriting the file && the expected permissions.

Also is 644 the default across the board, e.g. arch/fedora etc ?

actions/filesystem_deploy_action.go
```diff
@@ -66,7 +66,7 @@ func (fd *FilesystemDeployAction) setupFSTab(context *debos.Context) error {
 	}
 
 	fstab := path.Join(context.Rootdir, "etc/fstab")
```
obbardc, line 69: your commit message should be expanded to include the problem; something like this:

```
filesystem-deploy: overwrite existing /etc/fstab

`setupFSTab()` opens an existing `/etc/fstab` in the target filesystem
with `O_RDWR|O_CREATE` but does not truncate it before writing
the generated configuration.

After debootstrap, `/etc/fstab` may contain a default template. Since the
generated target filesystem entry may be shorter than this template,
writing from the beginning of the file leaves trailing data from the
previous contents.

The leftover text may be interpreted during boot as an
invalid mount entry which can result in warnings during boot.

Open `/etc/fstab` with `O_TRUNC` so that any existing contents are
removed before writing the new configuration.

Also change the file mode from 0755 to the conventional 0644.
```

obbardc (changes requested):

actions/filesystem_deploy_action.go
ginta1337: Docs updated. Yes 644 is the default

ginta1337:
According to PR #715 and rgrep, this project has systematic issues with os.OpenFile flags and the default permissions of configuration files. It's surprising that no one has fixed this before.

I can help address these issues, but all documentation and comments will be generated by AI.

obbardc:
> According to PR #715 and rgrep, this project has systematic issues with os.OpenFile flags and the default permissions of configuration files. It's surprising that no one has fixed this before.
> 
> I can help address these issues, but all documentation and comments will be generated by AI.

Sure, please propose a separate PR for each change like this one, thank you!

Alternatively you could create an issue detailling all of the OpenFile flags and permission changes which you think are wrong, we can go through them one by one.

obbardc (approved):
LGTM, thanks!

obbardc:
@sjoerdsimons requesting a review from you before merging as I know I've got the filemode stuff wrong before :-)

PR #738, opened 2026-07-15
image-partition: rework mountpoint documentation

obbardc:
Rework the documentation for mount points `options` and `buildtime` properties. Clarify that additional mount options are only written to the target filesystem's `/etc/fstab` and are not used during the debos build and describe the behaviour of build-time-only mountpoints.

### DIFF
--- a/actions/image_partition_action.go
+++ b/actions/image_partition_action.go
@@ -145,14 +145,18 @@ mountpoint.
 
 Optional properties:
 
-- options -- list of options to be added to appropriate entry in fstab file.
+- options -- list of additional mount options to include in the `/etc/fstab`
+entry for this partition in the target filesystem. This option only has an
+effect when a later 'filesystem-deploy' action enables 'setup-fstab'. The
+options are not applied to the mount performed during the debos build, which
+always mounts the partition with the filesystem's default options.
 
-- buildtime -- if set to true then the mountpoint only used during the debos run.
-No entry in `/etc/fstab` will be created.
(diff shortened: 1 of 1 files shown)


actions/image_partition_action.go
```diff
@@ -145,14 +145,18 @@ mountpoint.
...
-- buildtime -- if set to true then the mountpoint only used during the debos run.
-No entry in `/etc/fstab` will be created.
-The mountpoints directory will be removed from the image, so it is recommended
-to define a `mountpoint` path which is temporary and unique for the image,
-for example: `/mnt/temporary_mount`.
-Defaults to false.
+- options -- list of additional mount options to include in the `/etc/fstab`
+entry for this partition in the target filesystem. This option only has an
+effect when a later 'filesystem-deploy' action enables 'setup-fstab'. The
+options are not applied to the mount performed during the debos build, which
```
obbardc, line 151: Not sure if we should actually apply mount custom options during the debos build; so for now I have left this how it is.

This was motivated by https://github.com/go-debos/debos/pull/736#discussion_r3540490376

Yury-MonZon:
LGTM. Good to have the fstab-vs-build-mount distinction spelled out, that trips people up.

actions/image_partition_action.go
Yury-MonZon: Reasonable to defer. Though on btrfs it does matter: files written during the build only compress if the build mount has compress=zstd, otherwise they stay uncompressed and a later remount only affects new writes. So there's a real case for it eventually.

